PR #30, opened 2016-03-01
IntPtr text manipulation overloads and encoding support

kekekeks:
Your current implementation ignores SkPaint's encoding, I've fixed that. I've also added IntPtr-based overloads, so optimizations like passing a pointer to the part of pinned `string` can be used.

### DIFF
--- a/binding/Binding/SKCanvas.cs
+++ b/binding/Binding/SKCanvas.cs
@@ -240,7 +240,7 @@ namespace SkiaSharp
 			if (paint == null)
 				throw new ArgumentNullException ("paint");
 
-			var bytes = System.Text.Encoding.UTF8.GetBytes (text);
+			var bytes = Util.GetEncodedText(text, paint.TextEncoding);
 			SkiaApi.sk_canvas_draw_text (Handle, bytes, bytes.Length, x, y, paint.Handle);
 		}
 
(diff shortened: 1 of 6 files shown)


binding/Binding/SKCanvas.cs
```diff
@@ -240,7 +240,7 @@ public void DrawText (string text, float x, float y, SKPaint paint)
 			if (paint == null)
 				throw new ArgumentNullException ("paint");
 
-			var bytes = System.Text.Encoding.UTF8.GetBytes (text);
+			var bytes = Util.GetEncodedText(text, paint.TextEncoding);
```
kekekeks: Well, I personally recommend to use [CoreFX coding style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) since it's supported by default by tools like visual studio and R# and more common in general, but it's up to you.

migueldeicaza:
I looked at this, and I think that the approach that is being taken is wasteful.

C# is passing a UTF-16 string, and we end up always creating a managed array to convert to the currently configured text encoding.   Instead, I propose that we introduce a helper method that saves the current text encoding, sets it to UTF-16, renders the passed text, and restores the previous value for the current SKPaint text encoding.

migueldeicaza:
I implemented the described approach.

Issue #24, opened 2016-02-25
On-screen rendering and ganesh backend support


kekekeks:
Do you plan to add hardware acceleration support? I have some native code for win/linux/ios/android that we use in Perspex for hw-accelerated drawing using Skia. For Windows it will add dependency on [ANGLE](https://github.com/google/angle), but performance improvements are worth of adding such dependency.

kekekeks:
Since opengl-related calls are too platform-dependent and complex for a user it's better to keep Garesh backend initialization and surface-flipping in the native lib. I also can see that currently your bindings don't have built-in on-screen rendering support and you leave the task of blitting image to screen to the user.

In [Perspex](https://github.com/Perspex/Perspex) we have a concept of a render target that allows to create a [rendering session](https://github.com/Perspex/Perspex/blob/master/src/Perspex.SceneGraph/Media/IDrawingContext.cs) which, when disposed, is flipped back to screen. It may be worth to do the same.

```csharp
interface IRenderTarget
{
     IRenderingContext CreateContext();
}


interface IRenderingContext : IDisposable
{
      SkCanvas Canvas { get; }
}

```

To end user it will look like this:

```csharp
var target = new OSWindowRenderTarget(hWnd/xid/nswindow);
...

using(var ctx = target.CreateContext())
{
     ctx.Canvas.DrawRectangle(...);
}

```

That render target will manage everything required to initialize ganesh-backend using ANGLE/EGL/GLX/AGLX/whatever or fall back to sw rendering.


Currently I have some code for that in Perspex

https://github.com/Perspex/libperspesk/blob/master/src/gl.cpp - EGL/ANGLE hw-accel (Win/Linux/Android)
https://github.com/Perspex/libperspesk/blob/master/src/win32.cpp - win32 window (SetDIBitsToDevice)
https://github.com/Perspex/libperspesk/blob/master/src/linux.cpp - X11 (XPutImage)
https://github.com/Perspex/libperspesk/blob/master/src/pview.mm - iOS using GLKView

mattleibow:
@kekekeks I will be looking into this soon. I have got most of the GPU bits building for all platforms right now. The next step is to get ANGLE for Windows added, then wrap all the necessary types. 

As the GPU bits are building already, initial wrapper can be created for OpenGL right now, except the UWP will have to wait for ANGLE. Win32 can have initial GL support until ANGLE arrives - giving it more options for devs.

mattleibow:
[...]

PR #36, opened 2016-03-03
Initial implementation for render-target concept (win32 sw-only)

kekekeks:
Initial code for #24. 

For now I'm struggling a bit with building opengl-renderer support (your scripts are skipping ANGLE build for some reason, so I need more time), so stub is used instead.

### DIFF
--- a/binding/Binding/Binding.projitems
+++ b/binding/Binding/Binding.projitems
@@ -15,6 +15,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)SKImageFilter.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKColorFilter.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKPaint.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)SkRenderTarget.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKSurface.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKCanvas.cs" />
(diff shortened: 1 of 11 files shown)


kekekeks:
For example [this code](https://github.com/mono/SkiaSharp/blob/master/samples/Skia.WindowsDesktop.Demo/SkiaView.cs#L29) can be now simplified as:
```csharp
		protected override void OnPaint(PaintEventArgs e)
		{
			if (onDrawCallback == null)
				return;
		    using (var ctx = _renderTarget.CreateRenderContext())
                onDrawCallback(ctx.Canvas, Width, Height);
		}

```

grokys:
Hi all, any feedback on this PR? We'd like to get our skia backend in [Perspex](https://github.com/Perspex/Perspex) using Skia# and this was intended as a first step.

Or if you already have plans for something to cover this, please let us know.

danwalmsley:
Hi, I am also waiting on this PR!!!

migueldeicaza:
This patch predates our use of the CLA bot, would you guys be so kind to resubmit the change, so that our CLA bot can request the signature?

kekekeks:
Ehm. I've signed CLA as was requested by CLA bot in [this PR](https://github.com/mono/SkiaSharp/pull/29) which was at March the 1st. This PR was submitted at March the 3rd.

mattleibow:
Thanks for this PR, but before I merge this in I want to have a look at getting the GPU bits working for all the platforms. 

Right now, I am looking at other issues in SkiaSharp as well as updating our Skia to the "latest", or rather m52. Once this is done, I will start on the GPU/render targets.

WRT the update, I am nearly done. I just have a few kinks to work out with Android ARMv6

Issue #39, opened 2016-03-05
NullReferenceException when creating SkCanvas backed by a WriteableBitmap


mgnslndh:
The [following code](https://github.com/impsnldavid/skiasharpwpfextensions/blob/master/src/Imp.SkiaSharpWpfExtensions/SkiaControl.cs#L72-L92) from https://github.com/impsnldavid/skiasharpwpfextensions fails randomly with a `NullReferenceException` when the `SkCanvas` is constructed.

The member variable `_bitmap` is a [WriteableBitmap](https://msdn.microsoft.com/en-us/library/system.windows.media.imaging.writeablebitmap(v=vs.110).aspx)

```
		protected override void OnRender(DrawingContext dc)
		{
			if (_bitmap == null)
				return;

			_bitmap.Lock();

			using (var surface = SKSurface.Create((int)_bitmap.Width, (int)_bitmap.Height, 
				SKColorType.N_32, SKAlphaType.Premul, _bitmap.BackBuffer, _bitmap.BackBufferStride))
			{
				if (IsClearCanvas)
					surface.Canvas.Clear(_canvasClearColor);

				Draw(surface.Canvas, (int)_bitmap.Width, (int)_bitmap.Height);
			}

			_bitmap.AddDirtyRect(new Int32Rect(0, 0, (int)_bitmap.Width, (int)_bitmap.Height));
			_bitmap.Unlock();

			dc.DrawImage(_bitmap, new Rect(0, 0, ActualWidth, ActualHeight));
                }
```

The exception randomly occurs on calls to [surface.Canvas.Clear(_canvasClearColor);](https://github.com/impsnldavid/skiasharpwpfextensions/blob/master/src/Imp.SkiaSharpWpfExtensions/SkiaControl.cs#L83)

The `NullReferenceException` is wrapped in a `TargetInvocationException`. Here are the call stacks:

## Inner NullReferenceException Stack Trace

```
   at SkiaSharp.SKObject.RegisterHandle(IntPtr handle, SKObject instance)
   at SkiaSharp.SKCanvas..ctor(IntPtr handle)
```

## TargetInvocationException Stack Trace
```
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #40, opened 2016-03-05
Fixed issue when exception would be thrown when property was null

mgnslndh:
Fix for random `NullReferenceException` described in issue #39


migueldeicaza:
One line change, no need for CLA

PR #42, opened 2016-03-07
added 'assemble-docs' makefile target

joelmartinez:
new target is to simplify build server configuration to deploy API docs to https://developer.xamarin.com/api/namespace/SkiaSharp/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 tools/
 output/
 binding/SkiaSharp.Portable/SkiaPortable.cs
+docs/SkiaSharp.tree
+docs/SkiaSharp.zip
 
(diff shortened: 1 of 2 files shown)


joelmartinez:
If you're curious, yes I saw that `update-docs` also assembles the deployment artifacts ... but the reason I wanted a separate target that only assembles, is because `mdoc update` will modify the XML files if the resulting assembly is different ... and thus, these changes would need to be checked in. I didn't want the build server having to deal with that ... `update-docs` should be run separately, by a user that will check in the XML changes :)

joelmartinez:
@migueldeicaza when you get a chance ... this is for the teamcity build :)

Issue #4, opened 2016-02-22
SKColorFilter - Rename MIN_CUBE_SIZE and MAX_CUBE_SIZE


migueldeicaza:
To MinCubeSize and MaxCubeSize.

migueldeicaza:
fixed

PR #49, opened 2016-03-28
Add FilterQuality property on SKPaint

petergolde:
Requires PR #4 for mono/skia.

### DIFF
--- a/binding/Binding/SKPaint.cs
+++ b/binding/Binding/SKPaint.cs
@@ -140,6 +140,18 @@ namespace SkiaSharp
 			}
 		}
 
+		public SKFilterQuality FilterQuality
+		{
+			get
+			{
+				return SkiaApi.sk_paint_get_filter_quality(Handle);
+			}
+			set
+			{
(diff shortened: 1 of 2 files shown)


petergolde:
I am happy to add documentation also, but I am not familiar with the tools to do so. If you point me to documentation on how to update the documentation I can amend this PR.

bholmes:
@petergolde,

Yes adding docs would be a great help!  We need to write up some instructions in the README.

For now what I have been doing is running in bash

	sh bootstrapper.sh -t docs

This will modify the docs directory, adding stubs for the classes and members that were added.  The problem that I have is that a bunch of whitespace diffs are introduced when I run this as well.  (I need to get with miguel to see why that is.)  I revert the changes I do not want, then fill in the new stubbed out doc entries.  

To verify my changes I then run...

	mdoc-export-html -o bill docs/en/

The output from this is not perfect either, but will give you a set of html pages to view and review your new changes.  Finally commit the changes to the docs folder.  

Make sense?

PR #50, opened 2016-03-28
Add SKPaint.GetTextPath to convert text into a path.

petergolde:
Requires mono/skia PR #5.

### DIFF
--- a/binding/Binding/SKPaint.cs
+++ b/binding/Binding/SKPaint.cs
@@ -248,6 +248,35 @@ namespace SkiaSharp
 			return (long)SkiaApi.sk_paint_break_text (Handle, buffer, length, maxWidth, out measuredWidth);
 		}
 
+		public SKPath GetTextPath(string text, float x, float y)
+		{
+			if (text == null)
+				throw new ArgumentNullException(nameof(text));
+			return new SKPath(SkiaApi.sk_paint_get_utf16_text_path(Handle, text, (IntPtr)text.Length, x, y));
+		}
+
+		public SKPath GetTextPath(IntPtr buffer, IntPtr length, float x, float y)
+		{
+			if (buffer == IntPtr.Zero)
+				throw new ArgumentNullException(nameof(buffer));
+			return new SKPath(SkiaApi.sk_paint_get_text_path(Handle, buffer, length, x, y));
+
+		}
+
+		public SKPath GetTextPath(string text, SKPoint[] points)
+		{
+			if (text == null)
+				throw new ArgumentNullException(nameof(text));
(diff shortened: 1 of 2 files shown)


petergolde:
I am happy to add documentation also, but I am not familiar with the tools to do so. If you point me to documentation on how to update the documentation I can amend this PR.

binding/Binding/SKPaint.cs
```diff
@@ -248,6 +248,35 @@ public long BreakText (IntPtr buffer, IntPtr length, float maxWidth, out float m
...
 
+		public SKPath GetTextPath(string text, float x, float y)
+		{
+			if (text == null)
+				throw new ArgumentNullException(nameof(text));
+			return new SKPath(SkiaApi.sk_paint_get_utf16_text_path(Handle, text, (IntPtr)text.Length, x, y));
+		}
+
+		public SKPath GetTextPath(IntPtr buffer, IntPtr length, float x, float y)
+		{
```
bholmes: Do you need the IntPtr version of these?  It seems that the string version would be all that we need.

binding/Binding/SKPaint.cs
```diff
@@ -248,6 +248,35 @@ public long BreakText (IntPtr buffer, IntPtr length, float maxWidth, out float m
...
+
+		public SKPath GetTextPath(string text, SKPoint[] points)
+		{
+			if (text == null)
+				throw new ArgumentNullException(nameof(text));
+			return new SKPath(SkiaApi.sk_paint_get_pos_utf16_text_path(Handle, text, (IntPtr)text.Length, points));
+		}
+
+		public SKPath GetTextPath(IntPtr buffer, IntPtr length, SKPoint[] points)
+		{
```
bholmes: Do you need the IntPtr version of these?  It seems that the string version would be all that we need.

bholmes:
@petergolde I have added some inline comments to the file diffs view.  

Do you need the IntPtr version of these?  It seems that the string version would be all that we need.

Before I start the merge of this one I want to discuss this.

migueldeicaza:
I think this is inline with some of the work that we have done on some APIs to be able to pass a pointer to an existing piece of text.

bholmes:
You are correct.  MeasureText, BreakText do.  I will start on this now.

bholmes:
Merged with b470adb2470aecccc856c74c6bf19768645020c2 92b76679ba4d648188fec830d03ada67396f4d33 and 180ffc999dff1af3a935bcf637d95a4cf834f2bd

binding/Binding/SKPaint.cs
petergolde: I did this to be similar to the other APIs that handled text. 

Peter

Issue #61, opened 2016-04-11
Support for SkColor HSV <-> RGB


tdenniston:
It'd be great to wrap these Skia functions from SkColor: `SkRGBToHSV` and `SkHSVToColor`.

I've already started work on this, but I figured there should be an issue open so that it can be discussed. One problem I'm having already (it would be great if someone can chime in) is how to add a new file `sk_color.{cpp,h}` to the Skia C API. I've modified the `core.gypi` file to include it, and it is compiling, but doesn't seem to be linking it into `libskia_windows.dll`.

tdenniston:
I figured out the linking issue. It also occurs to me that we could simply reimplement these functions in SkiaSharp instead of calling into the library. Any thoughts on this?

bholmes:
Agree.  It feels like these methods should be implemented in C#.  

https://github.com/mono/SkiaSharp/blob/master/binding/Binding/Definitions.cs#L72

bholmes:
How about

`static public SKColor FromHSV (...)`

Can not think of a nice way to get HSV off hand.  Could be `GetHSV (out h, out s, out v)`

tdenniston:
Could do a subclass approach, `SKColorHSV : SKColor` with implicit conversion to RGB. That starts to depart from Skia, though, which does not have a separate class.

migueldeicaza:
Add the `FromHSV` method and `GetHSV` approaches.

Some code from Mono that implements bits of this:

https://github.com/mono/mono/blob/master/mcs/class/System.Drawing/System.Drawing/Color.cs#L231

mattleibow:
Added support for to/from both HSV and HSL in 555a6263951e78c692be8b441625a4117f060ef1

PR #62, opened 2016-04-11
Add RGB <-> HSV conversion functions for SKColor (#61)

tdenniston:
Issue #61.

This is a direct translation from the Skia functions `SkRGBToHSV` and `SkHSVToColor`. Happy to make any changes.

### DIFF
--- a/binding/Binding/Definitions.cs
+++ b/binding/Binding/Definitions.cs
@@ -231,7 +231,7 @@ namespace SkiaSharp
 			return (int)(x * (1 << 16));
 		}
 
-        public override string ToString ()
+		public override string ToString ()
 		{
 			return string.Format (CultureInfo.InvariantCulture, "#{0:x2}{1:x2}{2:x2}{3:x2}",  Alpha, Red, Green, Blue);
 		}
--- a/binding/Binding/Definitions.cs
+++ b/binding/Binding/Definitions.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Skia Definitions, enumerations and interop structures
 //
 // Author:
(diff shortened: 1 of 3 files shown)


migueldeicaza:
This looks good, but we should also expose properties like System.Drawing color does to get the individual HSV properties, very much like RGB exist today.

binding/Binding/Definitions.cs
```diff
@@ -97,7 +97,123 @@ public SKColor WithAlpha (byte alpha)
...
+			case 1: r = q; g = v; b = p; break;
+			case 2: r = p; g = v; b = t; break;
+			case 3: r = p; g = q; b = v; break;
+			case 4: r = t; g = p; b = v; break;
+			default: r = v; g = p; b = q; break;
+			}
+			return new SKColor ((byte)r, (byte)g, (byte)b);
+		}
+
+		public static void ToHSV(SKColor color, out float hue, out float saturation, out float value)
```
migueldeicaza: It is not clear why we would like to have this static method, as it requires an SKColor to be there.

I rather just keep the instance method.

migueldeicaza:
See my comment about the static public API entry point.

Oddly, my other comment is gone.

We should also add properties to get the individual HSV properties, like we surface RGB.

mattleibow:
Just want to say +1 for the unit test!

tdenniston:
Thank you @xamarin-cla-bot ...

I fixed the issues from the above comments. I didn't know if we should cache the values of HSV in the SKColor instance or not. Currently they are computed lazily when you use the Hue, Saturation and Value property getters.

Issue #67, opened 2016-04-20
Handling return strings from C


agallero:
Hi,
Trying to implement a method that returns a string, I wonder if there is any "standard" for skia/skiasharp in how to do it. 

I looked at what is already implemented, but the only methods that return strings seem to be:

```c
const char* sk_imagedecoder_get_format_name_from_format (sk_imagedecoder_format_t cformat);
const char* sk_imagedecoder_get_format_name_from_decoder (sk_imagedecoder_t* cdecoder);
```

And they don't seem to work fine. They aren't either in the "official" skia distribution yet ( https://github.com/google/skia/tree/master/include/c )

The main issue is with memory management. For example, sk_imagedecoder_get_format_name_from_decoder returns a char* which is generated here:

```c++
const char* SkImageDecoder::GetFormatName(Format format) {
    switch (format) {
        case kUnknown_Format:
            return "Unknown Format";
        case kBMP_Format:
            return "BMP";
        case kGIF_Format:
            return "GIF";
 ...
```

So in this case we are returning constants, and no memory is allocated in the C side. But on the other hand, they shouldn't be deallocated by the C# bindings either.
The current binding declaration is:

```c#
		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
		public extern static string sk_imagedecoder_get_format_name_from_decoder(sk_imagedecoder_t cdecoder);
```
And accodring to this:
http://stackoverflow.com/questions/370079/pinvoke-for-c-function-that-returns-char
This means that the CLR assumes it needs to free the native memory, and that the memory was allocated with CoTaskMemAlloc. None of this is true, in fact, I don't even know which would be the assumptions in OSX or somewhere else where CoTaskMemAlloc don't exist. As the CLR will try to free the unmanaged memory, but that memory was a constant (and not allocated with CoTaskMemAlloc), this should crash.
And indeed, if I try:

```C#
            var ms = new FileStream("r:\\test.png", FileMode.Open);
            var mst = new SKManagedStream(ms);
            SKImageDecoder k = new SKImageDecoder(mst);
            string s = k.FormatName;
```
The app crashes.

[...]

PR #73, opened 2016-05-01
Added methods to SkTypeface

agallero:
This branch adds some methods methods to the SKTypeface class. It also fixes an issue where if the typeface didn't exist, it would not raise an error, but generate an Access violation later.

I added tests for the new methods and another for the error in case the typeface doesn't exist


agallero:
Hi,
It might look like I haven't signed, but I actually did. Can you review it?

bholmes:
@agallero Yes I can confirm that you have completed the CLA.  We can now review the PR.

mattleibow:
Hi @agallero,

Thanks for this PR. I am having a look at this now and I was wondering if you could make a few changes:

- First, could you change the whitespace around? Although not documented exactly, we are using 4 spaces in the C / C++ code and hard tabs in the managed C# code.
- Also, do we really need the SkString type? As you are just getting the `c_str()` result, maybe you could do that in the method. Although we are keeping the C layer very thin, this case might be better just to return the `const char *`. Here is an example method in [C](https://github.com/mono/skia/blob/xamarin-mobile-bindings/include/c/xamarin/sk_x_imagedecoder.h#L22) and then the wrapper in [C#](https://github.com/mono/SkiaSharp/blob/master/binding/Binding/SkiaApi.cs#L657). This will reduce the code and boundaries on the managed side as well as remove a type.

Other than those two, I think we can merge.

agallero:
Hi,
I'll try to see the whitespace: the main issue is that I have VS with the standard settings and I think even if you manually change the whitespace it will reformat the file everytime you change anything. I'll try editing the files with Xamarin Studio and resaving them.

About the strings, I've written more about them in issue #67 but basically, as far as I know the example you  mention doesn't work. Actually it could/should be modified to return a SkString. I mention sk_imagedecoder_get_format_name_from_format specifically in issue #67 because it doesn't really allocate memory in the C side (it is returning a constant), but the managed side tries to free the char* anyway, causing an access violation, at least when I tried the code.

sk_imagedecoder_get_format_name_from_format could be fixed by changing the bindings not to free the char* (since at least for now, it is returning constants that don't need to be freed). But it is a dangerous solution, if in the future sk_imagedecoder_get_format_name_from_format returns something that isn't a constant and needs to be deallocated in the .net side.

In my case, the font name can't be a constant, so I need to allocate memory for the char* in the C side. But then you can't free it in the .NET side, because .NET expects the memory in the C side to have been allocated with CoTaskMemAlloc (which is windows specific). So by publishing SkString, what I do is to provide a wrapper for the char* that is both allocated and deallocated in the C side.

The SkString approach actually mimics the SkData approach which is a similar wrapper for a byte* which again, in the case of SkData is allocated and deallocated at the C side.

binding/Binding/SKTypeface.cs
```diff
@@ -108,6 +110,46 @@ public int CharsToGlyphs (IntPtr str, int strlen, SKEncoding encoding, out ushor
 				}
 			}
 		}
-	}
+
+        public string FamilyName {
+            get {
+                var r = SkiaApi.sk_string_new_empty();
```
mattleibow: @agallero - this might cause GC issues as this type may be moved in memory while we are in unmanaged code. To avoid this, we have get a GC handle... Might be better to move this whole operation into unmanaged code and allocate there. Shouldn't be a problem as we are going to be returning that result (SKString). The GC can then clean that up when not in use anymore

binding/Binding/SKTypeface.cs
```diff
@@ -108,6 +110,46 @@ public int CharsToGlyphs (IntPtr str, int strlen, SKEncoding encoding, out ushor
...
+                try {
+                    SkiaApi.sk_typeface_get_family_name(Handle, r);
+                    var cstr = SkiaApi.sk_string_get_c_str(r);
+                    var clen = SkiaApi.sk_string_get_size(r);
+                    return Util.GetString(cstr, (int)clen, SKTextEncoding.Utf8); 
+                } finally {
+                    SkiaApi.sk_string_destructor(r);
```
mattleibow: If we return a SKString from unmanaged code, we won't have to do this each time. Just get the IntPtr, and then get the SKString... the GC will do the rest

mattleibow:
Apologies, I broke the merge with my update to the skia submodule :(

mattleibow:
My bad, clicked the wrong button there :( comment, not close!

agallero:
Matthew,
Thanks for all the comments. I will see to fix the whitespace and move the result of the get_typeface to the function result indeed. The original reason to have it as a parameter was to mimic as much as possible the C++ Api (where for some reason they have ```void getFamilyName(SkString* name) const;``` ) but yes, I think returning the string directly will make more sense. I'll re check the possible GC issues too.

Thanks for all the time!
   Adrian

agallero:
Hi,
I've modified the whitespace to be consistent with the rest, and modified sk_typeface_get_family_name to return an sk_string instead of void.

But I am not really sure if the pull request must be redone or this can be used, as it seems that all the commits from other pull requests are now also included?

mattleibow:
@agallero This PR is still fine. The other changes are just going to be merged back, so it won't do anything. The only thing is the skia sub-module... but your or I can fix that before I merge - no biggie.

binding/Binding/SKTypeface.cs
```diff
@@ -16,6 +17,7 @@ public class SKTypeface : SKObject
 		internal SKTypeface (IntPtr handle)
 			: base (handle)
 		{
+			if (handle == IntPtr.Zero) throw new Exception("Invalid typeface");
```
mattleibow: I do have a question on this... We can't do this here as there might be a case when the managed instance is created, but the native instance hasn't yet. Also, as this is internal, it should never be called directly.

Rather, use the GetObject<T> method which will return null if a native method returned null. If this is in a constructor, then you can throw.

Here is a case of all 3:
https://github.com/mono/SkiaSharp/blob/master/binding/Binding/SKData.cs#L41-L65

The first `public SKData (IntPtr bytes, ulong length)` need to do a few checks before initializing the native instance, and then sets the `Handle` property.
The second `public SKData (byte[] bytes)` uses the base initializer, and then throws if the native was null.
The third `public static SKData FromMallocMemory (IntPtr bytes, ulong length)` uses the `GetObject<T>` method which does a few checks. If it was native null, then return managed null. Although not in this case, if the managed wrapper already exists, it will return that instance. Finally, it will create a new wrapper using reflection and the internal constructor.

binding/Binding/SKTypeface.cs
```diff
@@ -108,6 +110,54 @@ public int CharsToGlyphs (IntPtr str, int strlen, SKEncoding encoding, out ushor
...
+			}
+		}
+
+		public UInt32[] GetTableTags()
+		{
+			int tableCount = SkiaApi.sk_typeface_count_tables(Handle);
+			UInt32[] result = new UInt32[tableCount];
+			int r = SkiaApi.sk_typeface_get_table_tags(Handle, result);
+			if (r == 0) {
+				throw new Exception("Unable to read the tables for the file.");
```
mattleibow: @agallero I want to know what you think, but maybe we should have methods for each native operation. In this instance, we will have one for `sk_typeface_count_tables` and one for `sk_typeface_get_table_tags`. We can then have this one that calls those two. I am wanting to keep the APIs as close as possible.

Also, that exception... do we want it? Currently, exceptions are only thrown for invalid arguments. I am thinking we should do that here too. In this case, it might be better to return null.

If we have two separate methods, the user will get a `0` return, and I think this should match that with a null as `0` is an error and we can indicate that with `null` array.

binding/Binding/SKTypeface.cs
agallero: Hi,
Indeed, it can be changed to check when you call instead of when you create it, but I am not sure it is the best approach. The reasons I choose to check here are:

1)It implied the lest code changes in the existing codebase. 

2)Conceptually, I think a constructor that takes a handle, should check the handle is valid. The SkData(IntPtr, length) that you mention does that. I think if the need arises to create a managed instance before the native instance is created, the best would be to create a new constructor SKTypeface() and explicitly call that constructor from the methods that need to be created with a null handle, instead of calling the constructor that takes a handle with IntPrt.Zero.

3)I think it is less prone to errors. Here, you test at creation that the handle is invalid, and every method can safely assume the handle is valid. If we move the checks to the callers, we need to call GetObject or check for IntPrt.Zero in each methods that needs to deal with the handle. There are much more calls to change, and the possibility to forget one is higher.

But of course there is no "right" answer, both ways would work. I just wanted to state the reasons why I like it more to check at the constructor, but if you prefer to check later, I can change it, it is no problem.

mattleibow:
@agallero Just a note, if you want to chat directly, you can do so with Gitter: https://gitter.im/xamarin/XamarinComponents

binding/Binding/SKTypeface.cs
agallero: About the count_tables/get_table_tags duo, I tried to keep the C Api as close as possible to the C++ api, so there are 2 methods indeed. But the only reason I can think you would want to know the table count is so you can allocate memory and then call GetTableTags. It seems just the classing workaround for C arrays not having a length.

If you look at the comments in the C++ methods:
```
    /** Return the number of tables in the font. */
    int countTables() const;

    /** Copy into tags[] (allocated by the caller) the list of table tags in
     *  the font, and return the number. This will be the same as CountTables()
     *  or 0 if an error occured. If tags == NULL, this only returns the count
     *  (the same as calling countTables()).
     */
    int getTableTags(SkFontTableTag tags[]) const;

```
I think the idea is that you call CountTables, allocate an array of ints, and then call GetTableTags with the allocated array. You might even forget about CountTables and call twice GetTableTags: the first time to know the memory to allocate, the second to actually get the tags. I don't see any other reason why you would want to call CountTables alone.

Currently, if you want to know the tables in a font from C#, you would call:
```
  int[] tables = typeface.GetTableTags();
```

If we changed it to follow the C++ api, you would have to write:

```
int[] tables = new int[typeface.GetTableCount()];
typeface.GetTableTags(ref tables);
```

The first approach I think is more inline with what a C# user would expect, the second is more what a C user would expect. (and that is why I kept the C Api that way, similar to the C++ API).

A side note: SKIA internally uses FreeType and CoreGraphics to return the tables and tablecount. Coregraphics provides for getting the tables just returns an array too, there is no CountTables there:
https://developer.apple.com/library/mac/documentation/Carbon/Reference/CTFontRef/#//apple_ref/c/func/CTFontCopyAvailableTables

So if you look at SkTypeface_Mac::onGetTableTags(SkFontTableTag tags[]) you'll see it just gets an array with the tags, then returns the count as the array length.
In freetype there are indeed 2 calls to the same method ( FT_Sfnt_Table_Info ), but it is just using the standard C approach: First call with null to get the table count, then a second call to get the actual data.

About the exception, well, it can be indeed changed, but I don't think in this particular case it would be a problem: It should never fire. All fonts have a number of required tables, and a font without any table wouldn't have any data. I am not 100 % if there is any font format that doesn't have tables, but the common ones (ttf, otf, etc) have a number of required tables like the font header. A return of 0 or null here should be exceptional, and probably caused by a corrupt font or out of memory condition. I think raising an exception here makes sense so you don't have to check for null (or 0 tables) since that means a bigger error anyway, most likely one you can't handle.

What I am not 100% sure is about the exception in  GetTableData. I originally designed it so it owuld throw an Exception if the table tag doesn't exist, but this proved to be a problem even in my own code, where I regularly want to try to get data from optional tables and I don't want to first retrieve the list of available font tables with GetTableTags, and see if the tag is present in the array to only then call GetTableData. 

What I did was to add a new method TryGetTableData, which is similar to GetTableData but will return null instead of an exception. It follows the Parse/TryParse or GetValue/TryGetValue pattern in .NET, where you have a method that returns an exception, but if you need to call that method many times and it can fail, it is better to call TryMethod instead of Method and catch the exceptions. But we could modify GetTableData to return null if the tag doesn't exists (and so remove TryGetTableData)

mattleibow: Hi @agallero,

You have good points there. I agree fully.

migueldeicaza:
Can we merge this?

binding/Binding/SKTypeface.cs
mattleibow: Hi @agallero,

I get you. However, in this case this is an internal constructor (probably should be private) and should not be called outside the class. Also, the reflection-based instantiation might pass a `IntPtr.Zero` value during construction.

The correct way to instantiate this class is by overloading the constructor and passing a specific `IntPtr` and checking *that* value is not `Zero`. The other way is to use `GetObject<T>()` (which should be the only way to construct objects, but this cannot be done in constructors):

```csharp
// these are typical, but should only be used if we know for sure we are 
// constructing a new native object.
// because we are constructing a new object, we check for IntPtr.Zero
public SKMyType()
    : this(SkiaApi.create_my_type())
{
    if (Handle == IntPtr.Zero)
        throw new Exception("unable to create");
}
public SKMyType(int offset)
    : this(IntPtr.Zero) // valid case, possibly by a derived type
{
    var items = SkiaApi.get_some_array();
    if (items == IntPtr.Zero)
        throw new Exception("unable to create");
    Handle = SkiaApi.get_some_value(items, offset);
    if (Handle == IntPtr.Zero)
        throw new Exception("unable to create");
}

// for all other cases, we must go through GetObject so that it can be registered in case
// this method returns the same unmanaged instance on multiple calls.
// we do not want to create multiple managed instances for a single unmanaged instance.
// also, the unmanaged might be NULL, so we either have to do a specific null-check
// for each call, or we just use GetObject
public SKMyType GetMyType()
{
    return GetObject<SKMyType>(...);
}
```

I already made the changes in master to use `GetObject<T>()` (the omission was a bug): https://github.com/mono/SkiaSharp/blob/master/binding/Binding/SKTypeface.cs#L34

agallero: Hi @mattleibow 
That's ok, I have removed the check from the SKTypeface constructor taking an IntPtr handle.

mattleibow:
All right, merging now. Thanks for the work

PR #80, opened 2016-05-08
Fix typo in OSX demo.

bratsche:


mattleibow:
Thanks for this @bratsche ! All those typos...

PR #110, opened 2016-07-10
CharsToGlyphs(SKEncoding) does not use the specified encoding.

espresso3389:
I believe this is just a mistake on copy-n-paste the code. But CharsToGlyphs variant that accepts SKEncoding should use encoding instead of SKEncoding.Utf16.


mattleibow:
@espresso3389 Thanks for catching that. As soon as you sign the CLA, we can merge.

Issue #47, opened 2016-03-18
Support for dashed lines


Rungee:

bholmes:
I quickly found that this patch is more involved than adding a missing property or method.  I have created a [branch][1] to discuss the best way to integrate PathEffects.

Ensuring that we wrap the return of the getter of the PathEffect property with the correct managed object is the point of discussion.

[1]: https://github.com/mono/SkiaSharp/tree/PathEffects-2016-03

petergolde:
I don't think, in this particular case, that exposing subclasses of SkPathEffect serves any significant purpose, because the subclasses in the C++ API don't expose additional public members. I would just expose the single sealed class SKPathEffect, along with static Create function. E.g.:

static SKPathEffect CreateDashed(float[] intervals, float phase)

petergolde:
Obviously the other PathEffects would have their own CreateXXX function.

hypeartist:
Could anyone, please, make a windows binary (dll) of SkiaSharp with SKPathEffect added?

migueldeicaza:
I agree with Peter's take.  We do not need to bind the subclasses (and in fact, I am wondering if we need to surface all of those SKStream subclasses in the first place as well).

We should just provide convenient factory methods to create these.

bholmes:
@migueldeicaza First I agree that we should do this.  I need to check the remaining path effect classes to see if there are any critical members that will be missing.  

There are some members on these effects, but they feel advanced to me.  Also I do not think creating factory methods would stop us from later surfacing these if needed.  

I see two similar options.  One add the factory methods to the paint object in C# which will handle creating the effect and assigning in C.  The other option is to create a C# class (SkPathEffect) that only has static factory methods that take the paint as an argument to assign the new effect to.  Again in C.

In both options I think we need a way to clear the effect from the paint.  That could be as simple as a ResetEffect method on the paint class.  

Thoughts?

migueldeicaza:
I like it.

petergolde:
I would have a read/write property on SKPaint of type SKPathEffect.

SKPathEffect would have only static creation methods, e.g.:

public static SKPathEffect CreateDashedPathEffect(...);

I think this is

1) most like the skia C++ api
2) discoverable
3) doesn't overly pollute the SKPaint members
4) could extend later if we allow user-written subclasses of SKPathEffect (I don't think we should be go here now though)

Peter

hypeartist:
[...]

PR #112, opened 2016-07-13
Added Path Effects

mattleibow:
This is the code for issue #47

### DIFF
--- a/binding/Binding/Binding.projitems
+++ b/binding/Binding/Binding.projitems
@@ -16,6 +16,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)SKColorFilter.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKPaint.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKDocument.cs" />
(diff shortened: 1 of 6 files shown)


migueldeicaza:
LOVE IT!

bholmes:
I am not seeing ref and unref calls on the new SKEffect object.  What is keeping the object alive or causing it to be cleaned up?

migueldeicaza:
@bholmes it inherits the Finalizer/Disposer from SKObject which has the unref, and the object creation creates the objects with refcount=1

mattleibow:
@migueldeicaza In this case, @bholmes is right because `SKObject` does not have the unref. This is because not all SKObjects are `SKRefCnt` objects. An example would be `SKBitmap`/`SKPaint`, which is just a plain object and the destructor is used.

I think there are just those two variants, so I could create another type that uses the unref and the base uses the destructor, but for simplicity's sake, I just unref in the dispose methods. So, in this case, I forgot to add the unref, which I have added now.

bholmes:
@mattleibow we can merge this when you are ready.  I know you are working on a new version + upstreaming changes.  So I will let you decide when is the best time to merge.

mattleibow:
I am merging this in #116

PR #123, opened 2016-07-29
Merge for Release

mattleibow:


binding/Binding/SKTypeface.cs
```diff
@@ -110,10 +110,6 @@ public int CharsToGlyphs (string chars, out ushort [] glyphs)
 
 		public int CharsToGlyphs (IntPtr str, int strlen, SKEncoding encoding, out ushort [] glyphs)
 		{
-			if (str == null)
```
bholmes: Is this a correct?  You removed this arg check.

mattleibow: Yes, `str` is an `IntPtr` which will never be `null`

PR #169, opened 2016-10-24
Add SKRegion support

Tylerflick:
This changeset add the basic bindings needed for the SKRegion class.

### DIFF
--- a/binding/Binding/SkiaApi.cs
+++ b/binding/Binding/SkiaApi.cs
@@ -1132,22 +1132,32 @@ namespace SkiaSharp
 		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
 		public extern static sk_region_t sk_region_new2(sk_region_t r);
 		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
+		[return: MarshalAs(UnmanagedType.I1)] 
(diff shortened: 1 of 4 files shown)


mattleibow:
Awesome! This is looking good already!

binding/Binding/SkiaApi.cs
```diff
@@ -1120,7 +1125,32 @@ internal static class SkiaApi
...
 		public extern static bool gr_glinterface_has_extension (gr_glinterface_t glInterface, [MarshalAs(UnmanagedType.LPStr)] string extension);
-
+		
+		// SKRegion
+		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
+		public extern static sk_region_t sk_region_new();
+		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
+		public extern static sk_region_t sk_region_new2(sk_region_t r);
+		[DllImport(SKIA, CallingConvention = CallingConvention.Cdecl)]
+		public extern static bool sk_region_contains(sk_region_t r, sk_region_t region);
```
mattleibow: return types of `bool` don't play nice with P/Invoke. could you add the 
```csharp
[return: MarshalAs(UnmanagedType.I1)] 
```
to each that return Booleans?

Issue #186, opened 2016-11-07
Add pixel modification support for index8 bitmaps


Tylerflick:
This should be as simple as exposing `void SkBitmap::setPixels(void* p, SkColorTable* ctable)` for setting the pixel values, and adding index8 support to `sk_bitmap_get_pixel_colors `.

I don't believe that the SkBitmap class supports modification to a single pixel when the backing color type is index8.

PR adding support #196

Tylerflick:
Opened a [PR](https://github.com/mono/skia/pull/36) on mono/Skia to add support for the fix.

mattleibow:
@Tylerflick I just was thinking, since we don't support setting a specific pixel color for Index8, maybe we should throw (in managed code) an exception if someone calls set pixel with an Index8 color type.

If support is added, we just need to remember to remove that check. What do you think?

Tylerflick:
I like that idea.

We should also probably add a guard to `SetColorTable(SKColorTable ct)` & `SetPixels(IntPtr pixels, SKColorTable ct = null)` if the ColorType isn't Index8.

I also need to add some documentation to these new methods.

mattleibow:
to update the docs - it will fail because you need to be running on CI (both mac and Windows bits), but you can still generate the things for SkiaSharp.dll (where we are working). Basically, run this target:

    .\bootstrapper.ps1 -Target update-docs

This will mostly work, but you can just use the bits that you added inside the xml docs. This is a bit complicate to explain, but if you aren't running this for all the platforms, then it removes those types. So for you, run the command and then just add the `/docs/en/SkiaSharp/*.xml`. The other files will all be messed up because some types aren't there - so just revert them.

The final result will be a fairly small diff with new sections like this:

```xml
    <Member MemberName="ColorTable">
      <MemberSignature Language="C#" Value="public SkiaSharp.SKColorTable ColorTable { get; }" />
      <MemberSignature Language="ILAsm" Value=".property instance class SkiaSharp.SKColorTable ColorTable" />
      <MemberType>Property</MemberType>
      <AssemblyInfo>
        <AssemblyVersion>1.53.0.0</AssemblyVersion>
        <AssemblyVersion>1.54.0.0</AssemblyVersion>
        <AssemblyVersion>1.54.0.0</AssemblyVersion>
        <AssemblyVersion>1.55.0.0</AssemblyVersion>
      </AssemblyInfo>
      <ReturnValue>
[...]

PR #196, opened 2016-11-14
Add colortable setter support

Tylerflick:
Adds a support for setting the ColorTable of a bitmap. Technically this is for index8 bitmaps, but I've left in the logic to check that to the actual Skia library.

Addresses #186.
Requires https://github.com/mono/skia/pull/36.

### DIFF
--- a/binding/Binding/SKBitmap.cs
+++ b/binding/Binding/SKBitmap.cs
@@ -28,6 +28,7 @@ namespace SkiaSharp
 		{
 			releaseDelegate = new SKBitmapReleaseDelegateInternal (SKBitmapReleaseInternal);
 		}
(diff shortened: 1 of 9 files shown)


mattleibow:
I merged the https://github.com/mono/skia/pull/36 PR, that was good. I think all that is left is to remove the `null` default param value. Sorry about the conflict. But, this might be interesting to you: https://github.com/mono/SkiaSharp/commit/8db2bb0534934d054acf0f98ab273bed1234566d

Basically, you can tell the bitmap which pixels to use, and when it is finished it will invoke a callback to dispose of the pixel memory correctly. See the test: https://github.com/mono/SkiaSharp/commit/8db2bb0534934d054acf0f98ab273bed1234566d#diff-bb90be898c3b02230b061ed89cb39a5aR11

Tylerflick:
Sorry about the nasty merge. I tried to rebase but git was not having it.
I've the null optional removed.

PR #206, opened 2016-11-28
Make the iOS Forms views non-opaque

### LINKED ISSUE #198: Opaque Forms SKGL/CanvasView on iOS
Hi,

When trying to draw a simple filled circle, the Forms canvas rendered onto a black square on iOS, while the Android and UWP versions did just fine and were rendered on a transparent background.

I'm guessing the underlying issue is something along the lines of https://forums.xamarin.com/discussion/37571/problem-with-lost-transparency-using-cgcontext-drawimage.

For context, here's the code for the circle view:
```cs
    public class RoundView: SKCanvasView //*/SKGLView
    {
        private SKPaint paint;

        public static BindableProperty ColorProperty = 
            BindableProperty.Create(nameof(Color), typeof(Color), typeof(RoundView), Color.White,
                propertyChanged: (bindable, oldValue, newValue) => {
                    var ctrl = (RoundView)bindable;
                    ctrl.Color = (Color)newValue;
                });
        public Color Color {
            get{ return (Color)GetValue (ColorProperty); }
            set {
                SetValue (ColorProperty, value);
                // Oddly enough, the constructor doesn't work in Android !
                SKColor c = SKColor.Empty;
                c = c.WithAlpha((byte)(value.A * 255));
                c = c.WithRed((byte)(value.R * 255));
                c = c.WithGreen((byte)(value.G * 255));
                c = c.WithBlue((byte)(value.B * 255));
                //var c = new SKColor((byte)(value.R * 255), (byte)(value.G * 255), (byte)(value.B * 255), 
                paint.Color = c;
            }
        }

        public RoundView(): base()
        {
            paint = new SKPaint();
            paint.Style = SKPaintStyle.Fill;
        }

        protected override void OnPaintSurface(SKPaintSurfaceEventArgs args)
        //protected override void OnPaintSurface(SKPaintGLSurfaceEventArgs args)
        {
            base.OnPaintSurface(args);
            var width = args.Info.Width;
            var height = args.Info.Height;
            //var width = args.RenderTarget.Width;
            //var height = args.RenderTarget.Height;
            var canvas = args.Surface.Canvas;
            var opacity = args.Info.IsOpaque;
            canvas.Clear();
            canvas.DrawCircle(width / 2, height / 2, Math.Min(width, height) / 2, paint);
        }
    }
```

laarmen:
By default, the UWP and Android views are not opaque. These patches explicitly change the iOS renderers to copy this behavior.

I split the patches in two because I think the GL one could have some performance impacts.

This would fix #198

### DIFF
--- a/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.iOS/SKCanvasViewRenderer.cs
+++ b/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.iOS/SKCanvasViewRenderer.cs
@@ -86,6 +86,9 @@ namespace SkiaSharp.Views.Forms
 				UserInteractionEnabled = false;
 
 				this.controller = controller;
+
+				// Force the opacity to false for consistency with the other platforms
+				Opaque = false;
 			}
 
 			public override void DrawInSurface(SKSurface surface, SKImageInfo info)
--- a/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.iOS/SKGLViewRenderer.cs
+++ b/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.iOS/SKGLViewRenderer.cs
@@ -132,6 +132,9 @@ namespace SkiaSharp.Views.Forms
 				UserInteractionEnabled = false;
 
 				this.controller = controller;
+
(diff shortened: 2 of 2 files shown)


mattleibow:
@laarmen as soon as you sign the CLA, we can merge

PR #210, opened 2016-12-07
Add bitmap resize support

Tylerflick:
This commit adds support for leveraging Skia's methods for resizing bitmaps. Skia provides the use convolution matrices to resize images, which has a noticeable quality gain over using canvas scaling when downsampling bitmaps.

Depends on https://github.com/mono/skia/pull/38.

### DIFF
--- a/binding/Binding/Definitions.cs
+++ b/binding/Binding/Definitions.cs
@@ -536,7 +536,17 @@ namespace SkiaSharp
 		RgbaF16
 	}
 
-	[Obsolete ("May be removed in the next version.")]
+	public enum SKResizeMode {
+		Box,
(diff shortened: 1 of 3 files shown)


mattleibow:
@Tylerflick saw this and I was already working on something very similar. I did almost what you did but did most of the work in the C# layer, just exposing the `SkPixmap` type from the C layer. This will provide additional options for other APIs. Also, I moved the test in from the C layer to the C# layer so that they can be run by CI. I don't actually run the C tests - I do the testing via the interop so I can test both the C and the C# in one go.

Tylerflick:
Saw that 👍 . Closed this PR and the skia PR.

PR #211, opened 2016-12-13
Add SKMask & basic Blitting support

Tylerflick:
## Add SKMask and Blitting Support
This PR requires https://github.com/mono/skia/pull/39.

Details about this PR can be found @ https://github.com/mono/skia/pull/39.

### DIFF
--- a/binding/Binding/Binding.projitems
+++ b/binding/Binding/Binding.projitems
@@ -37,6 +37,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)SKStream.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKManagedStream.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKBitmap.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)SKMask.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKCodec.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)SKRegion.cs" />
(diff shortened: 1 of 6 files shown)


mattleibow:
Is there a reason that you used a class SkMask instead of a struct? Typically we want to keep structs structs and classes classes.

Also, what exactly is the blitter being used for in this case? the blitters are "internal"-ish (they have no SK_API and they live under the src folder instead of the public headers/include folder). Even though you may have a 1 bpp bitmap, you are still using a 1 byte mask and blitter. How is this better/different than using a `SKColorType.Gray8` image? that is also a 1 byte per pixel image.

Am I correct in my understanding that this was to be able to load a nice color image. and then after loading the mask, apply the mast to the bitmap, turning that bitmap into a transparent bitmap?

Tylerflick:
> Is there a reason that you used a class SkMask instead of a struct? Typically we want to keep structs structs and classes classes.

Makes sense. I'll fix that.

> Even though you may have a 1 bpp bitmap, you are still using a 1 byte mask and blitter.

This is not true with the SkMask class or stencil masks in general. In Skia the bitmaps raw byte array is loaded into the mask. This avoids inflating 1bpp masks 8 fold.

>  How is this better/different than using a SKColorType.Gray8 image?
For blended/soft masks it's not. In fact for most blended masks the default value of the mask is a 
Gray8 image.

I'm not really sure why Skia is exposing a 32-bit, 16-bit, etc. mask type. I'm still doing a bit of work on this and haven't updated that branch in a bit.

Tylerflick:
After digging through Skia a bit further, SkDraw may provide the needed functionality and is exposed in the include folder.

PR #212, opened 2016-12-20
Fix for memory leak in UWP SKXamlCanvas 

DerPate2010:
The ImageBrush used as background holds a reference to the WriteableBitmap


migueldeicaza:
This is a one-line change, no need for the CLA

PR #238, opened 2017-02-10
Use static libfreetype for Linux

kekekeks:
I've managed to produce a portable Linux binary using Holy Build Box.

However, I had to use static version of libfreetype.

https://files.gitter.im/AvaloniaUI/Avalonia/xFjI/libSkiaSharp.so - portable binary file.

I'll upload my docker container and instructions soon. That should allow to produce a single build of libSkiaSharp for most of modern Linux distributions based on glibc. It should work on

-    Debian >= 7
-    Ubuntu >= 12.04
-    Red Hat Enterprise Linux >= 6
-    CentOS >= 6

And their derivatives. I had to use more recent fontconfig, so library won't work on CentOS 5 and Debian 6, but I'm not sure that Skia# needs to target those ancient distros.

### DIFF
--- a/cake/BuildExternals.cake
+++ b/cake/BuildExternals.cake
@@ -439,7 +439,7 @@ Task ("externals-linux")
         SetEnvironmentVariable ("SKIA_OUT", outPath);
 
         // build skia_lib
-        RunGyp ("skia_os='linux' skia_arch_type='" + arch + "' skia_gpu=1 skia_pic=1 skia_pdf_use_sfntly=0", "ninja");
+        RunGyp ("skia_os='linux' skia_arch_type='" + arch + "' skia_gpu=1 skia_pic=1 skia_pdf_use_sfntly=0 freetype_static=1 skia_freetype_static=1", "ninja");
         RunProcess (ninja, new ProcessSettings {
             Arguments = "-C out/" + folder + "/Release " + targets,
             WorkingDirectory = SKIA_PATH.FullPath,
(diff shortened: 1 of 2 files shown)


kekekeks:
https://hub.docker.com/r/kekekeks/holy_build_box_for_skia_sharp/ - Docker image and build instructions

mattleibow:
@kekekeks why did you have link statically? I see that when you use fontconfig, that depends on freetype. If I link statically, then I have duplicate bits in the .so.

Maybe the HBB did not have a late enough version or something?

kekekeks:
If I understand correctly, even CentOS7 has too old version of libfreetype. While we can ship our own libfreetype which is up to date, we *have* to use system libfontconfig, since it has hardcoded config and font paths. BTW, it didn't introduce a big change in the file size.

mattleibow:
So if I am understanding right, skia uses the linked freetype for itself, but you/app uses the system library?
But, when skia is going to be doing font things, won't it also need the hardcoded system things?

kekekeks:
Skia will use a fresh version of libfreetype which is statically linked, so one applications will be able to use proper hinting, have support for recent changes in font formats, etc. At the same time it will use system libfontconfig to get system font locations and default settings.

mattleibow:
Aaaaahhhhh yeas. now is see - I continuously mix up libFONTCONFIG and libFREETYPE! It all makes sense now once you have the right libraries :P

PR #239, opened 2017-02-11
SKSvg - Fixed SVG not loaded when viewport / viewbox not specified

daniel-luberda:


mattleibow:
Hi @daniel-luberda, thanks for this. I saw this but I think we need to do this differently.
We can't just pull the viewbox from the width and height - the `ViewBox` property is specifically to see what was in the actual SVG.

There was a bug, and it is rather the division by zero when scaling. According to the SVG spec (http://www.w3.org/TR/SVG/coords.html#ViewBoxAttribute), if the value is set to 0 specifically, we shouldn't render:

> A value of zero disables rendering of the element.

However, if the value was not set at all, then nothing should happen:

> If specified, an additional transformation is applied to all descendants of the given element to achieve the specified effect.

I think we need to change this to rather check if it was specified, and then use that value. If it wasn't specified, then we just pretend that we have an infinite canvas (and no scale). This is what Chrome does as well.

mattleibow:
@migueldeicaza I think we should revert this and implement it according to the spec?

daniel-luberda:
Hi @mattleibow,

Sure, I understand. 👍  I didn't know that specification part, just wanted to bring back correct loading of some svg files that loaded properly and failed to do that after view port changes update.

mattleibow:
I also think we should do the clipping correctly: http://www.w3.org/TR/SVG/masking.html#OverflowProperty

Right now, it will just draw everything - which is fine since the `overflow` attribute isn't checked. Maybe we should also add logic to look at that and clip the canvas before painting?

mattleibow:
@daniel-luberda No problem. Thanks for the PR - do you want to have a go at making the changes for the viewbox?

Originally I was just going ahead with this as something similar to what NGraphics was doing, but then sometimes I reached a point where I needed to make decisions. Then I decided to just try and follow the spec as closely as I could since that is really what was expected. NGraphics is a bit free thinking :)

Issue #90, opened 2016-06-16
Linux support


tmds:
Does SkiaSharp work on Linux?

migueldeicaza:
SkiaSharp should, but the native supporting code needs to be compiled individually on each Linux distribution.

Sadly, NuGet was not designed to cope with this scenario, so there is not much that we can do to ship precompiled binaries.

tmds:
@migueldeicaza compilations could be provided for several linux flavors using the runtimes/RID feature: https://dotnet.github.io/docs/core-concepts/rid-catalog.html

mattleibow:
@tmds An ARM Linux was actually work with this guy: #18 
@migueldeicaza not too sure of all the limitations, but as the native dll is the only thing that needs switching... is it possible to do something like this for a few supported (popular) versions of Linux: https://github.com/mono/SkiaSharp/blob/master/binding/SkiaSharp.Desktop/SkiaSharp.dll.config

migueldeicaza:
The issue is not identifying the library, the issue is that creating a package with the binaries is incredibly cumbersome (you need to create a package with all available systems upfront, build the native bits, and then package all the bits together).

What should ideally happen is that these native binaries should be "addons" to the core package, and could be published independently of the core package.

KarolKolenda:
@migueldeicaza Any idea when makefiles for popular distros will be available? Is it going to happen this year or are we looking more at 2017?

kekekeks:
@migueldeicaza I think if you compile skia using sufficiently old distro (e. g. CentOS 6, Debian 5) it will just work everywhere. In the worst case you can just bundle dependencies (skia don't really need much: libfontconfig, libfreetype, libz, libpng12) and rely only on libc which is guaranteed to be present by LSB.

jameswalmsley:
I'm looking into the build.cake file and trying to work out how to convert it to build for linux.
Are there any attempts of pointers for me to start updating the build system.

mattleibow:
@jameswalmsley I would say you probably need to just get the externals building. These are grouped using the `Task ("externals-uwp")` steps. For the case of Linux, I would suggest following the Android style: `Task ("externals-android")` https://github.com/mono/SkiaSharp/blob/v1.53.0/build.cake#L674-L724

It pretty much follows the official docs by google: https://skia.org/user/quick/linux

It follows the pattern of:
 1. set global env vars
 2. set gyp defines for platform
 3. build `*.a` (using ninja) for platform
[...]

PR #241, opened 2017-02-13
[WIP] Optional GPU

mattleibow:
Linux build can now produce a non-GPU .so. Right now this part is working, but the configuration needs love. (basically, all the linux configuration needs love)

This must be merged first: https://github.com/mono/skia/pull/49

Related to https://github.com/mono/SkiaSharp/issues/90


kekekeks:
I'm pretty sure that it's possible to keep GPU support without referencing libGL/libGLU/libX11 by exposing `GrGLAssemble*Interface`: https://github.com/mono/skia/blob/xamarin-mobile-bindings/include/gpu/gl/GrGLAssembleInterface.h#L17

kekekeks:
That way skia will obtain references to OpenGL functions using user-provided glGetProcAddress

kekekeks:
BTW, GLX won't be present in "pure" Wayland/Mir environment, they use EGL instead.

mattleibow:
@kekekeks Thanks for the work, I will get it all merged. BTW, I already expose the `GRGLAssemble*Interface` members on the `GRGlInterface` type in SkiaSharp - that is actually how I do the ANGLE bits on UWP.

Doing it vial assemble is better than having no support at all. Good point.

PR #242, opened 2017-02-13
Don't link libGL, libGLU and libX11 at compile time

kekekeks:
To be used with https://github.com/mono/skia/pull/50

```
$ objdump -p bin/x64/libSkiaSharp.so.56.1.0 |grep NEEDED
  NEEDED               libpthread.so.0
  NEEDED               libdl.so.2
  NEEDED               libfontconfig.so.1
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2

```


mattleibow:
Thianks for this. Closing as it will be merged with https://github.com/mono/SkiaSharp/pull/241

Issue #260, opened 2017-03-08
Improvements for SKCanvasViewRenderer


gentledepp:
Hi!

I'ts been a while that we touched our SVG project, but now we're on the go again and it is amazing to see how SkiaSharp has evolved so far. Congratulations - you're doing a great job!

That said, we recently came across a problem when inheriting the SKCanvasViewRenderer as described in: https://github.com/mattleibow/SkiaSharpFormsRendererDemo

The problem is, that we need to create our own view on Android and iOS, as we need to override their "public override void TouchesBegan(NSSet touches, UIEvent evt)" and or "public override bool OnTouchEvent(MotionEvent ev)" methods.

Problem: when inheriting your SKCanvasViewRenderer, that viewtype is basically fixed.
So we refactored it a bit and created a "SKCanvasViewRendererBase<TFormsView, TNativeView>"

Note: I would love to create a PR for this, but it is sooooo cumbersome to get skia to build on my windows machine. If there was a script that did not only update the submodule but also build it, that would be soooooo great! (and bring a lot of contributors to your project)

In order to create our own renderer, we copied your SkiaSharp.Views.Forms.SKCanvasView (because its  interface ISKCanvasViewController is internal)
The only thing we changed is the "X" at the end of the classname and that the interface is now public
```

using System;
using Xamarin.Forms;

namespace SkiaSharp.Views.Forms
{
    public class SKCanvasViewX : View, ISKCanvasViewController
    {
        public static readonly BindableProperty IgnorePixelScalingProperty =
            BindableProperty.Create(nameof(IgnorePixelScaling), typeof(bool), typeof(SKCanvasViewX), default(bool));

        // the user can subscribe to repaint
        public event EventHandler<SKPaintSurfaceEventArgs> PaintSurface;

        // the native listens to this event
        private event EventHandler SurfaceInvalidated;
        private event EventHandler<GetCanvasSizeEventArgs> GetCanvasSize;

        // the user asks the for the size
        public SKSize CanvasSize
        {
            get
            {
                // send a mesage to the native view
                var args = new GetCanvasSizeEventArgs();
                GetCanvasSize?.Invoke(this, args);
                return args.CanvasSize;
            }
        }

        public bool IgnorePixelScaling
        {
            get { return (bool)GetValue(IgnorePixelScalingProperty); }
            set { SetValue(IgnorePixelScalingProperty, value); }
        }

        // the user asks to repaint
        public void InvalidateSurface()
        {
[...]

PR #269, opened 2017-03-22
Making the Forms views more extendable

mattleibow:
This is to provide support for custom native SkiaSharp views, but still use the existing renderer logic. #260


source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Native.Shared/SKCanvasViewRendererBase.cs
```diff
@@ -0,0 +1,129 @@
...
+
+		private void OnSurfaceInvalidated(object sender, EventArgs eventArgs)
+		{
+			// repaint the native control
+#if __IOS__
+			Control.SetNeedsDisplay();
+#else
+			Control.Invalidate();
+#endif
+		}
```
gentledepp: In our current beta, we had several crashes, because "OnSurfaceInvalidated" was called by a background thread.
Maybe it is better to post these on the main thread?

source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Native.Shared/SKGLViewRendererBase.cs
```diff
@@ -127,5 +139,22 @@ private void OnPaintSurface(object sender, SKNativePaintGLSurfaceEventArgs e)
...
+			{
+				this.controller = controller;
+			}
+
+			public void OnDrawFrame(SKSurface surface, GRBackendRenderTargetDesc renderTarget)
+			{
+				controller.OnPaintSurface(new SKPaintGLSurfaceEventArgs(surface, renderTarget));
+			}
+		}
+#endif
```
gentledepp: Why do you need this renderer construct? Is there a problem with the PaintSurface subscriptions? (just curious)

gentledepp:
Man it's really great that you're working on this! I owe you a beer.
Unfortunately, I did not (yet) have time to work on it - plus my skiasharp build works in powershell, but not in VS... always fighting the tools. ;-)

From what I can see you've nailed it. I left you some comments nevertheless. Thanks!

P.S + off topic.: Do you know someone that I could talk to to about the Xamarin.SignaturePad repo? It has been stale for 3 years now and there are vital PRs waiting...

source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Native.Shared/SKGLViewRendererBase.cs
mattleibow: The types are different. The incoming event is a native type in the SkiaSharp.Views.PLATFORM assembly, and I am constructing a SkiaSharp.Views.Forms assembly type. It is the programmer within that couldn't think of different names :)

gentledepp: oh ok 👍

mattleibow:
@gentledepp Thanks for the comments!

About the tools, what I do for dev on the sources other than the actual binding is to just install the NuGets. The project is set up so that the build system can run through and throw errors if there is anything out of place. But, when doing more rapid/focused dev, such as the forms bits, I just delete the project references and install the SkiaSharp and SkiaSharp.Views NuGet into the projects. That way I can dev everything on Windows in VS.

The first issue is that if I change something in the binding project, I want the build to fail if I don't update the other projects. But as you found out, this is not very friendly for everyone else. I need to figure out a way to have a CI-style project setup as well as a human-style setup.

source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Native.Shared/SKCanvasViewRendererBase.cs
mattleibow: This one is a bit tough... I think the two sides of the fence is to either let the user make sure that invalidate is done from the main/UI thread via `Xamarin.Forms.Device.BeginInvokeOnMainThread()`, or to let the framework do it automatically.

For most cases, the app will always be using the main thread, so this should not be an issue. Thus, the overhead for posting to the main thread on each call might be undesirable. However, if some event out of the user's control is causing this to be called, then we will have to wrap the invalidate.

In the case of your beta, are you able to determine if a call was being made from a background thread that you could control?

gentledepp: yeah of course. And I agree with you. We will apply the fix in our app logic ;-)

source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Native.Shared/SKGLViewRendererBase.cs
mattleibow: No wait. My bad.

The Android GL view does not use events. It is the surface-based system, which separates the view from the actual rendering via this render construct.

mattleibow: This pattern can be seen in the docs: https://developer.android.com/reference/android/opengl/GLSurfaceView.html

PR #289, opened 2017-05-02
Add a Xamarin Workbooks integration

bojanrajkovic:
The integration uses the new Xamarin Workbooks SDK and provides some
convenience renderers for SKImage, SKBitmap, SKSurface, SKPixmap, and SKColor to
aid in using those types in a workbook. With the integration loaded, those Skia
objects will be rendered directly in the viewport as images (in the case of the
first 4), or as a Workbooks color in the case of SKColor.


source/SkiaSharp.Workbooks/SkiaSharp.Workbooks.csproj
```diff
@@ -0,0 +1,61 @@
...
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{89736865-5272-4A42-A7FB-D802C6345502}</ProjectGuid>
+    <ProjectTypeGuids>{786C830F-07A1-408B-BD7F-6EE04809D6DB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <OutputType>Library</OutputType>
+    <RootNamespace>SkiaSharp.Workbooks</RootNamespace>
+    <AssemblyName>SkiaSharp.Workbooks</AssemblyName>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkProfile>Profile111</TargetFrameworkProfile>
```
mattleibow: You can sign the assembly by just adding:

```xml
<SignAssembly>true</SignAssembly>
<AssemblyOriginatorKeyFile>..\..\..\mono.snk</AssemblyOriginatorKeyFile>
```

You have to add an exception in the cake script if you don't want to sign...

bojanrajkovic: Ah, didn't realize signing was needed/desired. :) Can do.

mattleibow: For the workbooks, this is actually optional since other libraries won't use it. But I want to do this for the actual library so that other libraries can sign.

bojanrajkovic: Ah, actually, we can't sign, because our PCL isn't signed. :) I'll add an exception.

abock:
Squash these commits?

PR #293, opened 2017-05-10
Fixed dllmap entries for OSes with .so libs

kekekeks:


mattleibow:
Thanks for this @kekekeks! Next time I am in Russia, or you are in South Africa, I must buy you a beer!

Issue #282, opened 2017-04-26
DrawImage doesn't work properly with SKFilterQuality.High


Rungee:
The issue has already been posted on [google groups](https://groups.google.com/forum/#!topic/skia-discuss/7xcKq09kL0Y).

mattleibow:
I was not able to reproduce this.

I loaded my image like this (I load it via a bitmap because I am lazy):

```csharp

var assembly = GetType().GetTypeInfo().Assembly;
using (var stream = assembly.GetManifestResourceStream("SkiaSharpXamarinFormsDemo.image.png"))
{
	bitmap = SKBitmap.Decode(stream);
	image = SKImage.FromBitmap(bitmap);
}
```

Then, in my draw event, I draw:

```csharp
canvas.Clear(SKColors.Blue);
var paint = new SKPaint { FilterQuality = SKFilterQuality.High };
canvas.DrawImage(image, SKRect.Create(300, 300), paint);
```

Is it possible to provide a small test case or code that reproduces the problem?

![image](https://cloud.githubusercontent.com/assets/1096616/25498326/dd4976bc-2b87-11e7-8ed2-30b826df5b08.png)

Rungee:
In the post, there was a link to a [GitHub repository](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2FRungee%2FSkiasharp_Tests&sa=D&sntz=1&usg=AFQjCNFh3Lgk0gq78p0dtgT0le7sSBzZfg) containing a test case.

It is important to note that this bug happens only under certain conditions such as image scale or rect coordinates.

mattleibow:
All right! I am now able to reproduce it. It appears that when the image has a negative X and Y, the transparency is lost...

Not sure why yet, but I will have a look and see if this is a bug that was fixed in a later version of skia.

mattleibow:
@Rungee I think I have found the problem.

There appears to be a bug in skia that doesn't draw Unpremultiplied images correctly. When using `SKBitmap.Decode(path/stream)`, the image is decoded as Unpremultiplied. To decode as a Premultiplied image, use the overload that allows you to specify the desired alpha type:

```csharp
using (var codec = SKCodec.Create(new SKFileStream(path)))
{
    var info = new SKImageInfo(codec.Info.Width, codec.Info.Height);
    var image = SKImage.FromBitmap(SKBitmap.Decode(codec, info));
    
    // draw
}
```

[...]

PR #295, opened 2017-05-11
Always decode images with premultiplied alpha

mattleibow:
Decoded images will always be unpremul, but skia draws in premul.

This is a "fix" for #282. Not really a bug since we are preserving image data, but we will probably always have to use premul when drawing. So this change is rather an enhancement that a fix.


mattleibow:
Merging this because I am doing a major update to m58

PR #306, opened 2017-05-30
feat: render content type image

leonardors:


leonardors:
Render content type image

source/SkiaSharp.Svg/SkiaSharp.Svg.Shared/SKSvg.cs
```diff
@@ -193,6 +193,23 @@ private void ReadElement(XElement e, SKCanvas canvas, SKPaint stroke, SKPaint fi
...
+						var width = ReadNumber(e.Attribute("width"));
+						var height = ReadNumber(e.Attribute("height"));
+						var bytes = ReadBytes(e.Attributes().FirstOrDefault((arg) => arg.Name.LocalName == "href"));
+						var data = SKData.CreateCopy(bytes);
+						SKImage image = SKImage.FromEncodedData(data);
+						var rect = SKRect.Create(x, y, width, height);
+						canvas.DrawImage(image, rect);
+
+						data?.Dispose();
+						image?.Dispose();
```
mattleibow: Maybe wrap the `data` and `image` objects in a `using` statement - just in case of an exception.

leonardors:
Changed

leonardors:
changed

mattleibow:
I just want to tank you for this PR, and inform you that SVG support has moved from this repository to one that is more appropriate: https://github.com/mono/SkiaSharp.Extended

This new repository will make it easier to develop libraries that use SkiaSharp, but don't really have to be built with the main library. In addition, there will be no need to first build all the native code, just to test a fix for some SVG case.

Please open a new PR against this new repository, and we will be happy to merge it in.

PR #308, opened 2017-06-01
More output

leonardors:
SKBitmap Output


mattleibow:
I just want to tank you for this PR, and inform you that SVG support has moved from this repository to one that is more appropriate: https://github.com/mono/SkiaSharp.Extended

This new repository will make it easier to develop libraries that use SkiaSharp, but don't really have to be built with the main library. In addition, there will be no need to first build all the native code, just to test a fix for some SVG case.

Please open a new PR against this new repository, and we will be happy to merge it in.

Issue #311, opened 2017-06-08
Build libSkiaSharp.so and libHarfBuzzSharp.so entirely in make


mattleibow:
As mentioned by @hurricanehrndz (https://github.com/mono/SkiaSharp/issues/90#issuecomment-307112384), there is not that much needed to build on linux, and the dependency on `mono`, `dotnet` and `cake` is a bit overkill just to run `make`:

Reasons for current state (https://github.com/mono/SkiaSharp/issues/90#issuecomment-307124367):
 - I didn't know make or linux at the time of starting
 - building native skia was a bit herder when I started (few extra steps)

This is the previous linux issue: https://github.com/mono/SkiaSharp/issues/90

hurricanehrndz:
@mattleibow,
My only concern is getting libSkiaSharp.so built. I have been able to do that with a minimal make file. I will try and write a small bash script that ensures all the build dependencies are installed and that make is called. I will post something early next week.

andrew-boyarshin:
Why not just use Skia's GN build file?
GN is a robust solution used in Chromium project as well as some other projects, such as Skia. It's not that complicated to setup even on Windows. It also builds with harfbuzz.

mattleibow:
@andrew-boyarshin the reason you may want libSkiaSharp is that it comes with extras, as well as less :)
The basic skia is OK, if you are using C++, but for libSkiaSharp, we have a largish C API, and link out the rest. As a result, you get a tiny little, but comprehensive, package that can do CPU and GPU drawing. In addition, we have some extra APIs that help out with managed wrappers. Finally, we also provide a managed wrapper that can talk to it.

mattleibow:
@hurricanehrndz I may have something that might be good... I have just creating a makefile that will pop out the native bits. It is awful, should be taken out back and shot, but it works :)

I WILL be accepting ANY PRs, lol.

I will try get it up ASAP.

andrew-boyarshin:
@mattleibow I am afraid I still don't see your point or I just misunderstood the goal of this issue.

[mono/skia](https://github.com/mono/skia) has BUILD.gn (of course with modifications to take winrt platform addition and C source set expansions in account). There is also plenty of ```skia_use_*``` to control what is being built in. Why make the Makefile that would duplicate the GN build file just for one platform?

[...]

PR #323, opened 2017-06-23
Get a fully native linux build

mattleibow:
This is where I hopefully will get a way to build libSkiaSharp.so on any Linux, without having to install anything but the OS and C/C++ compilers.

This is to work on issue #311


mattleibow:
This is no longer relevant as the build is now done with GN and ninja directly: https://github.com/mono/SkiaSharp/wiki/Building-on-Linux

Issue #333, opened 2017-07-05
ToBitmap Fails for some images on Android


michaelstonis:
I can't exactly pin this one down, but with _some_ images on Android the call to ToBitmap will come back with null. It appears to be failing on the internal call to `success = skiaBitmap.CopyPixelsTo(ptr, bmp.Width * bmp.RowBytes);`

[Example image that fails](https://drive.google.com/file/d/0B9DZcZ_yPR7ETEg2a2NDbmlGNE0/view?usp=sharing)

aer127:
I'm having the same issue. Seems to be images whose height is greater than their width.

drobdymo:
I'm having the same issue. Seems to be images whose height is greater than their width.

drobdymo:
I have identified the problem. I will submit a pull request to fix the issue

drobdymo:
Pull request submitted.
https://github.com/mono/SkiaSharp/pull/340

mattleibow:
Thanks for the PR. I have merged it in and should be building soon: 
https://jenkins.mono-project.com/view/Components/job/Components-SkiaSharp/

michaelstonis:
I don't believe that this issue was resolved with the fix in 1.59.0. If you take a look at the `dstSize` parameter of `SKBitmap.CopyPixelsTo` it is not actually referencing or using it at all. With 1.59.0, my Android application now crashes with the following exception.

```[mono-rt]   at (wrapper managed-to-native) SkiaSharp.SkiaApi.sk_pixmap_read_pixels (intptr,SkiaSharp.SKImageInfo&,intptr,intptr,int,int) <0x00007>
[mono-rt]   at SkiaSharp.SKPixmap.ReadPixels (SkiaSharp.SKImageInfo,intptr,int,int,int) [0x00013] in <7398af18bbfa455c8a1e84f1a7eb063a>:0
[mono-rt]   at SkiaSharp.SKPixmap.ReadPixels (SkiaSharp.SKImageInfo,intptr,int) [0x00006] in <7398af18bbfa455c8a1e84f1a7eb063a>:0
```

I am going to see if I can cook up a different solution in the meantime. I will follow-up if I come up with anything useful.

mattleibow:
@michaelstonis Thanks for reporting this. This is actually a new issue.

[...]

PR #340, opened 2017-07-13
Issue #333, corrected formula for total bitmap bytes

drobdymo:
To calculate the total number of bytes for the bitmap, you must multiply HEIGHT by the number of bytes per row, not the width. If multiplied by the width, this can caused the copy to fail if the bitmaps height is greater than its width since not enough bytes will be copied.


drobdymo:
Also wanted to clarify, as stated in the commit message, this is a fix for Issue #333

mattleibow:
Thanks for the PR!

Issue #358, opened 2017-08-21
Expose GRContext on SKGLView


mattleibow:
When drawing to a GPU surface, you may want to draw off-screen, but still use the same context.

Maybe the views could have a property.

Original Source: https://forums.xamarin.com/discussion/99762/gpu-accelerated-off-screen-rendering

mattleibow:
Merged. Will be in 1.59.2 release.

PR #367, opened 2017-09-15
[WIP] Expose the Current `GRContext`

mattleibow:
This is the work for #358


source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.Shared/RendererTypes.cs
```diff
@@ -2,8 +2,8 @@
 
 namespace SkiaSharp.Views.Forms
 {
-	internal class GetCanvasSizeEventArgs : EventArgs
+	internal class GetPropertyValueEventArgs<T> : EventArgs
```
mattleibow: Just use a generic so that we don't have to create multiple types.

PR #369, opened 2017-09-15
[WIP] Make libSkiaSharp.so and libHarfBuzz.so Makefiles

### LINKED ISSUE #311: Build libSkiaSharp.so and libHarfBuzzSharp.so entirely in make

mattleibow:
This is a second PR for #323 to resolve #311

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,87 @@
+SKIA_PATH ?= $(shell readlink -f externals/skia)
+DEPOT_PATH ?= $(shell readlink -f externals/depot_tools)
+HARFBUZZ_PATH ?= $(shell readlink -f externals/harfbuzz)
+NATIVEBUILDS_PATH ?= $(shell readlink -f native-builds)
+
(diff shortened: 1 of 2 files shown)


mattleibow:
To make life easier for everyone, I moved code around and added a new target to the GN files. Now, the native `libSkiaSharp` build is done with GN and ninja directly: https://github.com/mono/SkiaSharp/wiki/Building-on-Linux

PR #458, opened 2018-02-28
Update SKBitmap docs for removal of Lock/UnlockPixels

seanofw:
LockPixels() and UnlockPixels() have been removed in the latest SkiaSharp, but nearly all documentation about SKBitmap on the Internet still refers to them (as does the official SKBitmap.xml document itself). After wasting an hour this morning wondering what was going on (since I've never used Skia before), and researching why even simple examples wouldn't compile, I added a note to the SKBitmap documentation that explains that these are no longer required operations, per discussion at https://bugs.chromium.org/p/skia/issues/detail?id=6481&desc=2 .  I don't know if the "note" I added is really the *correct* update for the documentation, but I sure would've saved me a lot of time today if it had been included.  Feel free to alter its wording.


mattleibow:
Thanks for fixing this.
Sorry that it wasted time - I try to catch it all, but sometimes things get through.

fitdev:
7 years later [the docs](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skbitmap?view=skiasharp-2.88) still refer to LockPixels...

mattleibow:
There is a note

> (Note: As of SkiaSharp 1.60.0, calls to SkiaSharp.SKBitmap.LockPixels and SkiaSharp.SKBitmap.UnlockPixels are no longer required, and they no longer exist as part of the API.)

But I think the earliest docs are 80 now, so probably time to update the texts for the new baseline.

PR #493, opened 2018-04-17
Add Tizen Support + More

mattleibow:
This includes all the changes from https://github.com/mono/SkiaSharp/pull/479 as well as quite a few other bits that improve the overall build process:
 - this includes the Tizen bits from the PR #479 
 - the config.h for harfbuzz is no longer from the repo, but an actual config
 - added a basic Tizen-only sample
 - the `SkiaSharp.HarfBuzz.NetStandard` and `SkiaSharp.views.Forms.NetStandard` projects should have died long time ago - they are now without the `.Netstandard` suffix


mattleibow:
build

mattleibow:
build

PR #494, opened 2018-04-18
iOS SKCanvasView - Disposing UIGraphics context to help GC

michaldobrodenka:
Adding using to help GC.

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.AppleiOS/SKCanvasView.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.AppleiOS/SKCanvasView.cs
@@ -85,17 +85,20 @@ namespace SkiaSharp.Views.iOS
 			if (designMode || drawable == null)
 				return;
 
-			var ctx = UIGraphics.GetCurrentContext();
+			using (var ctx = UIGraphics.GetCurrentContext())
+			{
+				// create the skia context
+				SKImageInfo info;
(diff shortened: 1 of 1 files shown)


mattleibow:
build

mattleibow:
Thanks for this PR!

michaldobrodenka:
I was desperately looking and auditing code for memory leak. I'm heavy using SkiaSharp SKCanvas in UITableView with many cell types. In the end, problem was elsewhere, turns out, that UITableView decides, that its reuse queue for given type is full, anbandone cells (but Dispose is not being called and in c# I have no notification on this). When UITableView needs items of that type again, and run out of items from reuse queue, creates new ones. 

So with right combination of cell types (some of them very big) and their presence in table, I had some cell abandoned by table reuse queue, but with big SKCanvas objects, cells in limbo, I don't know why Dispose wasnt called (cell is using only WeakReference to outside world).

Workaround for me is to keep set of created items and observe theirs state in GetCell - cell is Created, Visible or NotVisibleAnyMore. When table creates a new items in DequeueReusableCell instead of reusing, all existing created items with state NotVisibleAnyMore Dispose theirs SKCanvas objects.

Too bad, we don't have more control over reusing like in Android RecyclerView, where I had similar problem, but there we have recycle pool with given size for every type, we know when it is full so I can eagerly Dispose items when cache for item type is full etc.

PR #501, opened 2018-05-04
Building Android on Windows

mattleibow:
Since the update to m60, the Android native libraries should be able to be built on Windows.


mattleibow:
build

PR #514, opened 2018-05-10
Using SDK-style projects

mattleibow:


source/SkiaSharp.Build.props
```diff
@@ -6,6 +6,7 @@
     <SignAssembly>true</SignAssembly>
     <SkiaSharpBuildPropsImported>true</SkiaSharpBuildPropsImported>
     <MDocDocumentationDirectory>$(MSBuildThisFileDirectory)..\docs\en</MDocDocumentationDirectory>
+    <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
```
mattleibow: Since switching to SDK-style projects, this has to be added to to the changes in defaults: traditional projects never generated assembly info, SDK-style projects do.

binding/Binding/SkiaApi.cs
```diff
@@ -77,7 +77,7 @@ internal static class SkiaApi
 		private const string SKIA = "libSkiaSharp.so";
 #elif __MACOS__
 		private const string SKIA = "libSkiaSharp.dylib";
-#elif DESKTOP
+#elif __DESKTOP__
```
mattleibow: This should have been the correct way in the first place.

PR #522, opened 2018-05-17
Made some types public for 3rd-party Xamarin.Forms platforms

mattleibow:


mattleibow:
The build is succeeding, but I want to test a new platform to make sure that I have everything that is needed for more platforms.

… and then the docs must be added ...

mattleibow:
This is all ready in the new release: https://github.com/mono/SkiaSharp/releases/tag/v1.60.1

PR #524, opened 2018-05-22
Update to Xamarin.Forms v3.0

mattleibow:


mattleibow:
Moving this to v1.60.3 so we can get out a version that fixes a UWP bug (#536)

mattleibow:
This will be moved to a major skia update, namely m67 #588

PR #531, opened 2018-05-26
Some Improvements

mattleibow:
- Added some overloads to structs
 - Added an option to skip building externals to speed up builds that already have the externals built
 - Using a macro to help declare the mapping from the C++ type to the C type
 - Added correct file headers to the C/C++ files


mattleibow:
build

PR #544, opened 2018-06-05
[SKRect, SKRectI] Add convenience constructors

migueldeicaza:


mattleibow:
This looks good to merge, but I just wanted to check that you were aware of the `SKRect.Create (...)` overloads that take extra parameters. Technically, this is the same logic, but in two places.

binding/Binding/Definitions.cs
```diff
@@ -1348,6 +1356,15 @@ public SKRect (float left, float top, float right, float bottom)
 			this.bottom = bottom;
 		}
 
+		public SKRect (SKPoint location, SKSize sizeI)
```
aodpi: The second parameter is of type `SKSize` but it is named sizeI, maybe change it to size so it's more intuitive.

Issue #534, opened 2018-05-28
SKWidget for GTK# (on Linux) is rendering kind of slow


define-private-public:
### Description

This is a continuation of #462 .  That ticket's main issue has been solved, but it's been reported that the rendering of the `SKWidget`, on such events such as a window/widget resize are slow and produce flickering.  See for example:
![skiasharp-not-so-smooth](https://user-images.githubusercontent.com/4551425/40593046-89891d68-61f3-11e8-8973-80437eceb815.gif)

(This is using SkiaSharp 1.60.1 off of NuGet, and the linux `.so` files from this project's release page).

### Code

The code used is the same one as here: https://github.com/mono/SkiaSharp/blob/master/samples/Basic/Gtk/SkiaSharpSample/MainWindow.cs

The only different that I applied to (in that widget) is in the GUI designer, I set the Box Layout child properties of the `SKWidget` to `auto fill=false, expanding=true, fill=true`.

### Expected Behavior

Should be MUCH smoother, such as when using Cairo for rendering.  In fact, Skia typically renders faster than Cairo.

### Actual Behavior

[see above GIF]

define-private-public:
@mattleibow

mattleibow:
Thanks for this, I will have a look and see what can be done.

define-private-public:
Okay, did a little other testing.  It's something with the `PaintSurface` event, or whatever it connected in the pipeline.  I wrote a program that used a `StopWatch` and a `Timer` object to try to queue up a bunch of draws for an animation (~60 FPS).  The flickering is still happening.

The Skia drawing code itself is rendering in under a millisecond (very, very fast), but it's all of that update/event logic around it that makes it slow.

I was looking into creating an OpenGL context, have Skia render to it (or have it render to a texture and display that), but I'm not sure how easy that might be.  Mono's `GtkGLArea` bindings seem to be a tad old: https://github.com/mono/gtkglarea-sharp

define-private-public:
While this article is more Cairo focused, it might provide some insights:

https://cairographics.org/threaded_animation_with_cairo/

Smooth animation should be possible:
https://www.youtube.com/watch?v=Ri5R5eozUC8

define-private-public:
Hey, I'd like to report a bit of an update: I have been able to get some smooth animation working with SkiaSharp, and the `SKWidget`, but it still has some issues.  Here is my code:

```csharp
// Author:       Benjamin N. Summerton <https://16bpp.net>
// Description:  X
[...]

PR #549, opened 2018-06-14
Minor optimizations to the SKWidget drawing routine

define-private-public:
While these don't dramatically improve the performance of drawing on an SKWidget, this does get rid of an unnecessary function call.  These changes are made mainly to help mitigate issue #534

The unused `DoubleBuffered` variable was also removed.


define-private-public:
Hold on, I got the drawing VERY smooth.  I'll be posting an alternative PR soon.

PR #550, opened 2018-06-14
Improve rendering speed for Gtk#'s SKWidget

### LINKED ISSUE #534: SKWidget for GTK# (on Linux) is rendering kind of slow

define-private-public:
This gives the `SKWidget` for Gtk# a massive performance boost by moving
the creation of the `SKSurface` (and `SKImageInfo`) to the
`CreateDrawingObjects()` method (formally `CreatePixbuf()`).  This makes
it so that these objects aren't being made every render cycle.  That is
what was causing the slowdow.

This also gets rid of the flicker reported in #534; it's flicker from
window dragging & resizing is now on par with that of Cairo's.

Another minor optimization was made, removing the call to
`base.OnExposeEvent()`.  It's not needed and the extra fuction call can
siphon off CPU cycles we'd like to keep.  Some other optimizations can
be made, but most of them probably are very minor.

closes #534

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.Gtk/SKWidget.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.Gtk/SKWidget.cs
@@ -8,10 +8,11 @@ namespace SkiaSharp.Views.Gtk
 	public class SKWidget : global::Gtk.DrawingArea
 	{
 		private Gdk.Pixbuf pix;
+		private SKImageInfo imgInfo;
+		private SKSurface surface;
 
 		public SKWidget()
 		{
-			DoubleBuffered = false;
 		}
 
 		public SKSize CanvasSize => pix == null ? SKSize.Empty : new SKSize(pix.Width, pix.Height);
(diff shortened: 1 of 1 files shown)


mattleibow:
build

PR #771, opened 2019-01-31
Make sure to always clean the workspace

mattleibow:
Right now, the workspace is only cleared on success, making every failure clog the machine.


mattleibow:
build

PR #847, opened 2019-05-21
HarfBuzzSharp Allow loading face by table data

Gillibald:
This introduces support for http://www.manpagez.com/html/harfbuzz/harfbuzz-2.3.1/harfbuzz-hb-face.php#hb-face-create-for-tables
Currently, https://docs.microsoft.com/ja-jp/dotnet/api/skiasharp.sktypeface.openstream?view=skiasharp-1.68.0 is used to obtain https://docs.microsoft.com/ja-jp/dotnet/api/skiasharp.skstreamasset?view=skiasharp-1.68.0 and that is used to access the font file via `GetMemoryBase()` which currently fails on Linux

This new approach should work all the time because it just uses the table data that is already loaded into SKTypeface. It would be preferable if `SKTypeface.GetTableData` could return a pointer to the table instead of returning a copy.

https://chromium.googlesource.com/chromium/src/+/refs/heads/master/ui/gfx/harfbuzz_font_skia.cc just for reference how it is used.


Gillibald:
I think we have to change the way font files are used within unit tests. Tests fail when a file is still locked by another test.

Gillibald:
@mattleibow Could you have a look at this if you have time? Would be awesome if we could merge this in for next release.

Gillibald:
I just realized that this https://github.com/mono/SkiaSharp/blob/master/binding/HarfBuzzSharp.Shared/Definitions.cs#L54 should be removed. A hb_feature can have more values other than 0 and 1. That's just the regular case. I will fix this shortly. Outch

mattleibow:
@Gillibald With regards to the pointers for the table data, we could do something like this: https://github.com/mono/SkiaSharp/blob/master/binding/HarfBuzzSharp.Shared/Buffer.cs#L287

No need to introduce breaking changes on the return type.

binding/HarfBuzzSharp.Shared/Blob.cs
```diff
@@ -1,34 +1,16 @@
 ﻿using System;
 using System.IO;
-using System.Runtime.InteropServices;
 
 namespace HarfBuzzSharp
 {
-	// public delegates
-	public delegate void BlobReleaseDelegate (object context);
```
mattleibow: This will be a breaking change, so I think we should avoid this. It is a bit of a pain to have duplicate code, but it is what it is.

Gillibald:
@mattleibow I don't know which breaking change you mean. I introduced additional API.

binding/HarfBuzzSharp.Shared/Face.cs
```diff
@@ -24,6 +27,16 @@ public Face (Blob blob, int index)
 			Handle = HarfBuzzApi.hb_face_create (blob.Handle, index);
 		}
 
+		public Face (TableLoader tableLoader)
+			: this (IntPtr.Zero)
+		{
+			this.tableLoader = tableLoader;
+			tableLoadFunc = tableLoader.Load;
```
mattleibow: I will need to confirm this will work on iOS in Release mode on a device... I think the callbacks need to be a static method. That is why there is that static method with the `MonoPInvokeCallback` attribute. I don't think iOS supports the `UnmanagedFunctionPointer` like all the other platforms - might be an AOT thing.

Gillibald:
Ahh now I get it. You mean the changes to SkiaSharp API y we can just add additional API.

tests/Tests/HbBlobTest.cs
```diff
@@ -11,18 +11,18 @@ public class HbBlobTest : SKTest
...
 		}
 
 		[SkippableFact]
 		public void ShouldCreateFromStream()
 		{
-			using (var blob = Blob.FromStream(File.Open(Path.Combine(PathToFonts, "content-font.ttf"), FileMode.Open, FileAccess.Read)))
+			using (var blob = Blob.FromStream(File.Open(Path.Combine(PathToFonts, "Funkster.ttf"), FileMode.Open, FileAccess.Read)))
 			{
-				Assert.Equal(246224, blob.Length);
+				Assert.Equal(236808, blob.Length);
```
mattleibow: Rather add new tests, or use the `[SkippableTheory]` with `[InlineData]` to keep the old tests. I think the theory way might be better for future tests.

What was the reason for this particular change?

tests/Tests/HbFaceTest.cs
```diff
@@ -1,26 +1,24 @@
...
 using HarfBuzzSharp;
-
 using Xunit;
 
 namespace SkiaSharp.Tests
 {
 	public class HbFaceTest : SKTest
 	{
-		private static readonly Blob s_blob;
+		public static readonly Blob Blob;
```
mattleibow: I'm thinking it might be better to create a new class, such as `HbUtils` so that other tests don't directly reference this test class.

tests/Tests/HbBufferTest.cs
```diff
@@ -14,17 +14,20 @@ namespace SkiaSharp.Tests
...
 	{
+		private static readonly Font Font;
+
 		private const string SimpleText = "1234";
 
 		private const string SerializedSimpleText = "gid25=0+772|gid26=1+772|gid27=2+772|gid28=3+772";
 
+		static HbBufferTest()
+		{
+			Font = new Font(new Face(HbFaceTest.Blob, 0));
```
mattleibow: This can also go into the `HbUtils` class.

binding/HarfBuzzSharp.Shared/Blob.cs
Gillibald: Will revert it and introduce a duplicate

tests/Tests/HbBlobTest.cs
Gillibald: Parallel access to a font file was causing issues so I use a different file for that tests. Didn't know how to avoid that. We could add font files as resource to avoid this I guess.

tests/Tests/HbFaceTest.cs
Gillibald: Y that makes sense. Had this also in mind. This was also introduced to avoid parallel access to font files.

binding/HarfBuzzSharp.Shared/Face.cs
Gillibald: Not sure how to implement this for iOS with static function. Maybe it works the same as with BlobReleaseDelegate. We can store the context as user data aswell.

Gillibald:
Will revert breaking changes first. Then we can work on iOS support.

tests/Tests/HbBlobTest.cs
mattleibow: Ah. Yeah. I think there was an issue in skia somewhere that I noticed that it sometimes didn't close the file streams properly.

Gillibald:
@mattleibow I have reverted my changes to existing Api and also introduced iOS friendly callbacks. I am not sure if thats a decent solution. It seems to work but I think only until GC potentially moves the object.

mattleibow:
@Gillibald is this PR ready to be merged?

Gillibald:
Couldn't test this on iOS other than that it is ready to be merged

mattleibow:
Awesome. I think it is ready to be merged. I made a few changes to the way the native functions are managed. I made the whole internal bit be more useful. Still a bit annoying with iOS, but a bit better.
There is also the 2 cases where the `Feature` and `Script` types are constructed using the `FromString` method. I believe this is more of a "parse" type of thing, so I changed it to be `Parse` and added a `TryParse` to be more consistent with .NET. (let me know if they are not a parse-y thing).

Just waiting on PR #862 and then I'll merge and get it into the next preview for 1.68.1.

mattleibow:
Merged in #862

PR #862, opened 2019-06-16
[harfbuzz] Allow loading face by table data

mattleibow:
This includes the commits for #847 and a few extra - mainly to the managed-native interop with functions.


binding/HarfBuzzSharp.Shared/Blob.cs
```diff
@@ -1,35 +1,29 @@
...
-	// internal proxy delegates
-	[UnmanagedFunctionPointer (CallingConvention.Cdecl)]
-	internal delegate void hb_destroy_func_t (IntPtr context);
-
 	public class Blob : NativeObject
 	{
-		// so the GC doesn't collect the delegate
-		private static readonly hb_destroy_func_t destroy_funcInternal;
-		private static readonly IntPtr destroy_func;
+		private static readonly Lazy<Blob> emptyBlob = new Lazy<Blob> (() => new StaticBlob (HarfBuzzApi.hb_blob_get_empty ()));
```
Gillibald: What was the reason for this? The native side should always return the same instance. This is only needed for reference equality on managed side.

binding/HarfBuzzSharp.Shared/Face.cs
```diff
@@ -24,11 +41,35 @@ public Face (Blob blob, int index)
...
+				new Delegate[] { destroy, getTable });
+			Handle = HarfBuzzApi.hb_face_create_for_tables (table_func, ctx.NativeContext, DestroyFunction.NativePointer);
+		}
+
 		internal Face (IntPtr handle)
 			: base (handle)
 		{
 		}
 
+		public static Face Empty => new Face (HarfBuzzApi.hb_face_get_empty ());
```
Gillibald: If we want reference equality for this it should follow the same pattern as `Blob.Empty`

Gillibald:
I was just scared performance wise to do it that way. We should probably have special interop for iOS that is enabled by compiler variable. GetTableData is called quite frequently until all tables are loaded. Just a concern.

binding/HarfBuzzSharp.Shared/Face.cs
Gillibald: Should follow the same pattern as `Blob.Empty` for reference equality. This is a singleton on native side.

binding/HarfBuzzSharp.Shared/Blob.cs
mattleibow: This WAS for equality - I am using the `Lazy<T>` so it is only ever initialized once - i always return the same instance.

binding/HarfBuzzSharp.Shared/Face.cs
mattleibow: good one.

mattleibow:
> I was just scared performance wise to do it that way. 

I suppose hitting the dictionary is going to be a bit expensive at some point.

> We should probably have special interop for iOS that is enabled by compiler variable. 

100% I think I need to start looking at this at some point. But, I also need to measure this and see if this is really necessary - and then spend some time going that. We have been mostly lucky that is it usually for large object construction (bitmaps) or once-off (GL interfaces or font faces). I'm thinking we probably need to have a look at the other cases where we can make use of `Span<T>` instead of just converting to arrays in some of the bitmap/image types.

mattleibow:
@Gillibald Let me know if this is helpful at all: https://github.com/mono/SkiaSharp/pull/862/commits/ed6e2dea9b1844bd033de77478011b314f7ab878

I saw your comment about the `GetTableData` that copies the memory. I have basically added an overload that works directly with `IntPtr` - this means that YOU have to allocate the memory - but it allows you to create a `Blob` that wraps it, and no extra copies are made. See this test using it: https://github.com/mono/SkiaSharp/blob/dev/harfbuzz-updates/tests/Tests/SKShaperTest.cs#L77-L104

binding/HarfBuzzSharp.Shared/Blob.cs
Gillibald: We should compare handles instead of .Net instances. HarfBuzz will return the same singleton on every call to get_empty.

Gillibald:
This looks good and is the best we can get I guess.

Gillibald:
Just tested this with success 👍

binding/HarfBuzzSharp.Shared/Blob.cs
Gillibald:
```
private bool Equals(NativeObject obj)
{
    return Handle == obj.Handle;
}
```

mattleibow:
Just doing a last few tweaks to correct any breaking changes, parameter names and things. As soon as the build is done, I'll regen the docs and merge.

PR #863, opened 2019-06-19
Fix #849 : function ToSKPixmap create display artifacts with transparent bitmaps

### LINKED ISSUE #849: [BUG] ToSKBitmap() doesn't work properly
ToSKBitmap() causes artifacts and displays at the incorrect size in Windows Forms.

I'm able to generate the correct bitmaps with my own method below:

```
        public static SKBitmap BitmapToSKBitmap(Bitmap bitmap)
        {
            var info = new SKImageInfo(bitmap.Width, bitmap.Height);
            var skiaBitmap = new SKBitmap(info);

            for (int i = 0; i < bitmap.Width; i++)
            {
                for (int j = 0; j < bitmap.Height; j++)
                {
                    Color color = bitmap.GetPixel(i, j);
                    skiaBitmap.SetPixel(i, j, new SKColor(color.R, color.G, color.B, color.A));
                }
            }

            return skiaBitmap;
        }
```

Odirb:
This PR fixes #849

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.Shared/Extensions.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.Shared/Extensions.cs
@@ -187,6 +187,9 @@ namespace SkiaSharp.Views.Tizen
 				using (var tempBitmap = new System.Drawing.Bitmap(info.Width, info.Height, info.RowBytes, System.Drawing.Imaging.PixelFormat.Format32bppPArgb, pixmap.GetPixels()))
 				using (var gr = System.Drawing.Graphics.FromImage(tempBitmap))
 				{
+					// Clear graphic to prevent display artifacts with transparent bitmaps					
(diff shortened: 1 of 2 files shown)


mattleibow:
Thanks for this PR - merging as soon as the build finishes.

mlaily:
Hello !

Is there an estimation of when this fix will be available through nuget?

(And is there some kind of roadmap or public document where information about planned releases can be found? https://github.com/mono/SkiaSharp/projects/22 does not seem up to date, but I might be mistaken)

mattleibow:
@mlaily I can't say for sure at the moment, but we are getting ready to release soon.
If you need this now, you can download the latest preview from the CI feed: https://nugetized.blob.core.windows.net/skiasharp-eap/index.json

I am just waiting on a few more thumbs up before I push that out to NuGet.org.

Issue #842, opened 2019-05-14
[QUESTION] Copy a rectangular area within a bitmap


MichaelRumpler:
I want to copy a rectangular area in my SKBitmap to another position. The simple approach is:

    using (var canvas = new SKCanvas(Bitmap))
    {
        var sourceRect = new SKRect(sourceLeft, sourceTop, sourceLeft + width, sourceTop + height);
        var targetRect = new SKRect(targetLeft, targetTop, targetLeft + width, targetTop + height);
        canvas.DrawBitmap(Bitmap, sourceRect, targetRect);
        canvas.Flush();
    }

But this only works, if the rectangles don't intersect or if the sourceRect is below the targetRect.

As a workaround I check for this and copy the pixels myself starting at the end in this case.

	private void Copy(int sourceLeft, int sourceTop, int targetLeft, int targetTop, int width, int height)
	{
		using (var canvas = new SKCanvas(Bitmap))
		{
			var sourceRect = new SKRect(sourceLeft, sourceTop, sourceLeft + width, sourceTop + height);
			var targetRect = new SKRect(targetLeft, targetTop, targetLeft + width, targetTop + height);

			var sourceEnd = (sourceTop + height - 1) * Bitmap.Width + sourceLeft + width;
			var targetEnd = (targetTop + height - 1) * Bitmap.Width + targetLeft + width;

			if (sourceRect.IntersectsWith(targetRect)           // it they intersect
				&& sourceEnd < targetEnd)                       // and the source should be copied down
			{                                                   // then I can't use DrawBitmap
				int lineDelta = Bitmap.Width - width;
				IntPtr pixelsAddr = Bitmap.GetPixels();

				if (Bitmap.BytesPerPixel == 4)      // just to be sure, may have to work for other sizes too
				{
					unsafe
					{
						uint* ptr = (uint*)pixelsAddr.ToPointer();
						uint* sourcePtr = ptr + sourceEnd;
						uint* targetPtr = ptr + targetEnd;

						int x, y = height;
						while (y > 0)
						{
							x = width;
							while (x > 0)
							{
								x--;
								sourcePtr--;
								targetPtr--;
								*targetPtr = *sourcePtr;
							}
							sourcePtr -= lineDelta;
							targetPtr -= lineDelta;
							y--;
						}
					}
				}
			}
			else
			{
				canvas.DrawBitmap(Bitmap, sourceRect, targetRect);
			}
			canvas.Flush();
		}
	}

[...]

PR #865, opened 2019-06-20
Use Span<T> for data, bitmap data and text blobs

mattleibow:
This PR exposes several overloads and additional members that provide either `Span<T>` or `ReadOnlySpan<T>` for several types. Instead of requiring the use of `byte[]` or `IntPtr`, the new overloads will facilitate memory operations, but without requiring new allocations or just usability improvements.

The new APIs:
 - accessing pixel data from bitmaps, such as `SKBitmap`, `SKImage` and `SKPixmap`
 - accessing memory from data chunks, such as `SKData`, `Blob` and `Buffer`
 - accessing the destination memory for text blob run buffers, such as `SKRunBuffer` (and related)


mattleibow:
@Gillibald For naming, should it be Get<plural>Span or Get<singular>Span?

We could have GetPixelsSpan or GetPixelSpan. Or GetGlyphsSpan or GetGlyphSpan. Same with GetGlyphInfosSpan or GetGlyphInfoSpan.

The plurals sound like they have too many sss, but theoretically the wording is correct. or is it?

mattleibow:
Note: the .editorconfig is a bit of a lie but you can just leave the if statements for single-line logic without braces if it is a throw or a return. Not sure how to control that. For the rules in styling... not sure of them 😄  Half the time it is keep it the same or follow .net foundation or just use the mono style.

Gillibald:
Hmm hard to tell. Both are somehow correct. `GetGlyphSpan` returns a span of glyphs but if we have a property called `Glyphs` it makes sense to call it `GetGlyphsSpan` I am in favor of `GetGlyphSpan`

Gillibald:
Getting the code style right for this project is kinda hard for me. I just let Visual Studio produce the right format and hope it works.

mattleibow:
I think so too. Not a fan of the sssnakesss in the word. For the styling, I am trying to look for a rule that doesn't ask for braces on if statements.

mattleibow:
Found the braces thing: 

`csharp_prefer_braces = true:suggestion` -> `csharp_prefer_braces = true:silent`

mattleibow:
Also, sad choice, but all public types must start with SK :) That is for the new 3 buffer types.
And, is there a reason to make `SKTextBlobBuilderRunBuffer` public? Not sure if I want to do that yet. And, now that we have the new buffer types, we can keep that an internal thing - a sort of backing type for the classes.

mattleibow:
We can probably just have:

```csharp
[StructLayout (LayoutKind.Sequential)]
internal struct SKRunBufferInternal
{
	public IntPtr GlyphsBuffer;
	public IntPtr PositionsBuffer;
	public IntPtr Utf8textBuffer;
	public IntPtr ClustersBuffer;
}
```

Gillibald:
`SKTextBlobBuilderRunBuffer` doesn't need to be public, not sure why I did that. We can rename `RunBuffer` to `SKRunBuffer` even tho it is called `RunBuffer` by Skia

mattleibow:
I renamed the types, and it looks pretty good so far. We probably need some tests for this, just in case we break something later. The reason for skia not using the SK prefix is that it was actually a nested type - that is why I actually had the full name `SKTextBlobBuilderRunBuffer`. Not sure if I want to go with just a plain old `SKRunBuffer` - because google may introduce a new type with that name - they did it once or twice already. But, the fill name is going to get really long for the positioned buffers.

mattleibow:
Let me know if these changes work for you in the sense that the code looks intelligent when being consumed. What tests are you folks using in the app to make sure text is correct? We should probably get something like them in skia so we don't break you.

Something that I am trying to focus on is making sure the APIs are nice to use - in the beginning it was more of a "just get the features out" - now I want to make sure that when we expose an API it is actually usable.

mattleibow:
Oops, still need to rename the GetXxxSpan methods.

Gillibald:
I will think about proper unit tests.

In general this is used as a replacement for DrawPositionedText. By using Span<T> we can directly access the allocated buffer and avoid extra copies.

Gillibald:
The usage of new HarfBuzzSharp and new Span overloads is only part of current development so it is fine if things change. I had to find the needed API surface myself during development and changed a lot in that process.

Gillibald:
I recently found this: https://github.com/google/skia/blob/master/modules/skshaper/src/SkShaper_harfbuzz.cpp They seem to implement their own font funcs that call into Skia API to better share measurements between HarfBuzz and Skia.

We can't do that in managed code because that would involve UnmanagedFunctionPointer that needs special treatment for iOS

mattleibow:
We could write our own methods in native code and then expose them in managed code - this way we can create our own layer between - sort of what we did for the managed streams. 

But maybe I am not understanding, we are already doing a bit of these funcs with the new delegates in the last update you did? Sure, there is that annoying extra layer for iOS, but we _can_ use conditionals and maybe have a special compilation that only uses the intermediate layer that is _just_ for iOS. Is there a specific function that you think we won't be able to do?

On the other side, I am probably going to be binding this type when we update to the next major version - this will do away with the current sharper in the SkiaSharp.HarfBuzz package. Not only will there me less interop code, but we get all the performance of native code. And, Google maintains the code.

mattleibow:
I am renaming this PR so we can focus the scope here. If we are to merge this anytime soon, we might as well just get the bits that you need in and then merge for a release. We can always add more in another update.

And, I don't want to add anymore members for drawing text just yet as they are going away and will need to be replaced with managed implementations. I'd rather see what we are dealing with because we already have many members that are obsolete at this point.

Gillibald:
This is already a good start and covers good areas. 

The SkShaper maintained by Skia should be optional because it relies on ICU and that's a dependency you kinda want to avoid when you want a small application package.

Writing these font funcs in native code and exposing them in the binding could work I guess. The rest can be written in managed code.

I am not an expert in terms of UnmanagedFunctionPointer performance. If that is fine to use on a hot path we could write font funcs on managed side and have conditional compilation for iOS with the required fallback.

mattleibow:
@Gillibald how happy are you with the implementation for the spans on the text blob? Ready to merge?

Also, any final thoughts on the naming convention? Singular? Plural? Depends?

Gillibald:
@mattleibow For unit tests we could indeed just draw to a canvas and compare results. That's the only way to make sure everything works properly.

The naming looks right to me and we should keep that convention.

Will test this now.

binding/Binding/SkiaApi.cs
```diff
@@ -1908,24 +1908,24 @@ internal static unsafe class SkiaApi
...
+		public extern static void sk_textblob_builder_alloc_run_text_pos_h (sk_textblob_builder_t builder, sk_paint_t font, int count, float y, int textByteCount, sk_string_t lang, SKRect* bounds, out SKRunBufferInternal runbuffer);
 		[DllImport (SKIA, CallingConvention = CallingConvention.Cdecl)]
-		public extern static void sk_textblob_builder_alloc_run_text_pos (sk_textblob_builder_t builder, sk_paint_t font, int count, int textByteCount, sk_string_t lang, SKRect* bounds, out SKTextBlobBuilderRunBuffer runbuffer);
+		public extern static void sk_textblob_builder_alloc_run_text_pos (sk_textblob_builder_t builder, sk_paint_t font, int count, int textByteCount, sk_string_t lang, SKRect* bounds, out SKRunBufferInternal runbuffer);
 
 
 		// text blob builder run buffer
 		[DllImport (SKIA, CallingConvention = CallingConvention.Cdecl)]
-		public extern static void sk_textblob_builder_runbuffer_set_glyphs (ref SKTextBlobBuilderRunBuffer buffer, UInt16* glyphs, int count);
+		public extern static void sk_textblob_builder_runbuffer_set_glyphs (ref SKRunBufferInternal buffer, UInt16* glyphs, int count);
```
Gillibald: Do we still need these?

Gillibald:
Works perfectly in combination with HarfBuzz 👍

Can be merged

binding/Binding/SkiaApi.cs
mattleibow: Yes. We still need to allocate the memory on the native side.

Gillibald: Skia already allocates the memory with `SkiaApi.sk_textblob_builder_alloc_run_text` etc what do I miss here?

binding/Binding/SKTextBlob.cs
```diff
@@ -218,155 +221,199 @@ private void AddPositionedRunInternal (SKPaint font, ushort[] glyphs, SKPoint[]
...
+		{
+			if (glyphs.Count > Size)
+				throw new ArgumentOutOfRangeException (nameof (glyphs));
 
-		public IntPtr Utf8textBuffer => utf8text;
+			var glyphsSpan = GetGlyphSpan ();
 
-		public IntPtr ClustersBuffer => clusters;
+			for (var i = 0; i < Size; i++) {
+				glyphsSpan[i] = glyphs[i];
```
mattleibow: What is the performance of this manual copy? Could we use unsafe to copy?

binding/Binding/SkiaApi.cs
mattleibow: Ah, my bad. This is the set.

Just noticed that the manual copy might have performance implications https://github.com/mono/SkiaSharp/pull/865/files#r296629605

mattleibow: We can probably remove these.

binding/Binding/SKTextBlob.cs
mattleibow: We could modify the sk_textblob_builder_runbuffer_set_xxx to take a pointer and the offset and length to copy. Then we could do a memory copy if it is faster. In fact, how often will this set be a partial set? Would it be useful to expose a public api for copying only a subset?

Gillibald: I don't think you ever want to write a partial set. You allocate a buffer and want to write to it the full length. This can be done on native side for `public void SetGlyphs (ReadOnlySpan<ushort> glyphs)`

Gillibald: It makes sense to have multiple overloads

Gillibald: This is how I use it:

```
using (var builder = new SKTextBlobBuilder())
{
    var runBuffer = builder.AllocateRunPositioned(_paint,
    textRun.GlyphRun.GlyphIndices.Length, 0, null);
    var glyphs = runBuffer.GetGlyphSpan();
    var positions = runBuffer.GetPositionSpan();

    for (var i = 0; i < textRun.GlyphRun.GlyphIndices.Length; i++)
    {
        glyphs[i] = textRun.GlyphRun.GlyphIndices[i];
        positions[i] = textRun.GlyphRun.GlyphPositions[i];
    }

    var blob = builder.Build();
    canvas.DrawText(blob, 0, 0, _paint);
}
```

mattleibow: In that case, we can use the native method, since that is what is does. But, it might be the same thing to copy in unsafe. Need to run some profiling quickly.

mattleibow: Updated this to accept a Span<T> and then use the CopyTo member to just move the memory rather than looping through each item.

binding/Binding/SkiaApi.cs
mattleibow: Removed these in favor of the new Span<T>.CopyTo with better memory operations

mattleibow:
@Gillibald I changed (added new because backwards compat) the way you can create runs. Instead of allocating and then copying, you can just allocate and set in one line.

OLD:

```csharp
using (var builder = new SKTextBlobBuilder())
{
    var runBuffer = builder.AllocateRunPositioned(_paint,
    textRun.GlyphRun.GlyphIndices.Length, 0, null);

    var glyphs = runBuffer.GetGlyphSpan();
    var positions = runBuffer.GetPositionSpan();

    for (var i = 0; i < textRun.GlyphRun.GlyphIndices.Length; i++)
    {
        glyphs[i] = textRun.GlyphRun.GlyphIndices[i];
        positions[i] = textRun.GlyphRun.GlyphPositions[i];
    }

    var blob = builder.Build();
    canvas.DrawText(blob, 0, 0, _paint);
}
```

NEW:

```csharp
using (var builder = new SKTextBlobBuilder())
{
	builder.AddPositionedRun(_paint, textRun.GlyphRun.GlyphIndices, textRun.GlyphRun.GlyphPositions);
	var blob = builder.Build();
	canvas.DrawText(blob, 0, 0, _paint);
}
```

Because the `AddPositionedRun` takes an array/span, it is then passed directly to the run buffer's `SetGlyphs` and `SetPositions` methods. Each of those methods then use "native" code in the spans to copy the buffers.

Regardless of how you allocate the runs, you probably shouldn't loop through the code in the managed world as there is a lot of overhead with index checks and copying. This guy was wanting to copy memory for bitmaps and ran quite a few benchmarks: https://github.com/mono/SkiaSharp/issues/842 (https://github.com/MichaelRumpler/CopyBitmap)

As a result, it appears that the fastest way to copy memory is `Span<T>.CopyTo()`. And, this is what I am using under the hood of the various `SetXxx` methods of the runs. And, the code for the `CopyTo` method is very optimized with checks for the OS features.

PR #872, opened 2019-06-21
Update HarfBuzz 2.5.2

Gillibald:
- Windows [x] 
- Linux [x] 
- Tizen [x] 
- Android [x] 
- Mac [x] 

On Linux the compiler seems to have issues. Can we use clang to build HarfBuzz?
@mattleibow

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 [submodule "externals/harfbuzz"]
 	path = externals/harfbuzz
 	url = https://github.com/harfbuzz/harfbuzz.git
-	branch = 2.3.1
+	branch = 2.5.2
 [submodule "docs"]
 	path = docs
 	url = https://github.com/mono/SkiaSharp-API-docs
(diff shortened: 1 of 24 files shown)


mattleibow:
@Gillibald We can and probably should.
But, I see 2.5.2 has fixes for GCC 4.8 - so that might fix it.

Gillibald:
This is already HarfBuzz 2.5.2

mattleibow:
In that case, we can switch to clang. I don't think there is a real preference for compilers.

Gillibald:
Looks like the current Android compiler doesn't like C++11

Gillibald:
Linux and Android are blocked by compiler. For Mac Codex projects need and update I can't do.

mattleibow:
I switched to using clang for both Linux and Android, and then upgraded the Android build to C++11.
I also added the files to the xcode projects. Let's see how this goes...

mattleibow:
@Gillibald I think we are good to merge. Let me know if you spot anything that I missed.

Gillibald:
Looks good now 👍

PR #873, opened 2019-06-21
Spanify SKTextBlobBuilder

Gillibald:
Haven't tested this yet will do when build has finished


Gillibald:
This probably needs a rebuild. Some timeout happened.

mattleibow:
Merged this into the dev/spans branch to keep that up-to-date. Just open a new PR for more changes.

Also, the .editorconfig is a bit of a lie (or I don't know the rule) but you can just leave the if statements for single-line logic without braces if it is a throw or a return. Not sure how to control that.

PR #881, opened 2019-06-24
[HarfBuzzSharp] font funcs

Gillibald:


mattleibow:
@Gillibald looking good. I made a PR that will reduce the code to 2 lines per function: https://github.com/mono/SkiaSharp/pull/882

Feel free to merge that into your PR and then I can close it. If it works, we can merge with your PR. If it doesn't then you can adjust it on your PR to fit whatever is needed.

Gillibald:
Will have a look shortly 👍

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,553 @@
...
+			IntPtr firstUnicode, uint unicodeStride, out IntPtr firstGlyph, out uint glyphStride, IntPtr context)
+		{
+			var multi = DelegateProxies.Get<GetMultiDelegateDelegate> (context, out _);
+			var del = (NominalGlyphsDelegate)multi.Invoke (typeof (NominalGlyphsDelegate));
+
+			unsafe {
+				var unicodes = new ReadOnlySpan<uint> ((void*)firstUnicode, (int)unicodeStride);
+				var glyphCount = del.Invoke (null, null, count, unicodes, out var glyphs);
+
+				fixed (uint* glyphsPtr = glyphs) {
```
Gillibald: Not sure how to do this properly. If the delegate creates an array on native side everything is fine but if the array is managed this will break.

Gillibald:
@mattleibow Your PR helped a lot. This now covers all font funcs. I will write some unit tests tomorrow.

Gillibald:
Looks like something isn't right with the interop code or the way I wrote that failing unit test.

binding/HarfBuzzSharp.Shared/Font.cs
```diff
@@ -17,11 +18,34 @@ public Font (Face face)
...
+		}
+
 		internal Font (IntPtr handle)
 			: base (handle)
 		{
 		}
 
+		public Font Parent {
+			get {
+				var parent = HarfBuzzApi.hb_font_get_parent (Handle);
```
Gillibald: Introduce a backing field so we can store the .Net instance when a sub font is created by the binding

binding/HarfBuzzSharp.Shared/HarfBuzzApi.cs
```diff
@@ -118,6 +166,15 @@ internal class HarfBuzzApi
...
 		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
 		public extern static hb_bool_t hb_font_get_glyph_v_origin (hb_font_t font, hb_codepoint_t glyph, out hb_position_t x, out hb_position_t y);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static void hb_font_set_funcs (IntPtr font, hb_font_funcs_t klass, IntPtr font_data, ReleaseDelegateProxyDelegate destroy);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static hb_font_t hb_font_get_parent (hb_font_t font);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static hb_bool_t hb_font_get_glyph_name (hb_font_t font, hb_codepoint_t glyph, out StringBuilder name, out uint size);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static hb_bool_t hb_font_get_glyph_from_name (hb_font_t font, string name, int len, out hb_codepoint_t glyph);
```
Gillibald: UnmanagedType.LPStr attribute missing

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,553 @@
...
+		out int y, IntPtr context);
+	internal delegate int GlyphKerningProxyDelegate (IntPtr font, IntPtr fontData, uint firstGlyph,
+		uint secondGlyph, IntPtr context);
+	internal delegate bool GlyphExtentsProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		out GlyphExtents extents, IntPtr context);
+	internal delegate bool GlyphContourPointProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		uint pointIndex, out int x, out int y, IntPtr context);
+	internal delegate bool GlyphNameProxyDelegate (IntPtr font, IntPtr fontData, uint glyph, out StringBuilder name,
+		out uint size, IntPtr context);
+	internal delegate bool GlyphFromNameProxyDelegate (IntPtr font, IntPtr fontData, StringBuilder name,
```
Gillibald: change name to string with UnmanagedType.LPStr

binding/HarfBuzzSharp.Shared/HarfBuzzApi.cs
```diff
@@ -118,6 +166,15 @@ internal class HarfBuzzApi
 		public extern static hb_bool_t hb_font_get_glyph_h_origin (hb_font_t font, hb_codepoint_t glyph, out hb_position_t x, out hb_position_t y);
 		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
 		public extern static hb_bool_t hb_font_get_glyph_v_origin (hb_font_t font, hb_codepoint_t glyph, out hb_position_t x, out hb_position_t y);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static void hb_font_set_funcs (IntPtr font, hb_font_funcs_t klass, IntPtr font_data, ReleaseDelegateProxyDelegate destroy);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static hb_font_t hb_font_get_parent (hb_font_t font);
+		[DllImport (HARFBUZZ, CallingConvention = CallingConvention.Cdecl)]
+		public extern static hb_bool_t hb_font_get_glyph_name (hb_font_t font, hb_codepoint_t glyph, out StringBuilder name, out uint size);
```
Gillibald: Change StringBuilder to string with UnmanagedType.LPStr

binding/HarfBuzzSharp.Shared/Font.cs
```diff
@@ -211,6 +253,14 @@ public GlyphExtents GetGlyphExtents (uint glyph)
 			return new GlyphExtents ();
 		}
 
+		public string GetGlyphName (uint glyph) =>
+			HarfBuzzApi.hb_font_get_glyph_name (Handle, glyph, out var name, out _)
+				? name.ToString ()
+				: glyph.ToString ();
+
+		public uint GetGlyphFromName (string name) =>
+			HarfBuzzApi.hb_font_get_glyph_from_name (Handle, name, name.Length, out var glyph) ? glyph : 0;
```
Gillibald: `len = - 1`

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,553 @@
...
+		IntPtr firstGlyph, uint glyphStride, out IntPtr firstAdvance, out uint advanceStride, IntPtr context);
+	internal delegate bool GlyphOriginProxyDelegate (IntPtr font, IntPtr fontData, uint glyph, out int x,
+		out int y, IntPtr context);
+	internal delegate int GlyphKerningProxyDelegate (IntPtr font, IntPtr fontData, uint firstGlyph,
+		uint secondGlyph, IntPtr context);
+	internal delegate bool GlyphExtentsProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		out GlyphExtents extents, IntPtr context);
+	internal delegate bool GlyphContourPointProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		uint pointIndex, out int x, out int y, IntPtr context);
+	internal delegate bool GlyphNameProxyDelegate (IntPtr font, IntPtr fontData, uint glyph, out StringBuilder name,
```
Gillibald: No out parameters

binding/HarfBuzzSharp.Shared/HarfBuzzApi.cs
mattleibow: I don't think the size is an `out`. It looks to be calculated/assumed somewhere first: https://github.com/harfbuzz/harfbuzz/blob/2.5.2/src/hb-font.h#L507-L510

Gillibald:
@mattleibow This PR changes a lot more than just adding font funcs to the binding. I also added missing members and removed some overloads that take a `signed integer` as glyph id. HarfBuzz itself and all the binding methods return `an unsigned integer` as `glyph id` so it makes sense to remove these overloads. Skia and most other platforms use `ushort` as glyph id.

I also changed some methods to follow the Try-pattern because HarfBuzz itself works that way and it doesn't make sense to get a value from HarfBuzz all the time. This lets the consumer know that something wasn't found etc.

Gillibald:
@mattleibow Sry for the refactoring of some parts. I thought better change it now before the API | ABI is frozen. Hopefully, this doesn't take too long to review.

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,536 @@
...
+
+		public bool IsImmutable => HarfBuzzApi.hb_font_funcs_is_immutable (Handle);
+
+		public void MakeImmutable () => HarfBuzzApi.hb_font_funcs_make_immutable (Handle);
+
+		[MonoPInvokeCallback (typeof (FontExtentsProxyDelegate))]
+		private static bool FontExtentsProxyImplementation (IntPtr font, IntPtr fontData, out FontExtents extents, IntPtr context)
+		{
+			var multi = DelegateProxies.Get<GetMultiDelegateDelegate> (context, out _);
+			var del = (FontExtentsDelegate)multi.Invoke (typeof (FontExtentsDelegate));
```
mattleibow: Just pushed a commit to master that will compress these lines:

```csharp
 var del = GetMulti<FontExtentsDelegate> (context, out _);  
```

Gillibald: Will make the needed changes 👍

Gillibald:
@mattleibow I have implemented the delegate for font_data(user_data) it seems to work. Feel free to rename things etc.

mattleibow:
@Gillibald this is looking really good.

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,530 @@
...
+		IntPtr firstGlyph, uint glyphStride, IntPtr firstAdvance, uint advanceStride, IntPtr context);
+	internal delegate bool GlyphOriginProxyDelegate (IntPtr font, IntPtr fontData, uint glyph, out int x,
+		out int y, IntPtr context);
+	internal delegate int GlyphKerningProxyDelegate (IntPtr font, IntPtr fontData, uint firstGlyph,
+		uint secondGlyph, IntPtr context);
+	internal delegate bool GlyphExtentsProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		out GlyphExtents extents, IntPtr context);
+	internal delegate bool GlyphContourPointProxyDelegate (IntPtr font, IntPtr fontData, uint glyph,
+		uint pointIndex, out int x, out int y, IntPtr context);
+	internal unsafe delegate bool GlyphNameProxyDelegate (IntPtr font, IntPtr fontData, uint glyph, char* name,
```
mattleibow: Is there a reason why this is `char*` instead of `Span<T>`? Or even a `string`?

mattleibow:
Something is failing on Windows. I ran the Windows test 9 times on CI and they all failed. I also fails locally. Not sure what it is... Will keep looking.

mattleibow:
I think I may have discovered it. I build harfbuzz in Debug mode and got a nice exception:

> Run-Time Check Failure #0 - The value of ESP was not properly saved across a function call.  This is usually a result of calling a function declared with one calling convention with a function pointer declared with a different calling convention.

Just needed to add an attribute to the proxy delegates:

```csharp
[UnmanagedFunctionPointer (CallingConvention.Cdecl)]
```

binding/HarfBuzzSharp.Shared/FontFunctions.cs
Gillibald: I will change it to string. Wasn't sure at the beginning what HarfBuzz is expecting here.

mattleibow: I just pushed the changes - let me make some comments

binding/HarfBuzzSharp.Shared/DelegateProxies.font.cs
```diff
@@ -0,0 +1,223 @@
...
+
+	public delegate int GlyphKerningDelegate (Font font, object fontData, uint firstGlyph, uint secondGlyph);
+
+	public delegate bool GlyphExtentsDelegate (Font font, object fontData, uint glyph, out GlyphExtents extents);
+
+	public delegate bool GlyphContourPointDelegate (Font font, object fontData, uint glyph, uint pointIndex, out int x, out int y);
+
+	public delegate bool GlyphNameDelegate (Font font, object fontData, uint glyph, out string name);
+
+	public delegate bool GlyphFromNameDelegate (Font font, object fontData, string name, out uint glyph);
```
mattleibow: Both `GlyphNameDelegate` and `GlyphFromNameDelegate` now use strings.

binding/HarfBuzzSharp.Shared/DelegateProxies.font.cs
```diff
@@ -0,0 +1,223 @@
...
+		private static bool GlyphNameProxyImplementation (IntPtr font, IntPtr fontData, uint glyph, char* nameBuffer, int size, IntPtr context)
+		{
+			var del = GetMulti<GlyphNameDelegate> (context, out _);
+			var userData = GetFontData<FontUserData> (fontData);
+			var name = new Span<char> (nameBuffer, size);
+			var result = del.Invoke (userData.Font, userData.FontData, glyph, out var realName);
+
+			var nameSpan = realName.AsSpan ();
+			var bufferSpan = new Span<char> (nameBuffer, size);
+			nameSpan.CopyTo (bufferSpan);
```
mattleibow: Copy the managed string to the native buffer.

binding/HarfBuzzSharp.Shared/DelegateProxies.font.cs
```diff
@@ -0,0 +1,223 @@
...
+
+		[MonoPInvokeCallback (typeof (GlyphFromNameProxyDelegate))]
+		private static bool GlyphFromNameProxyImplementation (IntPtr font, IntPtr fontData, char* name, int len, out uint glyph, IntPtr context)
+		{
+			var del = GetMulti<GlyphFromNameDelegate> (context, out _);
+			var userData = GetFontData<FontUserData> (fontData);
+
+			var actualName = len < 0
+				? new string (name)
+				: new string (name, 0, len);
```
mattleibow: Create a string from the buffer, keeping in mind that a `-1` length is an null-terminated string.

binding/HarfBuzzSharp.Shared/Font.cs
```diff
@@ -163,54 +124,99 @@ public unsafe int[] GetVerticalGlyphAdvances (int count, IntPtr firstGlyph)
...
-			if (unicode < 0) {
-				throw new ArgumentOutOfRangeException (nameof (unicode), "Unicode must be non negative.");
+			var buffer = ArrayPool<char>.Shared.Rent (NameBufferLength);
+			try {
+				fixed (char* first = buffer) {
+					if (!HarfBuzzApi.hb_font_get_glyph_name (Handle, glyph, first, buffer.Length)) {
+						name = string.Empty;
+						return false;
+					}
+					name = new string (first);
```
mattleibow: Instead of checking for a null character, let the string constructor do the work.

binding/HarfBuzzSharp.Shared/Font.cs
```diff
@@ -163,54 +124,99 @@ public unsafe int[] GetVerticalGlyphAdvances (int count, IntPtr firstGlyph)
...
+
+		public unsafe string GlyphToString (uint glyph)
 		{
-			if (HarfBuzzApi.hb_font_get_glyph_extents (Handle, glyph, out var extents)) {
-				return extents;
+			var buffer = ArrayPool<char>.Shared.Rent (NameBufferLength);
+			try {
+				fixed (char* first = buffer) {
+					HarfBuzzApi.hb_font_glyph_to_string (Handle, glyph, first, buffer.Length);
+					return new string (first);
```
mattleibow: Avoid the manual loop, use the constructor.

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,270 @@
...
+
+		public void SetVerticalGlyphAdvanceDelegate (GlyphAdvanceDelegate del) =>
+			SetVerticalGlyphAdvanceDelegate (del, null);
+
+		public void SetVerticalGlyphAdvanceDelegate (GlyphAdvanceDelegate del, ReleaseDelegate destroy)
+		{
+			if (IsImmutable)
+				throw new NotSupportedException ($"{nameof (FontFunctions)} is immutable and can't be changed.");
+			if (del == null)
+				throw new ArgumentException (nameof (del));
```
mattleibow: These actually need to be null arg exceptions...

binding/HarfBuzzSharp.Shared/FontFunctions.cs
```diff
@@ -0,0 +1,270 @@
...
+			HarfBuzzApi.hb_font_funcs_set_glyph_h_advance_func (Handle, DelegateProxies.GlyphAdvanceProxy, ctx, DelegateProxies.ReleaseDelegateProxyForMulti);
+		}
+
+		public void SetVerticalGlyphAdvanceDelegate (GlyphAdvanceDelegate del) =>
+			SetVerticalGlyphAdvanceDelegate (del, null);
+
+		public void SetVerticalGlyphAdvanceDelegate (GlyphAdvanceDelegate del, ReleaseDelegate destroy)
+		{
+			if (IsImmutable)
+				throw new NotSupportedException ($"{nameof (FontFunctions)} is immutable and can't be changed.");
```
mattleibow: This should be InvalidOperationException.

Gillibald:
@mattleibow Thanks for cleaning this up. Looks so much better now. I can learn a lot from you. Just tried to get this in fast. Helps a lot to have a look from another perspective.

binding/HarfBuzzSharp.Shared/DelegateProxies.cs
```diff
@@ -79,7 +76,7 @@ private static void ReleaseDelegateProxyImplementation (IntPtr context)
 		private static IntPtr GetTableDelegateProxyImplementation (IntPtr face, Tag tag, IntPtr context)
 		{
 			var del = GetMulti<GetTableDelegate> (context, out var gch);
-			var blob = del.Invoke (null, tag, null);
+			var blob = del.Invoke (null, tag);
```
Gillibald: Do we want to use the UserDataDelegate here as well? If not we should remove the face parameter.

Gillibald:
https://github.com/harfbuzz/harfbuzz/releases/tag/2.5.3

I guess we have to retarget again

binding/HarfBuzzSharp.Shared/DelegateProxies.cs
mattleibow: Good catch. Let me fix that.

mattleibow:
Thanks for doing all the work for the functions. I probably wouldn't have been able to get them in as fast as you did - and I probably would have gotten it wrong. Nice to bounce all the ideas off you so that it is a better API.

mattleibow:
> I guess we have to retarget again

lol It has only been a few days and they are already pushing out releases :)

Gillibald:
Luckily it is trivial to update the submodule

mattleibow:
I had a look at the diff between 2.5.2 and 2.5.3 and there are no new files, so that is nice.

mattleibow:
@Gillibald let me know if these changes still work for the way you consume the APIs. We can merge if it is all good. I don't think I changed anything on the public API front, except the strings in the delegates.

Gillibald:
Haven't found the time yet to adopt these changes in my code. Should be fine but we can wait until I have made these changes if you want.

Gillibald:
Okay just tested against this PR and everything looks good 👍

PR #890, opened 2019-06-26
Remove the concept of "user data"

mattleibow:
**Description of Change**

Remove the concept of "user data" from the delegates.

 - User data is not really useful to C# and delegates because they can capture the data directly. This is more of a C/C++ thing.
 - User data adds complexity and overhead because new delegates have to be constructed to wrap the context.

### DIFF
--- a/binding/HarfBuzzSharp.Shared/Blob.cs
+++ b/binding/HarfBuzzSharp.Shared/Blob.cs
@@ -14,18 +14,19 @@ namespace HarfBuzzSharp
 		{
 		}
 
-		public Blob (IntPtr data, int length, MemoryMode mode, object userData, BlobReleaseDelegate releaseDelegate)
-			: this (Create (data, length, mode, userData, new ReleaseDelegate (releaseDelegate)))
+		[Obsolete ("Use Blob(IntPtr, int, MemoryMode, ReleaseDelegate releaseDelegate) instead.")]
+		public Blob (IntPtr data, uint length, MemoryMode mode, object userData, BlobReleaseDelegate releaseDelegate)
+			: this (data, (int)length, mode, () => releaseDelegate?.Invoke (userData))
 		{
 		}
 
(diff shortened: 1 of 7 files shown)


binding/HarfBuzzSharp.Shared/Blob.cs
```diff
@@ -80,18 +81,15 @@ public static Blob FromStream (Stream stream)
...
 		}
 
-		private static IntPtr Create (IntPtr data, int length, MemoryMode mode, object context, ReleaseDelegate releaseProc)
+		private static IntPtr Create (IntPtr data, int length, MemoryMode mode, ReleaseDelegate releaseProc)
 		{
-			var del = releaseProc != null && context != null
-				? new ReleaseDelegate ((_) => releaseProc (context))
-				: releaseProc;
-			var proxy = DelegateProxies.Create (del, DelegateProxies.ReleaseDelegateProxy, out _, out var ctx);
+			var proxy = DelegateProxies.Create (releaseProc, DelegateProxies.ReleaseDelegateProxy, out _, out var ctx);
```
mattleibow: This results in quite a bit less code and checks.

binding/HarfBuzzSharp.Shared/DelegateProxies.cs
```diff
@@ -6,14 +6,11 @@ namespace HarfBuzzSharp
...
 
-	public delegate Blob GetTableDelegate (Face face, Tag tag, object context);
+	public delegate Blob GetTableDelegate (Face face, Tag tag);
 
-	// bad choices.
-	// this should not have had the "blob" prefix
-	// it is a global dispose method, but we can't switch now
-	// it is a breaking change since it will become ambiguous
+	[Obsolete ("Use ReleaseDelegate instead.")]
 	public delegate void BlobReleaseDelegate (object context);
```
mattleibow: We score here with the name.

mattleibow:
This was merged in #881. Closing.

PR #891, opened 2019-06-26
Make sure everything is fine in Debug builds

mattleibow:
**Description of Change**

Make sure everything is fine in Debug builds.


Gillibald:
Probably an oversight of mine. If we shape a buffer or mutate it in some way we should call `GuessSegmentProperties` before we do that. Sry for that.

mattleibow:
That should be done by the caller, right?

Gillibald:
Yes. GuessSegmentProperties does not guarantee everything is configured right but will work most of the time. It can be left out if you know what you have to configure.

For example if you have a sequence of codepoints that represents RTL text GuessSegmentProperties should set the buffer's direction to RTL. You can set it explicitly or implicitly by GuessSegmentProperties. I hope this helps. 

I will add additional tests that cover "real world" scenarios. Current tests mostly just make sure that the call to some function works. I didn't think too much about the data that I use.

mattleibow:
Ah, I see. So you will have a check - just like in the native code - to throw an exception if the direction has not been set?

Gillibald:
We probably should mimic the assertions if we want this to be more user friendly

mattleibow:
Make that a separate PR as this one is just for build infrastructure.

Gillibald:
Will add exceptions shortly

mattleibow:
Moving the discussion to this issue (https://github.com/mono/SkiaSharp/issues/894) until we have a PR.

PR #892, opened 2019-06-27
Change native artifact name to match Azure Build.

### LINKED ISSUE #883: How do I get the native binaries for a Managed-only build?
I'm trying to do a managed-only build by following [the instructions here](https://github.com/mono/SkiaSharp/wiki/Building-SkiaSharp#preparation-1) since I wasn't able to get the externals/skia native builds to complete without compilation errors (in GrAHardwareBufferImageGenerator.cpp, line 217, GrBackendTexture backendTex: no matching constructor).

However, I don't know how to get the correct Azure Build IDs. If I go to the [Azure build page](https://dev.azure.com/xamarin/public/_build?definitionId=4) and pick one of the builds from SkiaSharp (Public) that appears to have built successfully with a green tick (e.g., [build 3348](https://dev.azure.com/xamarin/public/_build/results?buildId=3348)) ) and copy the buildId from the URL (e.g., 3348), and use that with the bootstrapper,

```
./bootstrapper.sh -t externals-download --azureBuildId=3348
```
(Also tried "3348" in quotes.)

I get the following error:

```
========================================
externals-download
========================================
An error occurred when executing task 'externals-download'.
Error: One or more errors occurred. (One or more errors occurred. (404 (Not Found)))
	404 (Not Found)

```

I found the AZURE_BUILD_URL that the SkiaSharp/build.cake and SkiaSharp/cake/BuildExternals.cake build scripts use to download the native binaries in a zip file. Using that URL with the buildId and "native" for the artifactName field, I get the following error:

```
{"$id":"1","innerException":null,"message":"Artifact native was not found for build 3348.","typeName":"Microsoft.TeamFoundation.Build.WebApi.ArtifactNotFoundException, Microsoft.TeamFoundation.Build2.WebApi","typeKey":"ArtifactNotFoundException","errorCode":0,"eventId":3000}
```

I've tried several of the other builds and also the ones from SkiaSharp.Extended (Public) and they all return the same error.

Is there a way I can figure out which builds contain the native binaries without guessing?

Thank you.

danien:
**Description of Change**

Changes the artifact name for the externals-download target from "native" to "native-default" to match the Azure Build.
This allows the [Build Preparation instructions](https://github.com/mono/SkiaSharp/wiki/Building-SkiaSharp#preparation-1) to work as documented instead of returning a 404 error.

**Bugs Fixed**

- Related to issue #889, #883 
- Fixes #883

**Changes**

Changes the artifact name from "native" to "native-default" to match the Azure Build Pipeline.
Also some minor refactoring of local string constants for the output path.

**Behavioral Changes**

Since the unzipped folder is named "native-default", it is renamed back to "native" so the rest of the build scripts can still work.

**PR Checklist**

- [ NA ] Has tests (if omitted, state reason in description)
- [ Y ] Rebased on top of master at time of PR
- [ Y ] Changes adhere to coding standard
- [ NA ] Updated documentation

### DIFF
--- a/cake/BuildExternals.cake
+++ b/cake/BuildExternals.cake
@@ -699,13 +699,17 @@ Task ("externals-download")
     if (string.IsNullOrEmpty (AZURE_BUILD_ID))
         throw new Exception ("Specify a build ID with --azureBuildId=<ID>");
 
-    var url = string.Format(AZURE_BUILD_URL, AZURE_BUILD_ID, "native");
+    var artifactName = "native-default";
+    var artifactFilename = $"{artifactName}.zip";
+    var url = string.Format(AZURE_BUILD_URL, AZURE_BUILD_ID, artifactName);
 
-    EnsureDirectoryExists ("./output");
-    CleanDirectories ("./output");
+    var outputPath = "./output";
+    EnsureDirectoryExists (outputPath);
+    CleanDirectories (outputPath);
 
-    DownloadFile(url, "./output/native.zip");
-    Unzip ($"./output/native.zip", $"./output");
+    DownloadFile (url, $"{outputPath}/{artifactFilename}");
+    Unzip ($"{outputPath}/{artifactFilename}", outputPath);
+    MoveDirectory ($"{outputPath}/{artifactName}", $"{outputPath}/native");
 });
 
 Task ("externals-angle-uwp")


cake/BuildExternals.cake
```diff
@@ -699,13 +699,17 @@ Task ("externals-download")
     if (string.IsNullOrEmpty (AZURE_BUILD_ID))
         throw new Exception ("Specify a build ID with --azureBuildId=<ID>");
 
-    var url = string.Format(AZURE_BUILD_URL, AZURE_BUILD_ID, "native");
+    var artifactName = "native-default";
```
mattleibow: Using variables - wise move.

danien: Thanks again for your help and for maintaining this project!

PR #897, opened 2019-06-28
Support for Vulkan context

tuccio:
**Description of Change**

Partial support for Vulkan backend, allowing to create GRContext, GrBackendRenderTarget and GrBackendTexture which can be used to render on/sampled from.

skia submodule changes refer to https://github.com/mono/skia/pull/60

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -9,6 +9,7 @@
 #tool nuget:?package=xunit.runner.console&version=2.4.0
 #tool nuget:?package=vswhere&version=2.5.2
 
+
 using System.Linq;
 using System.Net.Http;
 using System.Runtime.InteropServices;


binding/Binding/GRContext.cs
```diff
@@ -47,7 +47,21 @@ public static GRContext Create (GRBackend backend, IntPtr backendContext)
...
 				case GRBackend.Vulkan:
+					return GetObject<GRContext> (SkiaApi.gr_context_make_vulkan (backendContext));
+				default:
+					throw new ArgumentOutOfRangeException (nameof (backend));
+			}
+		}
+
+		public static GRContext Create (GRBackend backend, GRVkBackendContext backendContext)
+		{
+			switch (backend) {
```
mattleibow: I think we can improve this a fair bit by not using a switch.

Because this method only supports Vulkan, we might be better served with a:

```csharp
if (backend != GRBackend.Vulkan)
    throw new ArgumentOutOfRangeException (nameof (backend));
```

And we can probably do this for the other create methods as well. This will also help future-proof that enum.

tuccio: I guess it doesn't make sense to touch the obsolete one:

```
		[Obsolete ("Use Create(GRBackend, GRGlInterface) instead.")]
		public static GRContext Create (GRBackend backend, IntPtr backendContext)
```

This overload is also tricky, because I cannot easily implement anything that makes sense for Vulkan:
```
		public static GRContext Create (GRBackend backend)
```

mattleibow: We can just throw for Vulkan, and the exception message can say to use the overload.

mattleibow:
@tuccio I added a hook for the environment variable `VULKAN_SDK_HOME` so we can add a build that does Vulkan.

Are you able to add any tests so we can test things out? My Vulkan-foo is very low, so I might take a while to get up to the basics.

tuccio:
I added some test for context and surface creation, used SharpVk to setup vulkan objects, but it's not signed and ended up making a new project for my tests, couldn't think of a convenient solution.

binding/Binding/GRVkAlloc.cs
```diff
@@ -0,0 +1,16 @@
...
+	using VkDeviceMemory = VkNonDispatchableHandle;
+	using VkDeviceSize = System.UInt64;
+
+	[StructLayout (LayoutKind.Sequential)]
+	public struct GRVkAlloc
+	{
+		public VkDeviceMemory Memory;
+		public VkDeviceSize Offset;
+		public VkDeviceSize Size;
+		public uint Flags;
```
mattleibow: I would suggest making these fields private and then having properties to wrap them. This is both for consistency with the other types, and so that we can have more control over the order of fields and types of properties. Although this is not as fluid as skia, the API might change and we don't need to break backwards compatibility if we can avoid it. Unless if needs to be fields?

binding/Binding/GRVkAlloc.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Runtime.InteropServices;
+
+namespace SkiaSharp
+{
+	using VkDeviceMemory = VkNonDispatchableHandle;
+	using VkDeviceSize = System.UInt64;
```
mattleibow: Keep the usings out of the namespaces

binding/Binding/GRVkImageInfo.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Runtime.InteropServices;
+
+namespace SkiaSharp
+{
+	using VkImage = VkNonDispatchableHandle;
+	using VkImageTiling = System.Int32;
+	using VkImageLayout = System.Int32;
+	using VkFormat = System.Int32;
```
mattleibow: Keep usings out of namespaces

binding/Binding/GRVkImageInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+
+	[StructLayout (LayoutKind.Sequential)]
+	public struct GRVkImageInfo
+	{
+		public VkImage Image;
+		public GRVkAlloc Alloc;
+		public VkImageTiling ImageTiling;
+		public VkImageLayout ImageLayout;
+		public VkFormat Format;
+		public uint LevelCount;
```
mattleibow: Rather use properties with explicit backing fields.

binding/Binding/VkTypes.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace SkiaSharp
+{
+	public struct VkHandle
+	{
+		public IntPtr Handle;
```
mattleibow: Is there a reason for this as a separate struct as opposed to just an `IntPtr`? Also, are we using this type?

binding/Binding/VkTypes.cs
```diff
@@ -0,0 +1,14 @@
...
+	}
+
+	public struct VkNonDispatchableHandle
+	{
+		public ulong Handle;
```
mattleibow: Should this be a `UIntPtr`? But again, are we using this directly, or do we just need the handle and not the struct? I am not sure if we want to define vulkan types in skiasharp

tests/SkiaSharp.Desktop.Tests/SkiaSharp.Desktop.Tests.csproj
```diff
@@ -81,6 +84,9 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="SharpVk">
+      <Version>0.4.2</Version>
+    </PackageReference>
```
mattleibow: I see you created a new tests project, so should this have been reverted?

mattleibow: You added a new tests project, so should this file have been reverted?

mattleibow: But, looking again, I see the build succeeded and you have also signed the other project. Do you still need separate projects?

binding/Binding/VkTypes.cs
tuccio: https://vulkan.lunarg.com/doc/view/1.0.33.0/linux/vkspec.chunked/ch02s03.html

`Non-dispatchable handle types are a 64-bit integer type whose meaning is implementation-dependent, and may encode object information directly in the handle rather than pointing to a software structure.`

binding/Binding/VkTypes.cs
tuccio: IntPtr doesn't give any information on the semantics of the pointer, would rather have a type. It's currently unused because I didn't bind anything that required a vulkan handle (e.g. instance, device, allocators). Just seemed weird to define the non dispatchable handle and not the handle.

tests/SkiaSharp.Desktop.Tests/SkiaSharp.Desktop.Tests.csproj
tuccio: Not entirely sure about how signing works, but I can't seem to make it work with the other project.

binding/Binding/VkTypes.cs
tuccio: Sounds like `ulong` to me, why `UIntPtr`?

Issue #876, opened 2019-06-21
[FEATURE] Add a SkiaApi and SKAbstractManagedStream native abstraction


jeromelaban:
**Is your feature request related to a problem? Please describe.**
For the implementation of https://github.com/nventive/Uno.SkiaSharp, some internal changes have to be made to enable WebAssembly support. All of the changes need to be made specifically at the P/Invoke (parameter shuffling, types adjustments, custom delegates, etc...) while the rest of the code remains untouched.

This requires the creation of a wasm specific version of the SkiaSharp library, and therefore requires the creation of wasm-specific dependencies, such as SkiaSharp.Extended packages.

**Describe the solution you'd like**
SkiaSharp should provide a way to change the native backend. e.g. a `SkiaNativeBackend.Current` which provides an `ISkiaNativeBackend` implementation.

This will enable the support for WebAssembly, while minimally impacting the structure of SkiaSharp, until the situation of the packaging of WebAssembly support settles down (https://github.com/NuGet/Home/issues/8186).

mattleibow:
@jeromelaban I had a very, very quick think and this looks to be something that may work. Just tested it on Windows and it runs. Will it work for WASM?

https://gist.github.com/mattleibow/6f0cd3efd6152fb614aa0fd803867d25

mattleibow:
One important thing to note is that we initialize all the delegates, and then WASM can just override the specific ones that need adjustments. Is this useful? Will WASM override all of the methods and this is now just extra work?

jeromelaban:
@mattleibow most of the native functions in wasm don't change, so this could definitely work. I wondering about the performance impact of using an indirect call through a delegate, vs. a virtual invocation through an interface. I don't know what LLVM does in those cases, if there's some magic optimization that helps there :)

The case of the delegate registration for the `SKAbstractManagedStream` ends up being an abstraction for `Marshal.GetFunctionPointerForDelegate` which is not yet supported by mono-wasm (it has to go through emscripten's `addFunction`). The best way there would be to delegate the whole registration process to the platform specific implementation (the whole `SKAbstractManagedStream` type initializer).

Gillibald:
Is it likely to happen that we will get a full-featured Skia running via WASM? In my opinion, this is a special environment that needs a special build of Skia like CanvasKit. Cripleing other platforms just for one platform wouldn't be ideal. A user of SkiaSharp would expect to be able to use all features of SkiaSharp if there was a WASM build of it.

[...]

PR #900, opened 2019-07-01
Re-work the managed-native types

mattleibow:
**Description of Change**

 - Added `GCHandleProxy` to debug builds
    - this is used to track all `GCHandle` `Alloc` and `Free` calls to ensure that all allocations are freed.
    - added some unit tests to make sure this is actually enforced
    - as a result, several object are now freed correctly
 - Added `ISKReferenceCounted` and `ISKNonVirtualReferenceCounted` interfaces to represent the reference counting types used in the native library
    - this helps with automatically de-referencing objects
 - `SKAbstractManagedStream`, `SKAbstractManagedWStream` and `SKDrawable` have been re-written to use better delegates
   - instead of passing each of the delegates as parameters, they are now a struct that is passed as a single object
   - better for extensions (which there shouldn't be) and only a single static field on the type
   - removed the usage of `Marshal.GetFunctionPointerForDelegate`, which should help out with WASM (see #876)
   - the objects now only keep weak references, meaning that they can now be garbage collected
   - instead of trying to resolve the instances with a dictionary, a delegate is used and passed as "user context"
 - Moved some of the repetitive logic from the types into the base `SKObject` and `SKNativeObject`
    - some logic is automatically executed if the concrete type is `ISKReferenceCounted` or `ISKNonVirtualReferenceCounted`
    - with the more centralized logic and stricter patterns, better tests can be written to make sure all memory is freed correctly and timely
 - `SKData`, `SKFontManager` and `SKTypeface` now correctly prevent disposal of the "static" instances
 - `SKPaint` now references the `Shader`, `MaskFilter`, `ColorFilter`, `ImageFilter`, `Typeface` and `PathEffect` properties
    - this prevents accidental collection, or non-collection when the object goes out of scope
 - the `SKPath` iterators (`Iterator` and `RawIterator`) and op builder (`OpBuilder`) now correctly own and dispose their native objects
 - `SKRegion` objects are now disposed on the native side
 - `SKTypeface` construction from a `SKManagedStream` (via both `SKTypeface` and `SKFontManager`) now copy the contents of the .NET `Stream` into a native memory
    - typeface construction requires multiple seeks (previously, the stream was copied only if it was non-seekable)
    - it also requires "duplicating" the stream, which is not supported on .NET streams
       - duplicates or forks of a stream means that each of the streams need to be read concurrently from different locations
       - .NET streams can only have a single position
 - Updated the NuGets used for the tests
     - using the `Xunit.AssemblyFixture` and `Xunit.SkippableFact` NuGets instead of using the code directly
    - removed the `Xunit.Categories` NuGet as it was preventing tests from running

**Bugs Fixed**

- Related to issue #876 

**API Changes**
None visible.

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


tests/Tests/SKImageTest.cs
```diff
@@ -78,11 +79,8 @@ public void ReleaseImagePixelsWasInvoked()
 			using (var image = SKImage.FromPixels(pixmap, onRelease, "RELEASING!"))
 			{
 				Assert.False(image.IsTextureBacked);
-				using (var raster = image.ToRasterImage())
```
mattleibow: This is no longer supported as it is not valid C#/.NET code. You have to actually dispose of the same `SKImage` instance twice (or more depending).If the instance is the same, you cannot dispose it more time.

binding/Binding/SKDocument.cs
```diff
@@ -3,25 +3,19 @@
 
 namespace SkiaSharp
 {
-	public class SKDocument : SKObject
+	public class SKDocument : SKObject, ISKReferenceCounted
 	{
 		public const float DefaultRasterDpi = 72.0f;
 
+		// keep the stream alive for as long as the document exists
+		private SKWStream underlyingStream;
```
mattleibow: The streams were in danger of being disposed while the document was in progress. To fix this, we keep a reference to it to make sure the GC doesn't pick it up early.

mattleibow:
This PR has a big set of changes that may be breaking due to bug fixes:

The `SKAbstractManagedStream`, `SKAbstractManagedWStream` and `SKDrawable` no longer prevent the GC from collecting them. This means that if code no longer references them, they will be disposed. 
 - As far as I can tell, this should not be a problem for the streams as they are never kept around - they are just used for reading and writing and typically only need to live for as long as a single method, and then need to be disposed by the caller. The `SKTypeface` and `SKDocument` do keep it around for a bit, but then they also take ownership of the stream and keep a hard reference to the streams themselves. They will dispose the streams when they are disposed.
 - `SKDrawable` is never kept around and is entirely a user-controlled object. If it goes out of scope, skia doesn't have a reference anyway.

The `SKFontManager` and `SKTypeface` no longer use the managed streams (`SKManagedStream` or `Stream`) directly - they make a copy.
 - This is simply because skia streams can do things that are not possible for .NET - they can be read concurrently from different positions. If a `SKFileStream` or `SKMemoryStream` are passed, then the streams are not copied.
 - Further optimizations can be made in the case of a `MemoryStream` or `byte[]` to not actually copy but use GC pinning to get a handle to the managed data and work with pointers. But this can be done later so that this PR can be mreged and tested.

Issue #894, opened 2019-06-28
[FEATURE] Throw exceptions for invalid usage of harfbuzz


mattleibow:
**Is your feature request related to a problem? Please describe.**
As noticed in this PR #891, there are a few asserts that should be replicated in the managed implementations of harfbuzz.

**Describe the solution you'd like**
Add some exceptions and other checks to help out developers.

mattleibow:
As noticed by @Gillibald (https://github.com/mono/SkiaSharp/pull/891#issuecomment-506522150):

`GuessSegmentProperties` does not guarantee everything is configured right but will work most of the time. It can be left out if you know what you have to configure.

For example if you have a sequence of codepoints that represents RTL text `GuessSegmentProperties` should set the buffer's direction to RTL. You can set it explicitly or implicitly by `GuessSegmentProperties`.

I will add additional tests that cover "real world" scenarios. Current tests mostly just make sure that the call to some function works. I didn't think too much about the data that I use.

Some example:
```
// Direction == Invalid
// ContentType == Invalid
buffer.AddUtf8("H");
// ContentType == Invalid
// Direction == Invalid
buffer.GuessSegmentProperties();
// ContentType == Unicode
// Direction == LeftToRight
font.Shape(buffer);
// ContentType == Glyphs
// Direction == LeftToRight
```

PR #904, opened 2019-07-03
Update HarfBuzz to v2.5.3

Gillibald:
**Description of Change**

HarfBuzz is now updated to 2.5.3
UnicodeFunctions are now customizable via delegates like FontFunctions
Some overloads that except int as Unicode codepoint were removed for more consistency

**Bugs Fixed**

- Related to issue #894 

**API Changes**

Added: 
 
- `UnicodeFunctions (UnicodeFunctions parent)`
- `UnicodeFunctions UnicodeFunctions.Parent {get;}`
- `uint UnicodeFunctions.GetMirroring (uint unicode)`
- `bool UnicodeFunctions.TryCompose (uint a, uint b, out uint ab)`
- `bool UnicodeFunctions.TryDecompose (uint ab, out uint a, out uint b) `
- `voidUnicodeFunctions. SetCombiningClassDelegate (CombiningClassDelegate del, ReleaseDelegate destroy = null)`
- `void UnicodeFunctions.SetGeneralCategoryDelegate (GeneralCategoryDelegate del, ReleaseDelegate destroy = null)`
- `void UnicodeFunctions.SetMirroringDelegate (MirroringDelegate del, ReleaseDelegate destroy = null)`
- `void UnicodeFunctions.SetScriptDelegate (ScriptDelegate del, ReleaseDelegate destroy = null)`
- `void UnicodeFunctions.SetComposeDelegate (ComposeDelegate del, ReleaseDelegate destroy = null)`
- `void UnicodeFunctions.SetDecomposeDelegate (DecomposeDelegate del, ReleaseDelegate destroy = null)`

Changed:

 - `void FontFunctions.SetHorizontalFontExtentsDelegate (FontExtentsDelegate del, ReleaseDelegate destroy) => void FontFunctions.SetHorizontalFontExtentsDelegate (FontExtentsDelegate del, ReleaseDelegate destroy = null)` etc.

Removed:
- `void FontFunctions.SetHorizontalFontExtentsDelegate (FontExtentsDelegate del)` etc.

- `UnicodeCombiningClass UnicodeFunctions.GetCombiningClass (int unicode)`
- `UnicodeGeneralCategory UnicodeFunctions.GetGeneralCategory (int unicode)`
- `Script UnicodeFunctions.GetScript (int unicode)`

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


Gillibald:
I would like to remove all overloads that currently take `int` as codepoints. `UnicodeFunctions` returns `uint` as codepoint and HarfBuzz is general uses `uint` under the hood to it makes sense to remove these overloads.

mattleibow:
> I would like to remove all overloads that currently take `int` as codepoints. `UnicodeFunctions` returns `uint` as codepoint and HarfBuzz is general uses `uint` under the hood so it makes sense to remove these overloads.

Sounds good.

binding/HarfBuzzSharp.Shared/DelegateProxies.unicode.cs
```diff
@@ -0,0 +1,99 @@
...
+	public delegate Script ScriptDelegate (UnicodeFunctions ufuncs, uint unicode);
+
+	public delegate bool ComposeDelegate (UnicodeFunctions ufuncs, uint a, uint b, out uint ab);
+
+	public delegate bool DecomposeDelegate (UnicodeFunctions ufuncs, uint ab, out uint a, out uint b);
+
+	// internal proxy delegates
+
+	[UnmanagedFunctionPointer (CallingConvention.Cdecl)]
+	public delegate UnicodeCombiningClass hb_unicode_combining_class_func_t (IntPtr ufuncs, uint unicode, IntPtr context);
```
mattleibow: Make these internal so they aren't used accidentally by the users.

mattleibow:
I am having a look at the usage of `unsafe`, previously I was all "keep it hidden", but it just makes the code more verbose - and I have learned to not fear it. So, I am thinking that if it reduces the code a bit by adding it to the member definition, then let's do it.

Gillibald:
I changed it to match the style of recent additions but as you mentioned defining unsafe on the member definition looks much cleaner.

mattleibow:
I have been changed since working with you. I used to be a "safe" guy, but since seeing how awesome Spans and void* really are, I can see all the places where I was wrong. Thanks for putting up with me.

binding/HarfBuzzSharp.Shared/DelegateProxies.unicode.cs
Gillibald: Oops 👍

Gillibald:
@mattleibow Do you know why CI is always aborted?

mattleibow:
Not sure exactly... seems the builds are timing out after 2 hours - which means something got stuck. Rebuilding now, sometimes DevOps has hiccups.

mattleibow:
This looks good, just doing a last build before merging. Is this complete?

Gillibald:
Yes. This PR is complete.

Gillibald:
Should we add - static-libstdc++ only to the linker step in the makefile to avoid these warnings?

mattleibow:
> Should we add - static-libstdc++ only to the linker step in the makefile to avoid these warnings?

Ah, that might be good. Use the `LDFLAGS`.

mattleibow:
> Are you sure we can use a Tag here? I read only branches are valid.

I think I have always used the tag, not sure if it is valid... We probably don't need that at all - it was more to stop the docs tool from breaking - but that is now a separate repo. But, we can remove that in a separate PR - along with the other branches.

binding/HarfBuzzSharp.Shared/Buffer.cs
```diff
@@ -339,46 +271,39 @@ public void NormalizeGlyphs ()
...
-			if (end == -1) {
+			if (end == -1)
 				end = Length;
-			}
 
-			var builder = new StringBuilder (128);
-			var buffer = Marshal.AllocCoTaskMem (bufferSize);
-			var currentPosition = start;
+			using (var buffer = MemoryPool<byte>.Shared.Rent ())
+			using (var pinned = buffer.Memory.Pin ()) {
```
mattleibow: Looks like we can use the new memory pool types instead of allocating memory... Just need to now apply this to the rest of the repository, but this looks nice.

PR #917, opened 2019-07-23
Add the WPF backend for Xamarin.Forms

### LINKED ISSUE #486: Add a WPF renderer for Xamarin.Forms
> VS bug [#727963](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/727963)

mattleibow:
**Description of Change**

Add the WPF backend for Xamarin.Forms.

**Bugs Fixed**

- Fixes #486

**API Changes**

Added the `SKCanvasView` and `SKGLView` views.

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


entdark:
Nice addition, thank you.
Are you going to do anything with the hardcoded FPS problem in SKGLView?

mattleibow:
Are you talking about the hardcoded 60 fps? The other platforms have a 60 fps already, so this is to be consistent.

If there was a way to control the framerate, what would you use? What are your thoughts on just disabling the loop and using your own timer?

entdark:
Yes, about the hardcoded 60 fps.
I think using own timer is not a perfect but good idea.
Using OpenTK way where they set custom frequency with extra property is a better solution.

PR #922, opened 2019-07-30
Some of the colorspaces created are just references

### LINKED ISSUE #835: Getting the message "Not replacing existing, living, managed instance with new object."
I'm running locally on Windows and I find that if I run my application multithreaded, I get a lot of the debugger output:

Not replacing existing, living, managed instance with new object.

I don't seem to get this if I only run with a single thread. I haven't tracked down the reproduction any more than that yet.

The source code seems to point to this line:

https://github.com/mono/SkiaSharp/blob/master/binding/Binding/SKObject.cs#L124

But I don't yet understand when `RegisterHandle` gets called.

Is this a harmless message? Is there something I should change to prevent this from occurring?

mattleibow:
**Description of Change**

Some of the colorspaces created are just references
 - SRGB and sRGB Linear colorspaces are static and should never be disposed
 - In some cases, the native object gets created at the same native memory location, without the managed layer knowing. As a result, object disposal now first checks to make sure that the object in the lookup table is the current object.
 - **TODO: This needs to be added to the docs because it is unusual behaviour.**
**Bugs Fixed**

- Related to PR #900
- Fixes #835

**API Changes**

The public api is unchanged.

**Behavioral Changes**

In the past when references to a single sRGB colorspace was disposed, ALL the references (because there was only 1 reference really) were disposed. This PR makes sure that the "static" objects are not disposed.

Previously, if multiple native objects were created and disposed quickly, they may have the same memory location. This should no longer cause exceptions.

**PR Checklist**

- [ ] Updated documentation


nietras:
@mattleibow any chance this could be backported to 1.60.* release (before Index8 was removed)?

PR #929, opened 2019-08-14
HarfBuzzSharp 2.6.1

Gillibald:
**Description of Change**

This updates HarfBuzzSharp to version 2.6.1 and has added bindings for introduces OpenType Metrics API
This also adds some overloads to Buffer.cs that take an integer as codepoint to make the life easier when working with codepoints.

**API Changes**

Added: 
- Definitions.cs
-- `public enum OpenTypeMetricsTag : uint`
-- `public readonly structs OpenTypeMetrics`
- Font.cs
-- `public OpenTypeMetrics OpenTypeMetrics { get; }`
--- `public bool TryGetPosition (OpenTypeMetricsTag metricsTag, out int position)`
--- `public float GetVariation (OpenTypeMetricsTag metricsTag)`
--- `public int GetXVariation (OpenTypeMetricsTag metricsTag)`
--- `public int GetYVariation (OpenTypeMetricsTag metricsTag)`

**PR Checklist**

- [ ] Updated documentation


mattleibow:
Hehe, more updates to harfbuzz. Seems every week there is a new version 😆

Gillibald:
@mattleibow Could you update the XCode projects for me? I think when mac builds this is ready for review. Should `OpenTypeMetrics` be a nested struct within `Font.cs`

mattleibow:
Having a look now. Thanks for the PR.

tests/Tests/SKShaperTest.cs
```diff
@@ -58,7 +58,7 @@ public void CorrectlyShapesArabicScript()
 		{
 			var clusters = new uint[] { 4, 2, 0 };
 			var codepoints = new uint[] { 629, 668, 891 };
-			var points = new SKPoint[] { new SKPoint(0, 0), new SKPoint(28.375f, 0), new SKPoint(42.125f, 0) };
+			var points = new SKPoint[] { new SKPoint(0, 0), new SKPoint(28.25f, 0), new SKPoint(42, 0) };
```
mattleibow: This is interesting. It appears that the characters have shifted a tiny bit to the left by 0.125.

mattleibow:
@Gillibald Before I merge this, I see you changed the enum from the `Value = ('a' | 'b')` format into the int value. Was there a specific reason? Enums are constants, so they would be compiled to the integer value anyway. But, at we could see the string values.

Gillibald:
@mattleibow I think I changed it because Intellisence didn't work well with it if I remember correctly. If that isn't the case we should revert this and use bit shifting to produce the values. Too bad C# has no macros. To produce these values I wrote some code generation routine that reads the HarfBuzz headers and creates enums out of it.

Gillibald:
I have changed the enum definition to use the bit shifting representation

mattleibow:
Merged!
I want to get a release out before Christmas. Gonna try!

PR #938, opened 2019-09-04
Trying to fix devops

mattleibow:
This is just so that master can build, even though I left it in a green state. Seems the grass over there has died and is now a wasteland.


mattleibow:
This is the new place: https://github.com/mono/SkiaSharp/pull/939

PR #939, opened 2019-09-05
Getting DevOps to just build again

mattleibow:
Seems the world has moved on and left us with a broken build 😭 

A lot of the issues seem to be with .NET Core, one big one here: https://github.com/dotnet/sdk/issues/3376


VERSIONS.txt
```diff
@@ -4,8 +4,8 @@ mdoc.targets            release     5.7.4.9
 mdoc                    release     5.7.4.9
 harfbuzz                release     2.5.3
 skia                    release     m68
-xunit                   release     2.4.0
-xunit.runner.console    release     2.4.0
+xunit                   release     2.4.1
+xunit.runner.console    release     2.4.1
```
mattleibow: Not sure why xUnit stopped running properly, but this minor update fixed it.

global.json
```diff
@@ -1,4 +1,7 @@
 {
+  "sdk": {
+    "version": "2.2.199"
```
mattleibow: If this is not here, then the build will fail. Important that it be not the same as the actual .NET Core installed as that also fails. See more here: https://github.com/dotnet/sdk/issues/3376#issuecomment-529007198

scripts/azure-pipelines.yml
```diff
@@ -11,15 +11,20 @@ variables:
...
-  TIZEN_LINUX_PACKAGES: libxcb-xfixes0 libxcb-render-util0 libwebkitgtk-1.0-0 libxcb-image0 acl libsdl1.2debian libv4l-0 libxcb-randr0 libxcb-shape0 libxcb-icccm4 libsm6 gettext rpm2cpio cpio bridge-utils openvpn
-  MANAGED_LINUX_PACKAGES: ttf-ancient-fonts
-  MONO_VERSION: 5_18_1
-  XCODE_VERSION: 10.2.1
+  TIZEN_LINUX_PACKAGES: mono-complete msbuild libxcb-xfixes0 libxcb-render-util0 libwebkitgtk-1.0-0 libxcb-image0 acl libsdl1.2debian libv4l-0 libxcb-randr0 libxcb-shape0 libxcb-icccm4 libsm6 gettext rpm2cpio cpio bridge-utils openvpn
+  MANAGED_LINUX_PACKAGES: mono-complete msbuild ttf-ancient-fonts
+  MONO_VERSION_MACOS: 5_18_3
+  MONO_VERSION_LINUX: stable-xenial/snapshots/5.20.1.34
+  XCODE_VERSION: 10.3
+  DOTNET_VERSION: 2.2.108
```
mattleibow: We require a .NET Core 2.2.1xx for VS 2017 to work.

scripts/azure-pipelines.yml
```diff
@@ -11,15 +11,20 @@ variables:
...
   PREVIEW_LABEL: 'rc'
   NATIVE_LINUX_PACKAGES: curl mono-complete msbuild python git libfontconfig1-dev clang-3.8 make
-  TIZEN_LINUX_PACKAGES: libxcb-xfixes0 libxcb-render-util0 libwebkitgtk-1.0-0 libxcb-image0 acl libsdl1.2debian libv4l-0 libxcb-randr0 libxcb-shape0 libxcb-icccm4 libsm6 gettext rpm2cpio cpio bridge-utils openvpn
-  MANAGED_LINUX_PACKAGES: ttf-ancient-fonts
-  MONO_VERSION: 5_18_1
-  XCODE_VERSION: 10.2.1
+  TIZEN_LINUX_PACKAGES: mono-complete msbuild libxcb-xfixes0 libxcb-render-util0 libwebkitgtk-1.0-0 libxcb-image0 acl libsdl1.2debian libv4l-0 libxcb-randr0 libxcb-shape0 libxcb-icccm4 libsm6 gettext rpm2cpio cpio bridge-utils openvpn
+  MANAGED_LINUX_PACKAGES: mono-complete msbuild ttf-ancient-fonts
+  MONO_VERSION_MACOS: 5_18_3
+  MONO_VERSION_LINUX: stable-xenial/snapshots/5.20.1.34
```
mattleibow: We have to uninstall and reinstall an older version of mono on linux.

scripts/azure-pipelines.yml
```diff
@@ -11,15 +11,20 @@ variables:
...
+  XCODE_VERSION: 10.3
+  DOTNET_VERSION: 2.2.108
   # CONFIGURATION: 'Release'
   VM_IMAGE_WINDOWS: vs2017-win2016
   VM_IMAGE_MAC: macos-10.14
   VM_IMAGE_LINUX: ubuntu-16.04
 
+  # To speed up builds
+  DOWNLOAD_EXTERNALS: '6228'
```
mattleibow: In order to get any work done, There is now a way to tell DevOps to rather download the externals instead of building them. A build takes 1.5-2 hours; a download is 10 minutes.

scripts/azure-templates-bootstrapper.yml
```diff
@@ -79,9 +110,13 @@ jobs:
       # build
       - ${{ if endsWith(parameters.name, '_windows') }}:
         - powershell: .\scripts\retry-command.ps1 -RetryCount ${{ parameters.retryCount }} { .\bootstrapper.ps1 -t ${{ parameters.target }} -v $env:VERBOSITY -c ${{ coalesce(variables.CONFIGURATION, 'Release') }} ${{ parameters.additionalArgs }} }
+          env:
+            JavaSdkDirectory: $(JAVA_HOME)
```
mattleibow: Some versions of Xamarin.Android do not look at the JAVA_HOME envvar, so just make sure we set it. See https://github.com/xamarin/xamarin-android-tools/pull/74

scripts/download-file.ps1
```diff
@@ -0,0 +1,44 @@
...
+# This was modified from StackOverflow 
+# https://stackoverflow.com/questions/45574479/powershell-determine-new-url-of-a-permanently-moved-redirected-resource
+
+Param (
+    [Parameter(Mandatory, ValueFromPipeline)] [Uri] $Uri,
+    [string] $OutFile,
+    [int] $MaxRedirections = 50 # Use same default as [System.Net.HttpWebRequest]
+)
+
+process {
```
mattleibow: This file is not used anymore, but was needed during testing because PowerShell can't redirect when it gets a 3xx status code.

tools/packages.config
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-    <package id="Cake" version="0.33.0" />
+    <package id="Cake" version="0.34.1" />
```
mattleibow: Seems that cake 0.33 can't find VS 2019, so we need to move to 0.34.1. We are not using VS 2019 yet, but this is just so that we don't have to struggle when we do update.

mattleibow:
Fixed in #922

PR #986, opened 2019-10-22
Update to a much later version of skia (m80)

### LINKED ISSUE #711: Migrate to SKFont
### Description

Google is deprecating the old ways of drawing text as they are not useful for non-English, non-linear text.

According to them, `SKFont` is the new text paint, and `SKTextBlob` is the new string:
https://groups.google.com/forum/#!topic/skia-discuss/qRARbUW2qhw

> VS bug [#737626](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/737626)

mattleibow:
**Description of Change**

Update to the latest (m80 at this time) version of skia.

**This has been published that to the internal feed as 2.80.0-pr.986.37**

**Tasks**

- [X] Merge native skia
- [X] Get native skia building
    - [X] Hacking out the problems so it compiles
    - [x] Fixing the problems properly
    - [x] Bind any new members that are essential
- [x] Update the managed bindings
    - [x] Hacking out the problems so it compiles
    - [x] Fixing/adding unit tests
    - [x] Fixing the problems properly
    - [x] Bind any new members that are essential
- [x] Update all source libraries _(this shouldn't be too much work as these APIs are the same)_
    - [x] SkiaSharp.Views _(there are a LOT of platforms)_
    - [x] SkiaSharp.Views.Forms
- [ ] Update all samples _(there are even more samples)_
    - [ ] Update basic samples
    - [ ] Update feature samples

**Bugs Fixed**

 - Possibly a few...
 - Fixes #711

**API & Behavioral Changes**

Changes so far (excluding the changes listed in https://github.com/mono/skia/pull/64):

 - New minimum versions
    - Android API 16 (Jelly Bean)
    - Ubuntu 16.04 with later Clang versions (only for the default builds, but you can always build some ancient Linux if you build your own Clang)
    - UWP 10.0.16299 (Fall Creators Update)
 - Using Docker images on CI to build Linux
    - You can still build on Linux without Docker (you just need to manually install the tools)
 - HarfBuzz updated to v2.6.2
 - Targeting the newer tools
    - mono 6.4.0
    - Xcode 11
    - .NET Core 3.0
    - VS 2019
 - TODO


native-builds/libHarfBuzzSharp_windows/libHarfBuzzSharp.vcxproj
```diff
@@ -185,6 +187,7 @@
     <!--HB_BASE_RAGEL_GENERATED_sources-->
     <ClInclude Include="..\..\externals\harfbuzz\src\hb-buffer-deserialize-json.hh" />
     <ClInclude Include="..\..\externals\harfbuzz\src\hb-buffer-deserialize-text.hh" />
+    <ClInclude Include="..\..\externals\harfbuzz\src\hb-number-parser.hh.hh" />
```
Gillibald: Should be `..\..\externals\harfbuzz\src\hb-number-parser.hh`

mattleibow: ah, gotcha - let me fix that

mattleibow: More and more is green! I think at this point it is just UWP and the use of non-UWP members. But that can be easily fixed (I hope)

mattleibow: So I think that was just me including GDI on UWP - so I decided for this major version to also remove GDI from the Windows build - should be a smaller dll now. We got the newer and better DirectWrite.

Gillibald: Y should be fine to exclude it from the UWP build. If someone wanted to load GDI fonts one could do that on managed side.

Gillibald: To make things easier it is very possible to have a custom font manager that is customizable via callbacks from managed side. So it is possible to avoid these dependencies. The manager for DWrite, GDI, FontConfig, etc could then be written in C#. Sure that has some drawbacks but it would be possible to use SkiaSharp in more restricted environments.

So excited about this major update 👍

mattleibow:
Due to `SKCanvas.DrawTextOnPath` being dropped from the native code, it has to be added back. Thanks to @Ziriax, we have a good implementation in https://github.com/mono/SkiaSharp/pull/1198

binding/Binding/SKCanvas.cs
```diff
@@ -690,7 +735,9 @@ public void DrawTextOnPath (IntPtr buffer, int length, SKPath path, float hOffse
 			if (paint == null)
 				throw new ArgumentNullException (nameof (paint));
 
-			SkiaApi.sk_canvas_draw_text_on_path (Handle, (void*)buffer, (IntPtr)length, path.Handle, hOffset, vOffset, paint.Handle);
+			// TODO
```
mattleibow, line 738: Implementation in https://github.com/mono/SkiaSharp/pull/1198

Ziriax:
Is it expected that I can't run the unit tests? 

I get undeterministic crashes with the latest version #dbfa12b644099e78cc4e46e226daba332f204f01 in `dev/update-skia` branch. 

Since the downloadable externals don't seem to be compatible, I compiled the native bits myself. 

I'm on Windows 10 x64.

I use `dotnet test` to run the .NET Core unit tests.

Ater some random number of tests I get ```The active test run was aborted. Reason: Test host process crashed Test Run Aborted.```. Sometimes I immediately get an access violation exception.

I'm running this on an AMD CPU with 16 cores / 32 hyperthreads, so maybe this triggers a race condition?

Or maybe I compiled the native bits incorrectly (although I didn't get any errors)

mattleibow:
There is sometimes a concurrency issue. This is also in master - which is why there is no stable of 1.68.2 yet.

Is there a stack trace or anything to indicate which one?

Ziriax:
> There is sometimes a concurrency issue. This is also in master - which is why there is no stable of 1.68.2 yet.
> 
> Is there a stack trace or anything to indicate which one?

My specific problems were related to #1227 

I provided a PR #1227  for this.

These were not multi-threaded related, except the WGL deadlock I always have on all my machines, but that might be an driver bug on Windows...

Gillibald:
I gave this a try with current Avalonia master and it worked out well with software rendering. Only `SKPaint.BreakText` didn't work but that's not an issue. I had no luck with the GPU backend.

mattleibow:
> I gave this a try with current Avalonia master and it worked out well with software rendering. 

AWESOME! Was this the preview NuGet or did you build your own?

> Only `SKPaint.BreakText` didn't work but that's not an issue. 

Yes, This and a few others members are no longer around. But this will be added back soon. Already have a PR for another missing member #1198 

> I had no luck with the GPU backend.

Crashes? Exceptions? No rendering?

Gillibald:
My own build. I tried the m83 branch. That probably needs some updates to the enum mappings. I will try the older branch soon.

Gillibald:
Had no issues building with Clang 10

mattleibow:
@Gillibald  yeah, the m38 was just hacked together to test the `SKPlatformEffect` on the GPU. Very specific case. Seems we might be getting a nice fragment shader support in that one. Looks to be "working" so far, just needs a bit more love on the API. And then fix the enums since they changed things again there... But, it compiled without too much hassle.

And nice to see clang 10 is working.

Gillibald:
In my opinion SKFont should only measure and break spans of glyphs. Supplying a string to these methods leads to errors. This only works for text that doesn't need extra shaping when you supply a string.

mattleibow:
@Gillibald I agree. I just have to add it back for backwards compat 😢 

However... I can just keep SKFont to be glyph-only... 

Doesn't the `MeasureText` fall into the same category as `GetGlyphWidths` and `CountGlyphs`? They also take strings.

Finally, even though you and I know that we need to get glyphs, many people just want to use simple .NET strings as text...

Gillibald:
Y all these methods should follow this logic. I consider SKFont a low level component. SKPaint can expose overloads that take a string. SKFont doesn't need that and should do better. Maybe we should document it somehow that this only works for basic scenarios and link to the SkShaper component.

mattleibow:
@Gillibald 
> This also doesn't handle surrogate pairs

I think it does, which is quite nice: https://github.com/google/skia/blob/chrome/m80/src/utils/SkUTF.cpp#L70-L94

And I think I agree with that logic of keeping `SKFont` low-level and `SKPaint` as the "more general" usage.

Issue #686, opened 2018-11-15
Switch the compiler to Clang


mattleibow:
### Description

Google has indicated that they only optimize for the clang compiler:
 - Windows: https://groups.google.com/forum/#!topic/skia-discuss/ZZ7Ol3alB-I
 - Linux: https://groups.google.com/forum/#!topic/skia-discuss/P4tg5eBVIiU

dtoppani-twist:
I see work to use clang on linux to get more performance - is it possible to do so on Windows or there are specific blockers?

mattleibow:
It has been a long time, but we are moving Win32 to Clang. Noticing vast improvements already, just check out #758. Need a little more time for UWP, but that can come later as there are probably fewer users of that platform.

As soon as #1007 is merged, I'll get a preview release out for the world.

daltonks:
Hey Matt,
I am a user of UWP and would really appreciate Clang there too. I might be one of the few though 😄

mfkl:
Hi Matthew,
[This](https://code.videolan.org/videolan/docker-images/-/tree/master/vlc-debian-llvm-uwp) landed recently, thought you might find it useful (it uses llvm-mingw).

mattleibow:
Thanks! I must really get to this sometime.

daltonks:
I'm working on a feature for my UWP app that could really use this (rotated images). It's unfortunate how slow the rendering is at the moment.

rubiomiguel06:
I am also one of the few... But I am really having a hard time with the rendering performance of rotated images with Skia. Is this planned to be solved any time soon?
Thanks @mattleibow for all your research and work on this topic

daltonks:
I'm also still struggling with this on UWP. It's unfortunate that my desktop is slower than my Android phone, and users notice it too.
I'll eventually move to MAUI (hoping that it's better), but that's a long way off.

Issue #758, opened 2019-01-14
SKCanvas rendering rotated image very slow compared to System.Drawing (4x slower)


MathieuJack:
Rendering a rotated bitmap with SKCanvas is slow, compared with System.Drawing (4x slower)
See below code, it produces following output:
rotation: 20°
Skia: 6860ms
System.Drawing: 1643ms

           float rotation = 20;
            Console.WriteLine("rotation: "+rotation+"°");

            // Skia part...
            MemoryStream imageStream = new MemoryStream(new WebClient().DownloadData(@"https://via.placeholder.com/450"));
            SKBitmap skBitmapToBlit = SKBitmap.Decode(imageStream);
            SKBitmap renderSkBitmap = new SKBitmap(512, 512, isOpaque: false);
            SKCanvas canvas = new SKCanvas(renderSkBitmap);
            SKPaint paint = new SKPaint()
            {
                FilterQuality = SKFilterQuality.High
            };
            canvas.RotateDegrees(rotation);
            var sw = Stopwatch.StartNew();
            for (int i=0; i<100; i++)
            {             
                canvas.DrawBitmap(skBitmapToBlit, new SKPoint(10, 10), paint);
            };
            Console.WriteLine("Skia: "+sw.ElapsedMilliseconds + "ms");

            // System.Drawing part...
            imageStream = new MemoryStream(new WebClient().DownloadData(@"https://via.placeholder.com/450"));
            Bitmap gdiBitmapToBlit = new Bitmap(imageStream);
            Bitmap renderGdiBitmap = new Bitmap(512, 512, System.Drawing.Imaging.PixelFormat.Format32bppArgb);
            Graphics graphics = Graphics.FromImage(renderGdiBitmap);
            graphics.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.Bicubic;
            graphics.SmoothingMode = System.Drawing.Drawing2D.SmoothingMode.HighSpeed;
            var matrix = new Matrix();
            matrix.Rotate(rotation);
            graphics.Transform=matrix;
            sw = Stopwatch.StartNew();
            for (int i = 0; i < 100; i++)
            {
                graphics.DrawImage(gdiBitmapToBlit, new PointF(10, 10));
            };
            Console.WriteLine("System.Drawing: " + sw.ElapsedMilliseconds + "ms");

Note that, without rotating, Skia is much faster than System.Drawing (36x faster).

Symptom : Skia 4x slower than System.Drawing, with provided code.
[...]

PR #1006, opened 2019-11-11
Use Clang for Windows (m79)

mattleibow:
**Description of Change**

Describe your changes here. 

**Bugs Fixed**

- Related to #686
- Related to #1007

**API Changes**

Added: 
 
- `void Class.Method();`

Changed:

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


mattleibow:
This is very weird. I got a 32-bit build locally, now I can't. Seems to have issues with 64-bit and the linker. CI has all sorts of weird issues, and now I have those same issues. How did I get a working build locally? I know it was a real build because I saw a massive change in the performance of drawing. NICE!

Gillibald:
Maybe we should check out the original build file https://github.com/google/skia/blob/master/gn/BUILD.gn

mattleibow:
I don't think I changed anything around clang - I have been trying to make sure I make minimal changes. I updated to Clang 9 and it seemed to build locally. Just updating my agents to that and we shall see. I did a test a local build and ran the benchmarks for #758 - I saw MASSIVE improvements. Instead of taking 50000us, it took 900us - and the unrotated canvas was 200us. I think we got ourselves a winner. As soon as CI is green, then I'll post the binaries here for testing - it is the new version though. And, I'll see what it takes to get the currently released version using clang so we can get these same improvements there.

mattleibow:
Trying to get m68 (the current version) using clang in #1007

mattleibow:
Clang works, so merging all the PRs into branches. 🎉

Issue #252, opened 2017-02-17
Vulkan GPU backend


jcoder58:
Does SkiaSharp support the [Vulkan GPU Backend](https://skia.org/user/special/vulkan)?  If not, any plans to add it?  I'm working on a Vulkan app and am looking for a 2D solution for the UI.

**Current PRs**
 - native: https://github.com/mono/skia/pull/69
 - managed: #1010

> VS bug [#727962](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/727962)

mattleibow:
Most certainly we are going to be adding this. At this time no-one has asked, so we never got around to it, but it should be relatively simple to add. It just needs the few setup types, and then that gets handed over the the `GRContext` to continue as normal.

Feel free to have a go and send a PR :)

jcoder58:
>Feel free to have a go and send a PR :)

I'll take a look at it.  Is there anyway to build just the Desktop?

mattleibow:
I usually just build `externals-windows` on Windows and `externals-osx` on macOS. This is the native bits, and then you can use the unit test projects to test. If you want to run a UI with it, you can use the samples. Just uninstall the SkiaSharp NuGet and add the project references.

The samples use NuGets because they are aimed at people just downloading and running the samples. For dev, I use the `WindowsSample` or the `MacSample`. You should just need to add the `/binding/Binding/Binding.shproj` and the `/binding/SkiaSharp.[Desktop/OSX]/*.csproj`

Looking forward to seeing what you are doing.

jcoder58:
Tried running WindowExample and am getting:
    
    Unable to load DLL 'libSkiaSharp.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)

libSkiaSharp.dll is installed in the x86/x64 folders.

mattleibow:
@jcoder58 libSkiaSharp mush be in the same directory as the SkiaSharp.dll. Are you building as x86/x64?

mattleibow:
@jcoder58, If you are still working, we can chat on https://gitter.im/xamarin/XamarinComponents and get you started.

jcoder58:
Well, that didn't go far.  Skia doesn't have the Vulkan backend in the C API :(.

mattleibow:
That is what I am doing with SkiaSharp... the C API. And then the C# API on that.

[...]

PR #1010, opened 2019-11-14
Vulkan Support for Win32 (and APIs for everyone as well)

### LINKED ISSUE #1222: [QUESTION] (dev/update-skia) Wrong naming in new SkColorType members?
The `dev/update-skia` branch has new `SkColorType` members:

```cs
	public enum SKColorType
	{
		Unknown = 0,
		Alpha8 = 1,
		Rgb565 = 2,
		Argb4444 = 3,
		Rgba8888 = 4,
		Rgb888x = 5,
		Bgra8888 = 6,
		Rgba1010102 = 7,
		Rgb101010x = 8,
		Gray8 = 9,
		RgbaF16 = 10,
                // vvv NEW vvv
		RgbaF16Normalized = 11,
		RgbaF32 = 12,
		R8g8Unnormalized = 13,
		A16Float = 14,
		R16g16Float = 15,
		A16Unnormalized = 16,
		R16g16Unnormalized = 17,
		R16g16b16a16Unnormalized = 18,
	}
```

These correspond to Skia's 

```cpp
enum SkColorType {
    kUnknown_SkColorType,      //!< uninitialized
    kAlpha_8_SkColorType,      //!< pixel with alpha in 8-bit byte
    kRGB_565_SkColorType,      //!< pixel with 5 bits red, 6 bits green, 5 bits blue, in 16-bit word
    kARGB_4444_SkColorType,    //!< pixel with 4 bits for alpha, red, green, blue; in 16-bit word
    kRGBA_8888_SkColorType,    //!< pixel with 8 bits for red, green, blue, alpha; in 32-bit word
    kRGB_888x_SkColorType,     //!< pixel with 8 bits each for red, green, blue; in 32-bit word
    kBGRA_8888_SkColorType,    //!< pixel with 8 bits for blue, green, red, alpha; in 32-bit word
    kRGBA_1010102_SkColorType, //!< 10 bits for red, green, blue; 2 bits for alpha; in 32-bit word
    kBGRA_1010102_SkColorType, //!< 10 bits for blue, green, red; 2 bits for alpha; in 32-bit word
    kRGB_101010x_SkColorType,  //!< pixel with 10 bits each for red, green, blue; in 32-bit word
    kBGR_101010x_SkColorType,  //!< pixel with 10 bits each for blue, green, red; in 32-bit word
    kGray_8_SkColorType,       //!< pixel with grayscale level in 8-bit byte
    kRGBA_F16Norm_SkColorType, //!< pixel with half floats in [0,1] for red, green, blue, alpha; in 64-bit word
    kRGBA_F16_SkColorType,     //!< pixel with half floats for red, green, blue, alpha; in 64-bit word
    kRGBA_F32_SkColorType,     //!< pixel using C float for red, green, blue, alpha; in 128-bit word

    // The following 6 colortypes are just for reading from - not for rendering to
    kR8G8_unorm_SkColorType,   //<! pixel with a uint8_t for red and green

    kA16_float_SkColorType,    //<! pixel with a half float for alpha
    kR16G16_float_SkColorType, //<! pixel with a half float for red and green

    kA16_unorm_SkColorType,    //<! pixel with a little endian uint16_t for alpha
    kR16G16_unorm_SkColorType, //<! pixel with a little endian uint16_t for red and green
    kR16G16B16A16_unorm_SkColorType,//<! pixel with a little endian uint16_t for red, green, blue, and alpha
```

However, the `unorm` in the C++ Skia API means `unsigned normalized`, not `un-normalized`

Maybe `Unnormalized` means unsigned normalized, but in the industry, `UNORM` and `SNORM` are the naming conventions.

Also - at least in OpenGL - `UNORM` is the standard, so not explicitly mentioned. `SNORM` is non-standard, and explicitly mentioned. 

The existing color types in  SkiaSharp v1.68 are actually all `UNORM`, but this is not mentioned in the name. E.g.  `Bgra8888` is actually an `UNORM` 8-bit unsigned integer. 

One exception though, the `SkColorType.RgbaF16` in v1.68 is actually **not** normalized at all... The new `RgbaF16Normalized` is, but `normalized` here means something different, since the floating point numbers are not [normalized](https://en.wikipedia.org/wiki/Normalization_(image_processing)), but rather just **clamped** between `[0..1]`...  So maybe `SkColorType.RgbaF16Clamped` would be a better way to describe this? I'm not 100% sure this is correct, maybe `kRGBA_F16Norm_SkColorType` does something different...

Since the new members are not yet in `master`, maybe the C# API could use a more consistent naming? I would just drop the `unnormalized` in the name, but I'm not sure what to do with the floating point related members...

Tricky.

mattleibow:
**Description of Change**

This is the workspace for Vulkan dev. 

This PR has the support for Win32 builds. Android and Linux are in this PR: #1287

The native PR is https://github.com/mono/skia/pull/69

New tests were added in a separate project so that they can be run separately (or ignored on platforms that don't support Vulkan) as well as to work around the fact that the library we use to set up the Vulkan surfaces [SharpVk](https://github.com/FacticiusVir/SharpVk) not strong named.

**Building Vulkan**

In order to build SkiaSharp with Vulkan support, either set the environment variable `SUPPORT_VULKAN` or the command line arguments `--supportVulkan` to `true` or `1`.

> The Vulkan headers ship with the skia project source, so no need to install and SDKs.

**Bugs Fixed**

- Related to #252
- Related to #897 (thanks to @tuccio) 
- Fixes #1222

**API Changes**

 - Adding the Vulkan support APIs:
    - new class: `GRSharpVkBackendContext`
    - new class: `GRVkBackendContext`
    - new class: `GRVkExtensions`
    - new method: `GRContext.CreateVulkan(GRVkBackendContext)`
    - new ctor: `GRBackendTexture(int, int, GRVkImageInfo)`
    - new ctor: `GRBackendRenderTarget(int, int, int, GRVkImageInfo)`
    - new struct: `GRVkAlloc`
    - new struct: `GRVkImageInfo`
    - new struct: `GrVkYcbcrConversionInfo`
 - Renames the `SKColorType` enum values (#1222)  
    - `Unnormalized` suffix was incorrect (should be "unsigned normalized") and is the default, so remove it
    - `RgbaF16Normalized` was not a correct "normalization" and is actually clamped
    - Rename the new values to follow the pattern:  
      integral: `<components><sizes>` (RGB 565 = Rgb565)
      floating: `<component>F<component size>` (RGBA half-float = RgbaF16)
- Obsoletes and hide the `GRPixelConfig` as that is now private API
    - Remove the mapping code that used to pass through this type

**Behavioral Changes**

New members to use the Vulkan APIs. No changes to functionality.

**PR Checklist**

- [x] ~~Updated documentation~~ _(in a separate PR)_


binding/Binding/GRVkBackendContext.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+	public class GRVkBackendContext : SKObject, ISKReferenceCounted
+	{
+		[Preserve]
+		internal GRVkBackendContext (IntPtr h, bool owns)
+			: base (h, owns)
+		{
+		}
+
+		public static GRVkBackendContext Assemble (
```
mattleibow: See some comments in https://github.com/mono/skia/pull/69, but I would rather have the overload that takes delegates - but if this is useful, then we can keep this one.

binding/Binding/GRVkInterface.cs
```diff
@@ -0,0 +1,34 @@
...
+	{
+		[Preserve]
+		internal GRVkInterface (IntPtr h, bool owns)
+			: base (h, owns)
+		{
+		}
+
+		public static GRVkInterface Create (
+			IntPtr vkGetInstanceProcAddr,
+			IntPtr vkGetDeviceProcAddr,
```
mattleibow: Instead of having pointers, I think we need actual delegates as well - or maybe replace this version.

tuccio: The way you retrieve those function pointers doesn't depend just on the platform, but also from the loader you use. Not sure what you have in mind to simplify, but I don't see a way that is flexible enough to cover all the needs that doesn't involve passing function pointers provided by the loader.

binding/Binding/GRVkBackendContext.cs
tuccio: I can say it's quite useful in a case like ours, where we interop between C++ and C# and share the vulkan instance between the two, while also creating a separate vulkan queue for skia, to allow the UI to render in a separate thread.

tuccio:
As mentioned in the issue, in my company we have been running with those changes for a few months without problems. I built and tested this branch as well today and everything seems in order.

mattleibow:
Hmm, looking at the state today, I see the whole API has changed: https://github.com/google/skia/tree/chrome/m79/include/gpu/vk

I think this PR will only get merged after I get the m79 branch merged. I don't want to implement a feature in a minor version, and then break it in the next stable.

MarchingCube:
@mattleibow Can you merge latest master again into this branch? We would like to update again (to get recent memory fix) sometime next week.

And I agree that we should wait for m79 to proceed with this PR. Do you have an estimate when this will happen (since we would help with exposing new API) ?

mattleibow:
@MarchingCube updated - took a bit of a while as I was balling with my machine and Docker containers :( All fixed now so we have nice Docker samples :)

I can't give any dates with m79 because I got pulled off for a few days. I really hope to get it done soon - trying for January. The main issue is there is a fair bit of breaking changes, so we are going to have to do a v2 release and pretty much nothing will work. SKPaint is now two separate objects and that is going to be fun for everyone. It makes technical sense, but so do vaccines - but they also hurt :P

MarchingCube:
@mattleibow Since there was progress on getting m80 up and running - what would be the action plan for this PR? I guess it should `target` develop now and it would be really good to get this in sometime so we won't have to keep a fork running for another year :)

mattleibow:
@MarchingCube @tuccio I updated the code! It doesn't actually compile anymore since they changed a whole bunch. But, you should now be able to work from this and make the same changes.

mattleibow:
This has been pushed to the preview feed for initial testing:
 - feed: https://aka.ms/skiasharp-eap/index.json
 - version: **2.80.0-pr.1010.14**
 - packages: **SkiaSharp** and **SkiaSharp.Vulkan.SharpVk**

mattleibow:
One thing that I am not sure on yet is the namespace in the "SkiaSharp.Vulkan.SharpVk" package... I am using `SkiaSharp.Vulkan.SharpVk`. This is going to make the types harder to find, and the types are very few (only 2) so it is not going to clutter the namespace.

Should I just keep it in the root `SkiaSharp`?

Issue #1032, opened 2019-11-25
[BUG] Possible Leak with static Dictionary SKObject.instances


Knuddlbaer:
SKObject holds Handled to WeakReferences in an static Dictionary bit never removes entries.

In RegisterHandle a new Entry is created with the handle ob object to an WeakReference. In DeregisterHandle, the Entry will only updates with WeakReference(null) but never deleted. This leads to tons of entries with tons of  WeakReferences.

**Expected Behavior**

On Dispose the handle need to be removed from SKObject.instances

**Actual Behavior**

- Version with issue:  1.68.1
- Last known good version:  N/A
- Platform Target Frameworks: Xamarin.Android
  - Android: 29
- Target Devices:   Pixel 2 XL

**Screenshot**

![skia](https://user-images.githubusercontent.com/58169312/69532581-6ac3dd00-0f76-11ea-922f-a1d1ae68aae9.png)

MarchingCube:
I've noticed this in Avalonia (https://github.com/AvaloniaUI/Avalonia) as well, when we redraw we end up with a lot of `ConcurrentDictionary` allocations coming from SkiaSharp.

Gillibald:
I think the AddOrUpdate call is wrong in the method that is responsible for the unregister. Should be easy to fix but haven't found time yet.

Knuddlbaer:
AddOrUpdate do not remove the entry in the dictionary. If the SKObject is detroyed, the entry in dictionary is simple replaced by new weakreference. Instead the handle have to removed.

FoggyFinder:
@Knuddlbaer Why issue was closed?

mattleibow:
Reopening so we can get this fixed.

mattleibow:
Closed in #1039

mattleibow:
I just pushed the NuGets to the preview feed:
https://nugetized.blob.core.windows.net/skiasharp-eap/index.json

1.68.2-preview.5

PR #1039, opened 2019-12-01
TryRemove SKObject.instances WeakReferences

daltonks:
**Description of Change**

`SKObject.DeregisterHandle(...)` now uses `TryRemove` in order to remove handles from the `ConcurrentDictionary`. I kept the logic the same otherwise.

Should there be tests for this besides those that already exist? I might be able to add something if so. Don't know the codebase too well.

**Bugs Fixed**

- #1032 "Possible Leak with static Dictionary SKObject.instances"

**API Changes**

None

**Behavioral Changes**

None

**PR Checklist**

- [X] Rebased on top of master at time of PR
- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKObject.cs
+++ b/binding/Binding/SKObject.cs
@@ -143,22 +143,20 @@ namespace SkiaSharp
 			if (handle == IntPtr.Zero)
 				return;
 
-			var nonExistent = new WeakReference (null);
-			var weak = instances.AddOrUpdate (handle, nonExistent, (ptr, old) => {
-				if (old.Target is SKObject obj && !obj.IsDisposed)
-					return old;
-				return new WeakReference (null);
-			});
+			var existed = instances.TryRemove (handle, out var weak);
+			if (existed && weak.Target is SKObject obj && !obj.IsDisposed) {
+				// Existing object is not disposed, so re-register it
+				RegisterHandle (handle, obj);
+			}
 
 #if THROW_OBJECT_EXCEPTIONS
 			InvalidOperationException ex = null;
-			if (weak == nonExistent) {
-				// the dummy instance was added instead of being removed
+			if (!existed) {
+				// there was no object for the handle
 				ex = new InvalidOperationException (
 					$"A managed object did not exist for the specified native object. " +
 					$"H: {handle.ToString ("x")} Type: {instance.GetType ()}");
-			}
-			if (weak.Target is SKObject o && o != instance && !instance.IsDisposed) {
+			} else if (weak.Target is SKObject o && o != instance && !instance.IsDisposed) {
 				// there was a new living object there, but we are still alive
 				ex = new InvalidOperationException (
 					$"Trying to remove a different object with the same native handle. " +


binding/Binding/SKObject.cs
```diff
@@ -143,22 +143,20 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
-			var nonExistent = new WeakReference (null);
-			var weak = instances.AddOrUpdate (handle, nonExistent, (ptr, old) => {
-				if (old.Target is SKObject obj && !obj.IsDisposed)
-					return old;
-				return new WeakReference (null);
-			});
+			var existed = instances.TryRemove (handle, out var weak);
+			if (existed && weak.Target is SKObject obj && !obj.IsDisposed) {
+				// Existing object is not disposed, so re-register it
+				RegisterHandle (handle, obj);
```
mattleibow: We need to be careful here with concurrency. The reason I used `AddOrUpdate` was because the key was "locked" during this block so I could do everything.

With two separate statements, the object could be valid, then while starting to execute the `RegisterHandle` the object could be disposed. We would then be registering a `null` handle. (but, looking at the code for `RegisterHandle`, this may not actually be a problem because we have some checks and registering a `null` handle may not be such an issue)

I had a big, fat `lock` around the register/deregister logic before and that was nice to prevent any concurrency, but it was overkill.

(I wrote a test that maybe needs a bit more concurrency testing: https://github.com/mono/SkiaSharp/blob/v1.68.1/tests/Tests/SKObjectTest.cs#L267-L307)

Just a thought. I think this is good to merge, but just wanted to see if you had any further comments.

mattleibow:
The failing build is just because the internal iOS build does not have access to the secure signing keys.

Gillibald:
Register handle can deal with the case that an object is being disposed during the derigest. In the register method you just do the same check for IsDisposed == true and do nothing in that case. AddOrUpdate is the wrong approach because there are cases that an object is no longer valid during register and unregister. If you wanted to keep AddOrUpdate you have to scan the dictionary for dead references regularly.

binding/Binding/SKObject.cs
daltonks: Makes sense. Because of these concerns, I used `RegisterHandle` instead of more logic in this method. However, I am unsure if it is sufficient in every concurrent scenario.

daltonks: Gillibald mentioned "In the register method you just do the same check for IsDisposed == true and do nothing in that case."

Should something like that be part of this PR? If so, would that just be part of the guard clause, or would `RegisterHandle(...).Update(...)` need to be modified as well?

mattleibow: I think this PR is good.

daltonks: Cool deal, thanks 😎

mattleibow:
I just pushed the NuGets to the preview feed:
https://nugetized.blob.core.windows.net/skiasharp-eap/index.json

1.68.2-preview.5

PR #1067, opened 2019-12-19
Multitarget Views.Desktop.Common

bender2k14:
**Description of Change**

Recreating PR #1066, which was closed when I deleted by branch.

Changes `SkiaSharp.Views.Desktop.Common` to multitarget
- .NET Standard 2.0,
- .NET Framework 4.5, and
- .NET Core 3.0.

Also removes `SkiaSharp.Views.Desktop.Common.NetStandard` _and all references to it_ (which wasn't the case in PR #1066).

I don't know if any part of the automated builds/devops needs to change as well.  Someone should check that.

**Bugs Fixed**

Related to issue #1029.  This PR gets us closer to dual targeting .NET Framework and .NET Core for `SkiaSharp.Views.WPF`.

**API Changes**

None

**Behavioral Changes**

None (intended)

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
 - NA
- [x] Changes adhere to coding standard
- [ ] Updated documentation

PR #1068, opened 2019-12-19
Multitarget Views.Desktop.Common 3

bender2k14:
This PR is similar to #1067, which is for a branch with the two commits that fails the automated build.  To gain more information, I am creating this PR to see if just the first of those commits passes the automated build.  This PR is not meant to be completed.  I think it would be better if we could figure out how to get the other branch to pass the automated build and then complete that PR instead.


bender2k14:
Abandoning per https://github.com/mono/SkiaSharp/pull/1067#issuecomment-567554218

Issue #1029, opened 2019-11-22
[FEATURE] Make the Views.WindowsForms and Views.WPF actually target .NET Core


mattleibow:
**Is your feature request related to a problem? Please describe.**
Right now, installing the SkiaSharp.Views.WindowsForms or SkiaSharp.Views.WPF packages will give a warning saying that it is not targeting the right thing and all that.

- [x] SkiaSharp.Views.Desktop.Common
- [x] SkiaSharp.Views.WPF
- [x] SkiaSharp.Views.WindowsForms

**Describe the solution you'd like**
We would like a nice package that targets both net45 and netcoreapp

**Describe alternatives you've considered**
We can use the NoWarn feature, but that is extra work that is unnecessary.

bender2k14:
I might be able to help get dual targeting working for `View.WPF`.

----------------------

Currently, `View.WPF` references on NuGet packages `OpenTK` and `OpenTK.GLControl`, but the project still compiles if they are removed.

Are these 
----------------------

bender2k14:
I might be able to help get dual targeting working for `View.WPF`.

----------------------

Currently, `View.WPF` references NuGet packages `OpenTK` and `OpenTK.GLControl`, but the project still compiles if they are removed.

Are these references needed?  Are they only runtime dependencies on them?

----------------------

(After removing references to `OpenTK` and `OpenTK.GLControl`...)

I can get `View.WPF` to build without any warnings when targeting either `net45` or `netcoreapp3.0` but not when targeting both.  I could share more details about this, but there is another possibility.

Instead of targeting `net45` and `netcoreapp3.0`, what about targeting `net472` and `netcoreapp3.0`?

This makes things easier, because `Views.Desktop.Common.NetStandard` targets `netstandard2.0`, so both `net472` and `netcoreapp3.0` can target that ([source](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)).

----------------------

Even after removing the references to `OpenTK` and `OpenTK.GLControl` and dual targeting `net472` and `netcoreapp3.0`, I still had a problem (copying a `pdb` file).  I can share more details about this as well if this is the direction that should be taken.

bender2k14:
> I can get View.WPF to build without any warnings when targeting either net45 or netcoreapp3.0 but not when targeting both. I could share more details about this...

Quoting from https://github.com/mono/SkiaSharp/issues/1011#issue-523581067
[...]

PR #1069, opened 2019-12-19
Multitarget Views.WPF

bender2k14:
**Description of Change**

`SkiaSharp.View.WPF` now targets `.NET Core 3.0` in addition to `.NET Framework 4.5.1`.

**Bugs Fixed**

Related to issue #1029, which is about getting both `SkiaSharp.View.WPF` as well as `SkiaSharp.View.WindowsForms` to target `.Net Core`.

**API Changes**

None

**PR Checklist**

 - NA
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
@@ -1,12 +1,13 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
   <PropertyGroup>
-    <TargetFramework>net45</TargetFramework>
+    <TargetFrameworks>net451;netcoreapp3.0</TargetFrameworks>
+    <UseWpf>true</UseWpf>
     <RootNamespace>SkiaSharp.Views.WPF</RootNamespace>
     <AssemblyName>SkiaSharp.Views.WPF</AssemblyName>
     <PackagingGroup>SkiaSharp.Views.WPF</PackagingGroup>
     <DefineConstants>$(DefineConstants);__DESKTOP__;__WPF__</DefineConstants>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
     <Reference Include="PresentationCore" />
     <Reference Include="PresentationFramework" />
     <Reference Include="WindowsBase" />
--- a/source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
@@ -1,13 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
   <PropertyGroup>
-    <TargetFrameworks>net451;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>net45;netcoreapp3.0</TargetFrameworks>
     <UseWpf>true</UseWpf>
     <RootNamespace>SkiaSharp.Views.WPF</RootNamespace>
     <AssemblyName>SkiaSharp.Views.WPF</AssemblyName>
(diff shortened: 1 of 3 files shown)


source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
```diff
@@ -1,12 +1,13 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
   <PropertyGroup>
-    <TargetFramework>net45</TargetFramework>
+    <TargetFrameworks>net451;netcoreapp3.0</TargetFrameworks>
```
mattleibow: Can we add .netstandard2.0 here? Not sure a bout the rules of a netstandard referencing a netcoreapp. Does the whole pipeline of WPF require netcoreapp at the TFM?

mattleibow:
We can also update the nuspecs with the new netcoreapp3.0:
https://github.com/mono/SkiaSharp/blob/master/nuget/SkiaSharp.Views.Desktop.Common.nuspec#L39

And:
https://github.com/mono/SkiaSharp/blob/master/nuget/SkiaSharp.Views.WPF.nuspec#L38

> Just note, the nuspecs are pseudo-nuspec and are manipulated in the build. That is why they are a bit different.

source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
bender2k14: I don't think so.  Even though both .NET Framework (in this case, version 4.5.1) and .NET Core 3.0+ support WPF, I don't think any version .NET Standard does.

bender2k14:
What should the platform be for
https://github.com/mono/SkiaSharp/blob/master/nuget/SkiaSharp.Views.WPF.nuspec#L38
?

mattleibow:
The platform that builds it. It looks to build on all the platforms, so leave it out.

bender2k14:
> We can also update the nuspecs with the new netcoreapp3.0:
https://github.com/mono/SkiaSharp/blob/master/nuget/SkiaSharp.Views.Desktop.Common.nuspec#L39
>
> And:
> https://github.com/mono/SkiaSharp/blob/master/nuget/SkiaSharp.Views.WPF.nuspec#L38

Done :)

source/SkiaSharp.Views/SkiaSharp.Views.WPF/SkiaSharp.Views.WPF.csproj
mattleibow: I see you changed this to 451, is there a reason?

Either way, the build is failin on the packaging because the nuspec still expects net45:
> Could not find a part of the path 'D:\a\1\s\output\SkiaSharp.Views.WPF\nuget\lib\net45'.

bender2k14: Oh, it appears not.  Sorry about that.

I would have bet money that I saw some other `net45` to `net451` change between the `master` and `dev/vs-2019` branches...and then I would have lost that money!

I just pushed a commit that reverts this back to `net45`.

mattleibow:
Thanks for the PR! Merged!

bender2k14:
My pleasure :)

Is there a NuGet package (possibly a pre-release) with these changes?

I would test it by adding it to a project we are developing at work.

mattleibow:
We can! Just noticed that this branch failed to build because DevOps got disconnected from all its agents... Rebuilding and then I'll let you know when you can get your hands on the build.

I did start a build on our public CI:
https://dev.azure.com/xamarin/public/_build/results?buildId=11744&view=results

That should work for testing, but I am also going to push out the PR build to the preview feed:

> https://aka.ms/skiasharp-eap/index.json

The reason for this is that CI builds both the preview and the stable packages to make sure everything is working. Then when I release I first push the preview packages for a few weeks, and then go stable - no need to rebuild. But, sometimes you may be tempted to select the stable packages from the CI server - and then it gets cached... I can control what gets to the preview feed.

mattleibow:
Hehe, seems my certificates to build the samples expired today... I'm going to have to upload them to CI, but I need to get a sleep action going here. When I wake up in a few hours, I'll rebuild with the v1.68.2 and push that after merging the VS2019 build to master.

PR #1078, opened 2019-12-21
Remove OpenTK references from View.WPF

bender2k14:
**Description of Change**

In addition to having no compile-time dependency (and pointed out in https://github.com/mono/SkiaSharp/issues/1029#issuecomment-568156345), the nuspec file for View.WPF does not reference any OpenTK packages, so Views.WPF shouldn't need to reference these packages either.

**Bugs Fixed**

Related to issue #1029.  This change removes two warnings when targeting .NET Core in View.WPF.

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

 - NA
- [ ] Updated documentation


mattleibow:
I'm a cowboy 🤠 just merging PRs before the build even runs!
Thanks!

PR #953, opened 2019-09-06
Disallow intercept of touch event

FoggyFinder:
**Description of Change**

Added call to `RequestDisallowInterceptTouchEvent`.

Tests aren't provided - not sure is it required.

**Bugs Fixed**

Fixed: #937 

shfb

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Updated documentation

Issue #937, opened 2019-09-03
[BUG] [Android] Dissalow further propagation for touch events


FoggyFinder:
**Description**

When `SKCanvasView` is placed inside, for example, `ScrollViewer` then touch events are intercepted even though `e.Handled = args.Handled`.

Here is a link to the SO topic with proper description of the issue:

[ACTION_CANCEL while touching](https://stackoverflow.com/questions/6018309/action-cancel-while-touching)

And with advise how to handle it.

**Code**

With slight modification to demo app:

https://github.com/mattleibow/SkiaSharpDemo/tree/master/SkiaSharpDemo/SkiaSharpDemo

Add `ListView` to the bottom and `Label` to the top of the `StackLayout`:

```xml
    <ScrollView>
        <StackLayout>
            <Label Text="Some text" HorizontalOptions="Center" />
            <views:SKCanvasView HeightRequest="300"
                                WidthRequest="400"
                                HorizontalOptions="Center" 
                                PaintSurface="OnPainting" EnableTouchEvents="True" Touch="OnTouch" />
            <ListView  />
        </StackLayout>
    </ScrollView>
```
I also changed background color but it's doesn't related.

**Expected Behavior**

Content doesn't scroll when user drags inside of `SKCanvasView`.

**Actual Behavior**

It does scroll:

![skDroid](https://user-images.githubusercontent.com/16510738/64198264-9f9d2800-ce90-11e9-8360-fb02abc1fabe.gif)

**Basic Information**

- Version with issue:  latest
- Last known good version:  none
- IDE:  Visual Studio
- Platform Target Frameworks: - Android: 8.1

**Reproduction Link**

Since it's very easy to modify demo I haven't uploaded MCVE anywhere. But I could if it's required.

charlesroddie:
@axeluser can you look at a fix for this?
We tried disabling the scrollview on initial touch (after your PR in Xamarin.Forms) but that doesn't prevent subsequent touch events from being eaten by the scrollview.

mattleibow:
So, I had a look and tested a few things.

What I found was that UWP works in that as soon as you start drawing, it captures the pointer and won't scroll. This PR will fix it for Android.

I checked iOS and there is no easy way to prevent the bubbling of events, so it always scrolls. I fill that it is correct to stop scrolling, so we should probably fix iOS as well.

Not sure what Tizen, macOS, WPF will do.

[...]

PR #1096, opened 2020-01-10
Disallow intercept of touch event by adding option in SKTouchEventArgs

mackel90:
**Description of Change**

Updated "SKTouchEventArgs" so it contains boolean "RequestDisallowIntercept" which will call the equal function in Androids SKTouchHandler to enable more control for the user like the solution of @FoggyFinder in #953 

Tests made but hang after "SkiaSharp.Tests.SKCodecTest.ReadOnlyStream [FINISHED] Time: 0,0163149s"

**Bugs Fixed**

- Related to issue #937 

**API Changes**

Added: 
 
- bool SkiaSharp.Views.Forms.SKTouchEventArgs.RequestDisallowIntercept { get; set; }

**PR Checklist**

- [ ] Updated documentation


xpios01:
I don't see this in the current code on github. When is this going to make it in or was there a different fix added somewhere? I need to be able to prevent the scrollview from taking over touch events when I'm dragging an object on a surface within the scrollview.

thanks.

PR #1101, opened 2020-01-14
Updating the SKFont members

mattleibow:
So far, the `GetGlyphs` and `CountGlyphs` is what I am focusing on right now to decide what the best API is.

Previously it was all arrays and allocations. I am now trying to have the option to use arrays if need be, but more provide the option to have zero-allocation code.


binding/Binding/SKFont.cs
```diff
@@ -0,0 +1,718 @@
...
+		public ushort GetGlyph (int codepoint) =>
+			SkiaApi.sk_font_unichar_to_glyph (Handle, codepoint);
+
+		// GetGlyphs
+
+		public ushort[] GetGlyphs (ReadOnlySpan<int> codepoints)
+		{
+			var glyphs = new ushort[codepoints.Length];
+			GetGlyphs (codepoints, glyphs);
+			return glyphs;
```
mattleibow: This is option A for the most basic. Ask for a value, and we will make sure we give you the return you are expecting. This will result in an allocation, but does not require you to worry about releasing memory.

binding/Binding/SKFont.cs
```diff
@@ -0,0 +1,718 @@
...
+
+		public ushort[] GetGlyphs (ReadOnlySpan<char> text)
+		{
+			var n = CountGlyphs (text);
+			if (n <= 0)
+				return new ushort[0];
+
+			var glyphs = new ushort[n];
+			GetGlyphs (text, glyphs);
+			return glyphs;
```
mattleibow: This is an alternate option A that has 2 steps to allocation - first determine the size, then allocate. Similar to the previous, the allocation happens inside the member and the GC cleans up.

binding/Binding/SKFont.cs
```diff
@@ -0,0 +1,718 @@
...
+			GetGlyphs (codepoints, glyphs);
+			return glyphs;
+		}
+
+		public void GetGlyphs (ReadOnlySpan<int> codepoints, Span<ushort> glyphs)
+		{
+			fixed (int* up = codepoints)
+			fixed (ushort* gp = glyphs) {
+				SkiaApi.sk_font_unichars_to_glyphs (Handle, up, codepoints.Length, gp);
+			}
```
mattleibow: This is option B, no allocations internally. All memory must be provided and it is then used as is.

Potential usage with `ArrayPool`:
```csharp
// magic some codepoints up
int[] codepoints = GetCodepoints ();

// get from the pool
var glyphs = ArrayPool<ushort>.Shared.Rent (codepoints.Length);

var font = new SKFont ();
font.GetGlyphs (codepoints, glyphs);

ProcessGlyphs (glyphs);

// return to pool
ArrayPool<ushort>.Shared.Return (glyphs);
```

binding/Binding/SKFont.cs
```diff
@@ -0,0 +1,718 @@
...
+
+		public void GetGlyphs (string text, Span<ushort> glyphs) =>
+			GetGlyphs (text.AsSpan (), glyphs);
+
+		public void GetGlyphs (ReadOnlySpan<char> text, Span<ushort> glyphs)
+		{
+			fixed (void* p = text)
+			fixed (ushort* gp = glyphs) {
+				SkiaApi.sk_font_text_to_glyphs (Handle, p, (IntPtr)text.Length, SKTextEncoding.Utf16, gp, glyphs.Length);
+			}
```
mattleibow: Again, option B without allocations.

Potential usage:

```csharp
// some magic string
string text = "Hello World!";

var font = new SKFont ();

// first determine the size
var n = font.CountGlyphs (text);

// get from the pool
var glyphs = ArrayPool<ushort>.Shared.Rent (n);

// get the glyphs
font.GetGlyphs(text, glyphs);

ProcessGlyphs(glyphs);

// return to pool
ArrayPool<ushort>.Shared.Return(glyphs);
```

Gillibald:
Looks very promising. I think we should use SKFont in combination with SKTextBlobBuilder from the beginning to make sure both work well with each other and are easy to use.

Passing a string to SKFont to get corresponding glyphs always needs to allocate. Being in control of the allocation is always a good option.

Sometimes users just want the data and don't want to deal with creating buffers and sometimes reusing a buffer is the best perfoming option. Not easy to design an API around these requirements. 

Will add more comments in a bit.

Gillibald:
Your overloads let you pass a pre-allocated buffer and also provide a way to let SkiaSharp allocate everything for you. All variants cover most possible scenarios. 

In general, if SkiaSharp allocates it should return an array if the consumer of the API allocates a buffer the parameter should be of type Span<T>. 

So I think we agree here.

Gillibald:
In general, we will not work with these methods directly and instead `SkiaSharp.HarfBuzz` will interact with SKFont to get the required information for shaping etc.

```
var text = "MyText".AsSpan();

var textBlob = skFont.GetTextBlob(text);
paint.DrawText(textBlob);

var shapedTextBlob = skFont.GetShapedTextBlob(text); // This could be some extention methods that does the font fallback for you etc.
paint.DrawText(shapedTextBlob);
```

PR #1122, opened 2020-02-03
Improve Generator & Structs

mattleibow:
**Description of Change**

Generator:
 - make getters readonly
 - implement IEquatable<T>

Managed API:
 - mark obsolete members as invisible to the editor
 - make struct getters readonly
 - implement struct equality/hashcodes
 - SKColor calculations use the SKColorF implementations
 - some whitespace fixes to conform to styles
 - split SKPMColor out into a new file

**Bugs Fixed**

None.

**API Changes**

All public structs (excluding obsolete ones) are now implementing IEquatable<T> and have overrides for the Equals and GetHashCode members.

**Behavioral Changes**

None visible.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/GRContext.cs
+++ b/binding/Binding/GRContext.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 
 namespace SkiaSharp
 {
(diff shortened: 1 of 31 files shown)


binding/Binding/GRContext.cs
```diff
@@ -99,36 +79,24 @@ public void GetResourceCacheUsage (out int maxResources, out long maxResourceByt
...
+		public int GetMaxSurfaceSampleCount (SKColorType colorType) =>
+			SkiaApi.gr_context_get_max_surface_sample_count_for_color_type (Handle, colorType);
 
+		[EditorBrowsable (EditorBrowsableState.Never)]
 		[Obsolete ("Use GetMaxSurfaceSampleCount(SKColorType) instead.")]
-		public int GetRecommendedSampleCount (GRPixelConfig config, float dpi)
-		{
-			return GetMaxSurfaceSampleCount (config.ToColorType ());
-		}
+		public int GetRecommendedSampleCount (GRPixelConfig config, float dpi) => 0;
```
mattleibow: This was incorrectly selecting the max, which often resulted in a too value. Originally it was the "minimum" value that was "guessed" from properties. The max is actually the highest supported based on properties - which is usually higher that what the surface was actually created with. This causes some issues on some platforms.

PR #1138, opened 2020-02-11
Make sure SKPaint.GetGlyphWidths returns widths for all characters

### LINKED ISSUE #1093: [BUG] SKPaint.GetGlyphWidths returns wrong size array
**Description**

According to https://docs.microsoft.com/en-us/dotnet/api/skiasharp.skpaint.getglyphwidths?view=skiasharp-1.68.1 the return value of SKPaint.GetGlyphWidths is equal to the length of the input string

> Returns
Single[]
Returns the text advances **for each glyph**

In the given code sample this is not true, input string length and result length missmatch.
Also this seems to be a real bug, because if each char of the string is performed separately GetGlyphWidths works as expected.

**Code**

```
        [TestMethod]
        public void TestGlyphWidths()
        {
            //All ISO 10646 chars 
            const int maxChar = (char)65534;
            var builder = new StringBuilder(maxChar);
            for (var c = (char)0; c <= maxChar; c++)
            {
                builder.Append(c);
            }

            var s = builder.ToString();
            using (var newPaint = new SKPaint())
            {
                var glyphWidths = newPaint.GetGlyphWidths(s);

                //Assert below already fails: Assert.AreEqual failed. Expected:<65535>. Actual:<65534>. 
                Assert.AreEqual(s.Length, glyphWidths.Length);
            }

            //But the following works:
            foreach (var c in s)
            {
                using (var newPaint = new SKPaint())
                {
                    var glyphWidths = newPaint.GetGlyphWidths(c.ToString());
                    Assert.AreEqual(1, glyphWidths.Length);
                }
            }
        }
```
**Basic Information**

- Version with issue: 1.68.1.1
- Last known good version:  ?
- IDE:  Visual Studio 2017
- Platform Target Frameworks: 
I could reproduce it as a unit test, so it should be plattform indepentent (Originally observed in WPF and UWP).

mattleibow:
**Description of Change**

Make sure SKPaint.GetGlyphWidths returns widths for all characters. And all related functions which may also have similar issues.

**Bugs Fixed**

- Fixes #1093

**API Changes**

None.

**Behavioral Changes**

All members now correctly process all characters.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/tests/Tests/SKPaintTest.cs
+++ b/tests/Tests/SKPaintTest.cs
@@ -416,6 +416,42 @@ namespace SkiaSharp.Tests
 		}
 
 		[SkippableFact]
+		public void AllCharactersGetGlyphWidthsByString()
+		{
+			// all ISO 10646 chars 
+			const int maxChar = (char)65534;
+			var s = string.Empty;
+			for (var c = (char)0; c <= maxChar; c++)
+			{
+				s += c;
+			}
+
+			using var newPaint = new SKPaint();
+			var glyphWidths = newPaint.GetGlyphWidths(s);
+
+			Assert.Equal(s.Length, glyphWidths.Length);
+		}
+
+		[SkippableFact]
+		public void AllCharactersGetGlyphWidthsByChar()
+		{
+			// all ISO 10646 chars 
+			const int maxChar = (char)65534;
+			var s = string.Empty;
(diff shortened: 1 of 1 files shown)


mattleibow:
Not an issue, see https://github.com/mono/SkiaSharp/issues/1093#issuecomment-584693409

PR #1152, opened 2020-02-21
Improve the experience with the SKTextBlobBuilder

### LINKED ISSUE #1151: [BUG] SKTextBlobBuilder.AddHorizontalRun requires SKPaint.TextEncoding to be set to SKTextEncoding.GlyphId
**Description**

A similar issues #1094 was reported and closed, but I would like to discuss this again.

Calling `SKTextBlobBuilder.AddHorizontalRun` throws a `NullReferenceException` unless the `SKPaint.TextEncoding` is set to `SKTextEncoding.GlyphId`.

IMO this doesn't make sense, since the text encoding is not needed, as the glyph ids are already known when calling this method?

This is caused by the [following old Skia m68 code](https://github.com/mono/skia/blob/2411fe768b0e3d16a1e59839f65b4d802b74499a/src/core/SkTextBlob.cpp#L631) (that doesn't use `SKFont` yet) in `SKTextBlob.cpp`:

```cpp
void SkTextBlobBuilder::allocInternal(const SkPaint &font,
                                      SkTextBlob::GlyphPositioning positioning,
                                      int count, int textSize, SkPoint offset,
                                      const SkRect* bounds) {
    if (count <= 0 || textSize < 0 || font.getTextEncoding() != SkPaint::kGlyphID_TextEncoding) {
        fCurrentRunBuffer = { nullptr, nullptr, nullptr, nullptr };
        return;
    }
```

The `getTextEncoding` check was [removed by Google in this commit](https://github.com/google/skia/commit/b3f4aac0ba02d13af8fa85a33e856d09ab7855b0), where `SKFont` is used instead. 

I'm not sure if the check is actually required?

Personally the imperative API of setting `TextEncoding` on the `SKPaint` feel errors prone, in the latest version Skia the text encoding has to be passed as an argument, and was removed from `SKFont`. But I guess SkiaSharp wants to maintain backwards compatibility? Maybe a major version bump could use a newer version of Skia, where `SKFont` is used? It seems you are already planning this in #711 ?

**_Maybe a simple patch for now could be to validate the `SKPaint.TextEncoding` and throw an exception? Not ideal, but at least the user wouldn't have to dig into the source code to find why the call crashed 😉_** 

**Code**

```CS
  var glyphIds = paint.GetGlyphs("X");
  var glyphXs = new float[] {0};
  var builder = new SKTextBlobBuilder();
  builder.AddHorizontalRun(paint, 0, glyphIds, glyphXs);
```

**Expected Behavior**

A horizontal text run is added to the builder

**Actual Behavior**

A `NullReferenceException` is thrown

**Basic Information**

- Version with issue:  `1.68.2-preview.29`
- Last known good version:  none
- IDE: any
- Platform Target Frameworks: all
- Target Devices:   all

**Workaround**

```CS
  var glyphIds = paint.GetGlyphs("X");
  var glyphXs = new float[] {0};
  var builder = new SKTextBlobBuilder();
  paint.TextEncoding = SKTextEncoding.GlyphId;
  builder.AddHorizontalRun(paint, 0, glyphIds, glyphXs);
```

mattleibow:
**Description of Change**

When the paint encoding is anything but GlyphId, the underlying allocation fails. This results in not only strange behaviour, but also invalid data. The run buffer is given a size, but nothing was allocated. This does 2 things:

- if a buffer is null, override the size
- if the encoding is not glyph, then change it temporarily

**Bugs Fixed**

- Fixes #1151

**API Changes**

None.

**Behavioral Changes**

Instead of throwing an unexpected exception when trying to USE the run buffer, the allocation will now succeed. And, if the each buffer on the run is null, then don't return a `Span<T>` that is not null.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKRunBuffer.cs
+++ b/binding/Binding/SKRunBuffer.cs
@@ -19,13 +19,13 @@ namespace SkiaSharp
 		public int TextSize { get; }
 
 		public Span<ushort> GetGlyphSpan () =>
-			new Span<ushort> (internalBuffer.glyphs, Size);
+			new Span<ushort> (internalBuffer.glyphs, internalBuffer.glyphs == null ? 0 : Size);
 
 		public Span<byte> GetTextSpan () =>
-			new Span<byte> (internalBuffer.utf8text, TextSize);
+			new Span<byte> (internalBuffer.utf8text, internalBuffer.utf8text == null ? 0 : TextSize);
 
 		public Span<uint> GetClusterSpan () =>
-			new Span<uint> (internalBuffer.clusters, Size);
+			new Span<uint> (internalBuffer.clusters, internalBuffer.clusters == null ? 0 : Size);
 
 		public void SetGlyphs (ReadOnlySpan<ushort> glyphs) =>
 			glyphs.CopyTo (GetGlyphSpan ());
@@ -45,7 +45,7 @@ namespace SkiaSharp
 		}
 
 		public Span<float> GetPositionSpan () =>
-			new Span<float> (internalBuffer.pos, Size);
+			new Span<float> (internalBuffer.pos, internalBuffer.pos == null ? 0 : Size);
 
 		public void SetPositions (ReadOnlySpan<float> positions) =>
 			positions.CopyTo (GetPositionSpan ());
@@ -59,7 +59,7 @@ namespace SkiaSharp
 		}
 
 		public Span<SKPoint> GetPositionSpan () =>
-			new Span<SKPoint> (internalBuffer.pos, Size);
+			new Span<SKPoint> (internalBuffer.pos, internalBuffer.pos == null ? 0 : Size);
 
 		public void SetPositions (ReadOnlySpan<SKPoint> positions) =>
 			positions.CopyTo (GetPositionSpan ());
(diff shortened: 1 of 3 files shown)


Ziriax:
Nice! I was making a similar PR, but noticed the repeating pattern:

```cs
var originalEncoding = font.TextEncoding;
			try {
				font.TextEncoding = SKTextEncoding.GlyphId;

				using var lang = new SKString ();
                       ...           
```

and made an internal disposable struct that combined all this into reuseable code:

```cs
using (var scope = new TextBlobGlyphScope(...)) {
     ...scope.lang...
}
```
Maybe that was too much magic 😉

mattleibow:
Oh, nice! Thanks for starting.

Maybe just open the PR with what you have, then we can see if I missed anything. Or even just so there is a record. I was also thinking of a scope thing, but then in v2 the language goes away and all that.

But, I wouldn't spend too much time on this, as the #986 PR basically is almost done and all this is changing with fonts and other things. I am hoping to get a preview of v2 out in March, so the v1 version is more in "bugfixing" mode. I will add things, but I want to focus on v2. Massive improvements there.

Ziriax:
No, you actually covered more than I did :-)

Good to know V2 is almost out.

Is this already usable when building from source?

I am evaluating SkiaSharp for a customer (smartphoto.com), to replace my previous rendering backend written in SharpDX, so I better use the latest version, or I can start my evaluation again  :-)

mattleibow:
I am busy trying to get the build working. It builds locally, but some of the tests are crashing for some reason. Seems to be a leak or a memory thing. If I run it in serial, most tests pass. If I run in in parallel... 💥

PR #1193, opened 2020-03-27
Remove reflection usage for most SKObject

Gillibald:
**Description of Change**

This PR tries to remove the performance impact that comes from reflection usage within the SKObject.GetObject factory

**Bugs Fixed**

- Related to issue #

**API Changes**

Added: 
 
- `void Class.Method();`

Changed:

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


Gillibald:
Only objects that are using `internal static TSkiaObject GetObject<TSkiaObject, TSkiaImplementation>` create their instances via reflection. I wasn't sure how the concrete implementations of SKStream are working.

I would love to know what you think about these changes @mattleibow

mattleibow:
I have been thinking about this for some time now, so it is great that you are having a look too.

I am not a total fan of duplicating the logic in each class, especially when it comes to changes - such as #1200. Is there a way we can use delegates?

Basically, instead of duplicating code, or using reflection, we pass a cached delegate in (this code will be in the new HandleDictionary, but just a demo):

```csharp
class SKObject {
    internal static TSkiaObject GetObject<TSkiaObject, TSkiaImplementation> (
        IntPtr handle, bool owns = true, bool unrefExisting = true, bool refNew = false, Func<IntPtr, bool, TSkiaObject> ctorDel = null)
    {
        if (GetInstance(out instance))
            return instance;
        if (ctorDel != null)
            return ctorDel(handle);
        // do reflection
    }
} 
```

Then in the usage, we can do something similar to what you did with adding a method to each type:

```csharp
class SKSurface {
    public SKCanvas Canvas => SKCanvas.GetObject(sk_surface_canvas());
}

class SKCanvas {
    internal static SKCanvas GetObject (IntPtr handle, bool owns = true, bool unrefExisting = true, bool refNew = false) =>
        SKObject.GetObject<SKCanvas>(handle, owns, unref, ref, (h, o) => new SKCanvas(h, o));
}
```

This way, there is no code duplication (better for maintenance) and no reflection (perf).

This is just me thinking here, so you may spot some issues or maybe improve on this?

Gillibald:
Your solution is indeed more maintainable. Will adapt the delegate approach in this PR and benchmark it.

Gillibald:
Closing this in favor of the concurrency PR. Will work in this when that is finished.

PR #1180, opened 2020-03-19
Re-work views to avoid incorrect disposal

mattleibow:
**Description of Change**

In some instances, the managed canvas is being collected at the same time the new one is being created. This causes the the app to crash for some reason. The technically should not happen as the canvas is still valid - and we can see this when we set the canvas as a class field.

But, for now, just work around the issue by adding the canvas as a field.

In addition, do more granular construction of the GL objects, and then also reduce the amount of interop in the render loop.

**Bugs Fixed**

- Fixes #1121 
- Fixes #910
- Fixes #1173
- Fixes #1168
- Fixes #1144

**API Changes**

None.

**Behavioral Changes**

Some fixes for crashes.

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Updated documentation

Issue #1173, opened 2020-03-12
[BUG] Crashes on Android


hajmajeboss:
**Description**

I'm using mapping library Mapsui, error occurs randomly when moving the map, noticed some of the crashes even with map staying still. 

**Code**

```
I/DEBUG   (  409): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   (  409): UUID: 7bf113ad-494f-44e9-a0fe-61d0704bbee9
I/DEBUG   (  409): Build fingerprint: 'Sony/SGP321/SGP321:5.1.1/10.7.A.0.228/2020576680:user/release-keys'
I/DEBUG   (  409): Revision: '0'
I/DEBUG   (  409): ABI: 'arm'
I/DEBUG   (  409): pid: 19172, tid: 19229, name: GLThread 1370  >>> eu.pds.proplamobilenet <<<
I/DEBUG   (  409): signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1110
I/DEBUG   (  409):     r0 00001110  r1 00000001  r2 9ec01088  r3 00000011
I/DEBUG   (  409):     r4 00000000  r5 12dcc740  r6 9ec01088  r7 9a8d3848
I/DEBUG   (  409):     r8 b74ebc10  r9 b849e9e0  sl 00000000  fp 9a8d3980
I/DEBUG   (  409):     ip afb8acbc  sp 9a8d3838  lr 9974f96c  pc 997d391e  cpsr 600d0030
I/DEBUG   (  409):
I/DEBUG   (  409): backtrace:
I/DEBUG   (  409):     #00 pc 0006191e  /data/app/eu.pds.proplamobilenet-1/lib/arm/libSkiaSharp.so
```

I also noticed an error **with address 0x40,** and p**ossibly 0x1130 (no .so file path in this case)** - can't find the stack trace now.

**Basic Information**

- Version with issue: 1.68.1.1.
- Last known good version: probably 1.60.3
- IDE:  Visual Studio
- Platform Target Frameworks: <
  - Android: API 27 (8.1)
- Target Devices:   
  - Xiaomi Mi A2
  - Sony Xperia Tablet Z

**Reproduction Link**

https://github.com/Mapsui/Mapsui - building a Mapsui.Samples.Droid project and running it on an Android device, trying especially geometries samples .

mattleibow:
[...]

Issue #1168, opened 2020-03-06
[BUG]  Random segfaults on Android


hajmajeboss:
**Description**

I use a mapping library Mapsui, which uses SkiaSharp 1.68.1.1 in my Xamarin.Android app. When zooming in/out or moving around the map, or sometimes even doing nothing, the app crashes. I was able to recover following report from Logcat. 

My app is using AOT compilation.

I was wondering, if you could decode the message using addr2line and find the problem. I'm not entirely sure that it's SkiaSharp error, but I don't know, how to know for sure anyway. I'm using only mono .so libraries besides Skia. 

**Code**

```
03-06 10:45:44.549 10828 11427 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1130 in tid 11427 (GLThread 25266), pid 10828 (proplamobilenet)

03-06 10:45:44.777 11508 11508 I crash_dump64: obtaining output fd from tombstoned, type: kDebuggerdTombstone
03-06 10:45:44.781  1067  1067 I /system/bin/tombstoned: received crash request for pid 11427
03-06 10:45:44.784 11508 11508 I crash_dump64: performing dump of process 10828 (target tid = 11427)
03-06 10:45:44.807 11508 11508 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
03-06 10:45:44.808 11508 11508 F DEBUG   : Build fingerprint: 'xiaomi/jasmine/jasmine_sprout:9/PKQ1.180904.001/V10.0.17.0.PDIMIXM:user/release-keys'
03-06 10:45:44.808 11508 11508 F DEBUG   : Revision: '0'
03-06 10:45:44.808 11508 11508 F DEBUG   : ABI: 'arm64'
03-06 10:45:44.808 11508 11508 F DEBUG   : pid: 10828, tid: 11427, name: GLThread 25266  >>> eu.pds.proplamobilenet <<<

signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1130
x0  0000000000000000  x1  0000000000000001  x2  00000078d1e32d38  x3  00000078d401d678
x4  00000078dbc00000  x5  00000078ce2a2e00  x6  00000078d939b770  x7  00000000038a4ae8
x8  00000000ffffffff  x9  0000000000000001  x10 0000000000000001  x11 00000078e1e70238
[...]

Issue #1144, opened 2020-02-13
[BUG] iOS app freezes randomly due to native crashes


miszu:
**Description**

We're using skia quite extensively in our app - many views, some of them being redrawn often (with constant FPS). Recently we've been observing random freezes on views containing skia components, we think problems began after updating from 1.68 to 1.68.1.1 (it's being validated at the moment, so far problems can't be reproduced with 1.68).

When freeze happens, we observe native crashes like below:

```
=================================================================
    Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================


=================================================================
    Native stacktrace:
=================================================================
2020-02-12 16:35:59.973 TrackMan.SoccerApp.Ios[431:121609] 
    0x1082c090c - /private/var/containers/Bundle/Application/65953D5D-80AF-4DF4-AA8D-569A523C80DE/TrackMan.SoccerApp.Ios.app/TrackMan.SoccerApp.Ios : mono_dump_native_crash_info
    0x1082b66f0 - /private/var/containers/Bundle/Application/65953D5D-80AF-4DF4-AA8D-569A523C80DE/TrackMan.SoccerApp.Ios.app/TrackMan.SoccerApp.Ios : mono_handle_native_crash
2020-02-12 16:36:00.170 TrackMan.SoccerApp.Ios[431:121571] 
    0x1082c4fdc - /private/var/containers/Bundle/Application/65953D5D-80AF-4DF4-AA8D-569A523C80DE/TrackMan.SoccerApp.Ios.app/TrackMan.SoccerApp.Ios : mono_sigsegv_signal_handler_debug
    0x191372894 - /usr/lib/system/libsystem_platform.dylib : <redacted>
    0x11231de70 - /private/var/containers/Bundle/Application/65953D5D-80AF-4DF4-AA8D-569A523C80DE/TrackMan.SoccerApp.Ios.app/Frameworks/libSkiaSharp.framework/libSkiaSharp : gr_backendrendertarget_get_gl_framebufferinfo
    0x11231de70 - /private/var/containers/Bundle/Application/65953D5D-80AF-4DF4-AA8D-569A523C80DE/TrackMan.SoccerApp.Ios.app/Frameworks/libSkiaSharp.framework/libSkiaSharp : gr_backendrendertarget_get_gl_framebufferinfo
[...]

PR #1200, opened 2020-03-29
Resolve some concurrency issues

### LINKED ISSUE #1188: [BUG] macOS native crashes in gr_backendrendertarget_get_gl_framebufferinfo
**Description**
i using SkiaSharp in a Xamarin.Forms app on macOS, after following the discussion in #1144 i have updated to the latest preview .45 which unfortunately doesn't solve the issue with native crashes for me (`gr_backendrendertarget_get_gl_framebufferinfo`). I have a different issue open which is related to `SKImage` #1164 but i'm not sure if this is the same - so here is a new open issue 🤓

 I'm using an `SKGLView` to draw several shapes and images which lead to a freeze and the following native crash.

**Code**

in my case it might be related with `GRContext.GetResourceCacheUsage` 🤷‍♂️

```
=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x107350d28 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_dump_native_crash_info
	0x1073446e5 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_handle_native_crash
	0x107661303 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : altstack_handle_and_restore.cold.1
	0x1072c1dac - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : altstack_handle_and_restore
	0x107d34465 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MonoBundle/libSkiaSharp.dylib : gr_backendrendertarget_get_gl_framebufferinfo
	0x142ab5e43 - Unknown
	0x142ab5c3b - Unknown
	0x142ab4803 - Unknown
	0x14294fec3 - Unknown
	0x142949c63 - Unknown
	0x142946c14 - Unknown
	0x142946ae8 - Unknown
	0x142944bed - Unknown
	0x142943d6b - Unknown
	0x142943f7e - Unknown
	0x10735b1ee - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_jit_runtime_invoke
	0x10748d318 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_runtime_invoke_checked
	0x107491c9e - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_runtime_invoke
	0x1072133d2 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : xamarin_invoke_trampoline
	0x10721499b - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : xamarin_arch_trampoline
	0x107215bc6 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : xamarin_x86_64_common_trampoline
	0x7fff33acfe29 - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSWindow(NSGestureRecognizer_Routing) _sendEventToGestureRecognizers:requireAcceptsFirstMouse:]
	0x7fff33ad7ac2 - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:]
	0x7fff33a41c5c - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:]
	0x7fff33a41005 - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSWindow(NSEventRouting) sendEvent:]
	0x7fff33a3f37c - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSApplication(NSEvent) sendEvent:]
	0x7fff3388b0cf - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : -[NSApplication run]
	0x7fff3385d465 - /System/Library/Frameworks/AppKit.framework/Versions/C/AppKit : NSApplicationMain
	0x1113e8aec - Unknown
	0x1112bc2a3 - Unknown
	0x10879d8fb - Unknown
	0x10879da91 - Unknown
	0x10735b1ee - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_jit_runtime_invoke
	0x10748d318 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_runtime_invoke_checked
	0x1074947c5 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_runtime_exec_main_checked
	0x1072b7e9c - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_jit_exec
	0x1072bafd4 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : mono_main
	0x107216578 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : xamarin_main
	0x1072173e4 - /Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MacOS/Project : main
	0x7fff6dd7a7fd - /usr/lib/system/libdyld.dylib : start

=================================================================
	Telemetry Dumper:
=================================================================
Pkilling 0x70000a25d000 from 0x108a2edc0
Pkilling 0x70000e14f000 from 0x108a2edc0
Pkilling 0x70000a866000 from 0x108a2edc0
Pkilling 0x70000b587000 from 0x108a2edc0
Pkilling 0x70000e758000 from 0x108a2edc0
Pkilling 0x70000ae6f000 from 0x108a2edc0
Pkilling 0x70000bb90000 from 0x108a2edc0
Pkilling 0x70000ed61000 from 0x108a2edc0
Pkilling 0x70000c199000 from 0x108a2edc0
Pkilling 0x70000c7a2000 from 0x108a2edc0
Pkilling 0x70000d740000 from 0x108a2edc0
Pkilling 0x70000e352000 from 0x108a2edc0
Pkilling 0x70000aa69000 from 0x108a2edc0
Pkilling 0x70000b78a000 from 0x108a2edc0
Pkilling 0x70000e95b000 from 0x108a2edc0
Pkilling 0x70000b072000 from 0x108a2edc0
Pkilling 0x70000bd93000 from 0x108a2edc0
Pkilling 0x70000c39c000 from 0x108a2edc0
Pkilling 0x700009d51000 from 0x108a2edc0
Pkilling 0x70000c9a5000 from 0x108a2edc0
Pkilling 0x70000d943000 from 0x108a2edc0
Pkilling 0x70000df4c000 from 0x108a2edc0
Pkilling 0x70000a663000 from 0x108a2edc0
Pkilling 0x70000e555000 from 0x108a2edc0
Pkilling 0x70000ac6c000 from 0x108a2edc0
Pkilling 0x70000b98d000 from 0x108a2edc0
Pkilling 0x70000eb5e000 from 0x108a2edc0
Pkilling 0x70000b275000 from 0x108a2edc0
Pkilling 0x70000bf96000 from 0x108a2edc0
Pkilling 0x700009f54000 from 0x108a2edc0
Pkilling 0x70000cba8000 from 0x108a2edc0
Pkilling 0x70000d1b1000 from 0x108a2edc0
Could not exec mono-hang-watchdog, expected on path '/Users/Documents/Dev/Project/Project.MyProject.MacOS/bin/Debug/Project.app/Contents/MonoBundle/../bin/mono-hang-watchdog' (errno 2)
Entering thread summarizer pause from 0x108a2edc0
Finished thread summarizer pause from 0x108a2edc0.

Waiting for dumping threads to resume

=================================================================
	External Debugger Dump:
=================================================================
socket_transport_send Cannot transition thread 0x108a2edc0 from STATE_BLOCKING with DO_BLOCKING
```
**Expected Behavior**

SKGLView should render without any problems+

**Actual Behavior**

random freezes and stuck app - only force close or `killall lldb` helps in my case

**Basic Information**

- Version with issue:  1.68.2-preview.45
- IDE:  Visual Studio for Mac 8.5 (build 3183)
- Platform Target Frameworks: 
  - macOS: 10.15.3 (19D76)

- Target Devices:   
  - MacBook Pro (13-inch, 2018, Four Thunderbolt 3 Ports) | 16 GB 2133 MHz LPDDR3 | 2,3 GHz Quad-Core Intel Core i5 | Intel Iris Plus Graphics 655 1536 MB

**Reproduction Link**

mattleibow:
**Description of Change**

Resolve some concurrency issues. In some cases the construction and/or destruction overlap and this causes invalid states. This PR improves the locking to avoid this, but still preserves the read and write level of locking.

It also includes a few other improvements, such as making sure objects that require their "parent" object to not be GC'ed add a reference to it. In most cases it is simple, such as a canvas being owned by a surface. But, in the case of a pixmap, the ownership is a little different in that they are copies, but actually require the parent object to stick around. So, the ownership is "reversed".

There is also a few other tweaks to improve general GC reliability/interop performance.

**Bugs Fixed**

- Fixes #1188
- Resolves underlying issue for #1121 #910 #1173 #1168 #1144
- Resolves the underlying issue for the need for #1180

**API Changes**

None.

**Behavioral Changes**

Less crashes due to correct object management.

**PR Checklist**

- [x] Updated documentation

### DIFF
--- a/binding/Binding/HandleDictionary.cs
+++ b/binding/Binding/HandleDictionary.cs
@@ -1,0 +1,215 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+
+namespace SkiaSharp
+{
+	internal static class HandleDictionary
+	{
+		private static readonly Type IntPtrType = typeof (IntPtr);
+		private static readonly Type BoolType = typeof (bool);
+
+#if THROW_OBJECT_EXCEPTIONS
+		internal static readonly ConcurrentBag<Exception> exceptions = new ConcurrentBag<Exception> ();
+#endif
+
+		internal static readonly ConcurrentDictionary<Type, ConstructorInfo> constructors = new ConcurrentDictionary<Type, ConstructorInfo> ();
+		internal static readonly Dictionary<IntPtr, WeakReference> instances = new Dictionary<IntPtr, WeakReference> ();
+
+		internal static readonly ReaderWriterLockSlim instancesLock = new ReaderWriterLockSlim ();
+
+		/// <summary>
+		/// Retrieve the living instance if there is one, or null if not.
+		/// </summary>
+		/// <returns>The instance if it is alive, or null if there is none.</returns>
+		internal static bool GetInstance<TSkiaObject> (IntPtr handle, out TSkiaObject instance)
+			where TSkiaObject : SKObject
+		{
+			if (handle == IntPtr.Zero) {
+				instance = null;
+				return false;
+			}
+
+			instancesLock.EnterReadLock ();
+			try {
+				return GetInstanceNoLocks (handle, out instance);
+			} finally {
+				instancesLock.ExitReadLock ();
+			}
+		}
+
+		/// <summary>
+		/// Retrieve or create an instance for the native handle.
+		/// </summary>
+		/// <returns>The instance, or null if the handle was null.</returns>
+		internal static TSkiaObject GetObject<TSkiaObject, TSkiaImplementation> (IntPtr handle, bool owns = true, bool unrefExisting = true, bool refNew = false)
+			where TSkiaObject : SKObject
+			where TSkiaImplementation : SKObject, TSkiaObject
+		{
+			if (handle == IntPtr.Zero)
+				return null;
+
+			instancesLock.EnterUpgradeableReadLock ();
+			try {
+				if (GetInstanceNoLocks<TSkiaObject> (handle, out var instance)) {
+					// some object get automatically referenced on the native side,
+					// but managed code just has the same reference
+					if (unrefExisting && instance is ISKReferenceCounted refcnt) {
+#if THROW_OBJECT_EXCEPTIONS
+						if (refcnt.GetReferenceCount () == 1)
+							throw new InvalidOperationException (
+								$"About to unreference an object that has no references. " +
+								$"H: {handle.ToString ("x")} Type: {instance.GetType ()}");
+#endif
+						refcnt.SafeUnRef ();
+					}
+
+					return instance;
+				}
+
+				var type = typeof (TSkiaImplementation);
+				var constructor = constructors.GetOrAdd (type, t => GetConstructor (t));
+
+				// we don't need to go into a writable here as the object will do it in the Handle property
+				var obj = (TSkiaObject)constructor.Invoke (new object[] { handle, owns });
+				if (refNew && obj is ISKReferenceCounted toRef)
+					toRef.SafeRef ();
+				return obj;
+			} finally {
+				instancesLock.ExitUpgradeableReadLock ();
+			}
+
+			static ConstructorInfo GetConstructor (Type type)
+			{
+				var ctors = type.GetTypeInfo ().DeclaredConstructors;
+
+				foreach (var ctor in ctors) {
+					var param = ctor.GetParameters ();
+					if (param.Length == 2 && param[0].ParameterType == IntPtrType && param[1].ParameterType == BoolType)
+						return ctor;
+				}
+
+				throw new MissingMethodException ($"No constructor found for {type.FullName}.ctor(System.IntPtr, System.Boolean)");
+			}
+		}
+
+		/// <summary>
+		/// Retrieve the living instance if there is one, or null if not. This does not use locks.
+		/// </summary>
+		/// <returns>The instance if it is alive, or null if there is none.</returns>
+		private static bool GetInstanceNoLocks<TSkiaObject> (IntPtr handle, out TSkiaObject instance)
+			where TSkiaObject : SKObject
+		{
+			if (instances.TryGetValue (handle, out var weak) && weak.IsAlive) {
+				if (weak.Target is TSkiaObject match) {
+					if (!match.IsDisposed) {
+						instance = match;
+						return true;
+					}
+#if THROW_OBJECT_EXCEPTIONS
+				} else if (weak.Target is SKObject obj) {
+					if (!obj.IsDisposed) {
+						throw new InvalidOperationException (
+							$"A managed object exists for the handle, but is not the expected type. " +
+							$"H: {handle.ToString ("x")} Type: ({obj.GetType ()}, {typeof (TSkiaObject)})");
+					}
+				} else if (weak.Target is object o) {
+					throw new InvalidOperationException (
+						$"An unknown object exists for the handle when trying to fetch an instance. " +
+						$"H: {handle.ToString ("x")} Type: ({o.GetType ()}, {typeof (TSkiaObject)})");
+#endif
+				}
+			}
+
+			instance = null;
+			return false;
+		}
+
+		/// <summary>
+		/// Registers the specified instance with the dictionary.
+		/// </summary>
+		internal static void RegisterHandle (IntPtr handle, SKObject instance)
+		{
+			if (handle == IntPtr.Zero || instance == null)
+				return;
+
+			SKObject objectToDispose = null;
+
+			instancesLock.EnterWriteLock ();
+			try {
+				if (instances.TryGetValue (handle, out var oldValue) && oldValue.Target is SKObject obj && !obj.IsDisposed) {
+#if THROW_OBJECT_EXCEPTIONS
+					if (obj.OwnsHandle) {
+						// a mostly recoverable error
+						// if there is a managed object, then maybe something happened and the native object is dead
+						throw new InvalidOperationException (
+							$"A managed object already exists for the specified native object. " +
+							$"H: {handle.ToString ("x")} Type: ({obj.GetType ()}, {instance.GetType ()})");
+					}
+#endif
+					// this means the ownership was handed off to a native object, so clean up the managed side
+					objectToDispose = obj;
+				}
+
+				instances[handle] = new WeakReference (instance);
+			} finally {
+				instancesLock.ExitWriteLock ();
+			}
+
+			// dispose the object we just replaced
+			objectToDispose?.DisposeInternal ();
+		}
+
+		/// <summary>
+		/// Removes the registered instance from the dictionary.
+		/// </summary>
+		internal static void DeregisterHandle (IntPtr handle, SKObject instance)
+		{
+			if (handle == IntPtr.Zero)
+				return;
+
+			instancesLock.EnterWriteLock ();
+			try {
+				var existed = instances.TryGetValue (handle, out var weak);
+				if (existed && (!weak.IsAlive || weak.Target == instance)) {
+					instances.Remove (handle);
+				} else {
+#if THROW_OBJECT_EXCEPTIONS
+					InvalidOperationException ex = null;
+					if (!existed) {
+						// the object may have been replaced
+
+						if (!instance.IsDisposed) {
+							// recoverable error
+							// there was no object there, but we are still alive
+							ex = new InvalidOperationException (
+								$"A managed object did not exist for the specified native object. " +
+								$"H: {handle.ToString ("x")} Type: {instance.GetType ()}");
+						}
+					} else if (weak.Target is SKObject o && o != instance) {
+						// there was an object in the dictionary, but it was NOT this object
+
+						if (!instance.IsDisposed) {
+							// recoverable error
+							// there was a new living object there, but we are still alive
+							ex = new InvalidOperationException (
+								$"Trying to remove a different object with the same native handle. " +
+								$"H: {handle.ToString ("x")} Type: ({o.GetType ()}, {instance.GetType ()})");
+						}
+					}
+					if (ex != null) {
+						if (instance.fromFinalizer)
+							exceptions.Add (ex);
+						else
+							throw ex;
+					}
+#endif
+				}
+			} finally {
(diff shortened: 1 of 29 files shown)


binding/Binding/HandleDictionary.cs
```diff
@@ -0,0 +1,215 @@
...
+		private static readonly Type BoolType = typeof (bool);
+
+#if THROW_OBJECT_EXCEPTIONS
+		internal static readonly ConcurrentBag<Exception> exceptions = new ConcurrentBag<Exception> ();
+#endif
+
+		internal static readonly ConcurrentDictionary<Type, ConstructorInfo> constructors = new ConcurrentDictionary<Type, ConstructorInfo> ();
+		internal static readonly Dictionary<IntPtr, WeakReference> instances = new Dictionary<IntPtr, WeakReference> ();
+
+		internal static readonly ReaderWriterLockSlim instancesLock = new ReaderWriterLockSlim ();
```
mattleibow: Instead of pure locking and concurrent dictionaries, just use a reader/writer lock as that is what we actually want. And, when reading and we find that there is no handle, then we can upgrade in place.

binding/Binding/HandleDictionary.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+
+namespace SkiaSharp
+{
+	internal static class HandleDictionary
+	{
+		private static readonly Type IntPtrType = typeof (IntPtr);
+		private static readonly Type BoolType = typeof (bool);
```
mattleibow: Caching on .NET Core is a major perf boost. Major as in 3x, but still in ns.

binding/Binding/HandleDictionary.cs
```diff
@@ -0,0 +1,215 @@
...
+								$"H: {handle.ToString ("x")} Type: {instance.GetType ()}");
+#endif
+						refcnt.SafeUnRef ();
+					}
+
+					return instance;
+				}
+
+				var type = typeof (TSkiaImplementation);
+				var constructor = constructors.GetOrAdd (type, t => GetConstructor (t));
```
mattleibow: Anonymous delegates are cached rather than direct method references.

binding/Binding/SKObject.cs
```diff
@@ -1,29 +1,23 @@
...
-#if THROW_OBJECT_EXCEPTIONS
-		internal static readonly ConcurrentBag<Exception> exceptions = new ConcurrentBag<Exception> ();
-#endif
+		private ConcurrentDictionary<IntPtr, SKObject> ownedObjects;
+		private ConcurrentDictionary<IntPtr, SKObject> keepAliveObjects;
 
-		internal static readonly ConcurrentDictionary<Type, ConstructorInfo> constructors;
-		internal static readonly ConcurrentDictionary<IntPtr, WeakReference> instances;
+		internal ConcurrentDictionary<IntPtr, SKObject> OwnedObjects =>
+			ownedObjects ??= new ConcurrentDictionary<IntPtr, SKObject> ();
```
mattleibow: Lazy-instantiate this as it is hardly ever used. Typically for documents and some for some streams, codecs and typefaces.

tests/SkiaSharp.NetCore.Tests/SkiaSharp.NetCore.Tests.csproj
```diff
@@ -12,18 +12,13 @@
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
     <PackageReference Include="XunitXml.TestLogger" Version="2.1.26" />
-    <PackageReference Include="xunit.assemblyfixture" Version="2.0.3" />
```
mattleibow: This never worked for .NET Core anyway. Use the xUnit-approved way from the xUnit samples.

mattleibow:
Thanks for the review!

Technically, the objects should only be referenced from a single thread at any given time. However, I will add some locking to prevent any future issues.

binding/Binding/SKBitmap.cs
```diff
@@ -718,7 +718,10 @@ public bool PeekPixels (SKPixmap pixmap)
 			if (pixmap == null) {
 				throw new ArgumentNullException (nameof (pixmap));
 			}
-			return SkiaApi.sk_bitmap_peek_pixels (Handle, pixmap.Handle);
+			var result = SkiaApi.sk_bitmap_peek_pixels (Handle, pixmap.Handle);
+			if (result)
+				Referenced (pixmap, this);
```
mattleibow, line 723: The GC sometimes likes to collect the bitmap in use by a pixmap. This case is important because the bitmap OWNS the pixmap, and the reference is not counted. So a disposal destroys the pixmap.

binding/Binding/SKData.cs
```diff
@@ -251,6 +251,7 @@ public void SaveTo (Stream target)
 			} finally {
 				ArrayPool<byte>.Shared.Return (buffer);
 			}
+			GC.KeepAlive (this);
```
mattleibow, line 254: Just prevent the GC from going to town, because we are not actually using the object, but a value from one of the properties.

binding/Binding/SKObject.cs
```diff
@@ -1,29 +1,41 @@
...
 using System.Threading;
 
 namespace SkiaSharp
 {
 	public abstract class SKObject : SKNativeObject
 	{
-#if THROW_OBJECT_EXCEPTIONS
-		internal static readonly ConcurrentBag<Exception> exceptions = new ConcurrentBag<Exception> ();
-#endif
+		private readonly object locker = new object ();
```
mattleibow, line 10: A single locker as this doesn't need to be too complicated.

mattleibow:
I think this is it. Fixed some other bugs. Concurrency due to construction and destruction, but also when the GC starts to collect things too soon.

mattleibow:
I have pushed what I think is now a good version to the preview feed: https://aka.ms/skiasharp-eap/index.json
Check out the PR version: **1.68.2-pr.1200.20**

Gillibald:
Will have a look shortly 👍

mattleibow:
@Gillibald have you been able to have a look and see if things are better? I am thinking of just merging and pushing to nuget.org so that the mapsui folks can check. But, I just want to make sure that I get your opinion/hands on the pie here.

Gillibald:
Did not have time today. Could do tomorrow morning. In a few hours 😀. My current test creates and disposes millions of SKObjects. So minimal improvements should come out clearly. Mostly text related objects.

mattleibow:
Ah, millions? Have you noticed any random crashes or weird behavior at all 😄 

The issues that mostly come up is with concurrent execution because objects were being created with the same handle as an object being destroyed. As a result, the managed side got all mixed up as it was busy nuking things and then it is added back, or worse the reverse. Most of the cases coming now are when people use the GL views because that is usually for high frequency updates and often on separate threads. 

So this would be interesting to see if you were seeing weird things, or even having to do some weird things to prevent weird things 😄

Gillibald:
Luckily I only have to deal with single threaded processing. I might want to put text processing on a dedicated thread in the near future.

Gillibald:
The preview runs just fine. After this is merged I will work on the changes to SKObject.GetObject
Using cached delegates instead of reflection should improve things a bit.

Do we really need to keep track of all SKObject instances? Ideally, we could just turn off caching.

mattleibow:
We have to track them because if I call `SKSurface.Canvas` twice, I need to find it again. I can't really cache it directly, because some operations might destroy the canvas and then the object is destroyed in native code, but we still got a valid object in managed world. Sometimes we don't even know when it will happen, as it may be a result of another operation. It is easy from a developer's perspective because certain operations may do things, but to get it right it is going to be more difficult.

I know @kekekeks was asking for a pointer-only API: #876 

But, this is not to say we can't actually have a look at certain APIs and see if we can just store things in a field. For example, a surface will always have the same canvas, so no need to even go through the interop. But this is not true for `SKBitmap.PeekPixels` because a copy is made of the pixmap, and you can always change the underlying pixel pointer at time - and this will not reflect in the copied pixmap.

Overall, the slowest point of this library is the lookup. And this is made harder because some objects are reference counted, others are not. And then, some are reference counted, but not used like one. The API is not quite perfect.

mattleibow:
Thanks for having a look. Merging this now so I can get a preview out.

PR #1209, opened 2020-04-03
 Remove reflection usage for SKObject creation

Gillibald:
**Description of Change**

Describe your changes here. 

**Bugs Fixed**

- Related to issue #

**API Changes**

Added: 
 
- `void Class.Method();`

Changed:

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/HandleDictionary.cs
+++ b/binding/Binding/HandleDictionary.cs
@@ -35,40 +35,39 @@ namespace SkiaSharp
 			}
 		}
 
+		/// <summary>
+		/// Retrieve or create an instance for the native handle.
+		/// </summary>
+		/// <returns>The instance, or null if the handle was null.</returns>
 		internal static TSkiaObject GetOrAddObject<TSkiaObject> (IntPtr handle, bool owns, bool unrefExisting, bool refNew, Func<IntPtr, bool, TSkiaObject> objectFactory)
 			where TSkiaObject : SKObject
 		{
-			if (handle == IntPtr.Zero) {
+			if (handle == IntPtr.Zero)
 				return null;
-			}
-
-			TSkiaObject obj = null;
 
 			instancesLock.EnterUpgradeableReadLock ();
-
 			try {
-				if (GetInstanceNoLocks<TSkiaObject> (handle, out obj)) {
+				if (GetInstanceNoLocks<TSkiaObject> (handle, out var instance)) {
 					// some object get automatically referenced on the native side,
 					// but managed code just has the same reference
-					if (unrefExisting && obj is ISKReferenceCounted refcnt) {
+					if (unrefExisting && instance is ISKReferenceCounted refcnt) {
 #if THROW_OBJECT_EXCEPTIONS
 						if (refcnt.GetReferenceCount () == 1)
 							throw new InvalidOperationException (
 								$"About to unreference an object that has no references. " +
-								$"H: {handle.ToString ("x")} Type: {obj.GetType ()}");
+								$"H: {handle.ToString ("x")} Type: {instance.GetType ()}");
 #endif
 						refcnt.SafeUnRef ();
 					}
 
-					return obj;
+					return instance;
 				}
 
-				obj = objectFactory.Invoke (handle, owns);
+				var obj = objectFactory.Invoke (handle, owns);
 
 				if (refNew && obj is ISKReferenceCounted toRef)
 					toRef.SafeRef ();
 				return obj;
-
 			} finally {
 				instancesLock.ExitUpgradeableReadLock ();
 			}
--- a/binding/Binding/HandleDictionary.cs
+++ b/binding/Binding/HandleDictionary.cs
@@ -1,21 +1,16 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Reflection;
 using System.Threading;
 
 namespace SkiaSharp
 {
 	internal static class HandleDictionary
 	{
-		private static readonly Type IntPtrType = typeof (IntPtr);
-		private static readonly Type BoolType = typeof (bool);
 
 #if THROW_OBJECT_EXCEPTIONS
 		internal static readonly ConcurrentBag<Exception> exceptions = new ConcurrentBag<Exception> ();
 #endif
-
-		internal static readonly ConcurrentDictionary<Type, ConstructorInfo> constructors = new ConcurrentDictionary<Type, ConstructorInfo> ();
 		internal static readonly Dictionary<IntPtr, WeakReference> instances = new Dictionary<IntPtr, WeakReference> ();
 
 		internal static readonly ReaderWriterLockSlim instancesLock = new ReaderWriterLockSlim ();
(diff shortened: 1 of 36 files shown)


mattleibow:
Just a unit test failure due... because we no longer have to have a test for a field:

         D:\a\1\s\tests\Tests\GarbageCleanupFixture.cs(23,34): error CS0117: 'HandleDictionary' does not contain a definition for 'constructors' [D:\a\1\s\tests\SkiaSharp.Desktop.Tests\SkiaSharp.Desktop.Tests.csproj]

Gillibald:
Woops. Will fix that shortly.

mattleibow:
Pity there is no generic constraint for complex constructors `new(IntPtr, bool)` 😢

Gillibald:
Yep but we could just introduce parameterless ctor and move the current logic to a internal virtual method I guess.

mattleibow:
Hmmmm. We could do that.

Already the code is set up to support setting the handle later in the construction. So technically, we can do:

```csharp
var obj = new SKSomething();
obj.SetHandle(handle: handle, owns: true);
```

The issue is with already-default constructors, like `new SKPaint()`. That already constructs a new native object. From user code.

Not sure how to deal with that...

Gillibald:
Might a special paint help here? `ReferencedSKPaint : SKPaint` that is used by code that wants to expose a already existing instance?

Gillibald:
Okay. A parameterless ctor will not work because it has to be public. So the delegate approach is the best we can get for now without duplicating code.

Gillibald:
Haven't fixed `GarbageCleanupFixture` yet

mattleibow:
No prob. I was just updating from master with some more unit tests for other things.

mattleibow:
What project were you using to make these changes? Was it the source project? I managed to fix the unit tests in VS2019 (I think) so it might also be cool to just open the .NET Core tests and then use them as you work on things. And you can use it to test chunks of code. Not sure what is happening with the full framework test not running in VS, but it work on the command line so no biggie.

Gillibald:
I am working with the Netcore Tests project. That isn't including all tests.

Gillibald:
Do we get a public ci build again that let's us see the failing tests?

mattleibow:
We were working on that, but we also keeping our usage of DevOps down so we can let the COVID-19 responders get priority. 

_I say we smack this virus a big one and scare nature into a state where it doesn't want to pop another for a very long time 😆 (my expert medical opinion)_

Gillibald:
What did fail this time?

mattleibow:
DevOps. Sometimes it just fails to download the Windows SDK or some random thing. Restarting it. Sometimes the asset copy between stages also just forgets to copy a file.

mattleibow:
I have been thinking about this PR... The delegates inline make is easier to maintain, but still a bit... well... not awesome.

I did like the way you did the static methods of `GetObject` in each type in https://github.com/mono/SkiaSharp/pull/1193. Maybe we could make the new, local `GetObject` just call the `SKObject` version:

```csharp
internal static GRContext GetObject (IntPtr ptr, bool owns = true, bool unrefExisting = true) =>
    SKObject.GetObject<GRContext>(handle, (h, o) => new GRContext(h, o), owns, unrefExisting);
```

This will also reduce the amount of IL generated for the thousands of delegates. I suppose that is less duplication and we just have to worry about the method signature, which is easy.

I know you have been working on this for several days and it is a pain to keep going back. So, thanks for sticking with it. Did you managed to run any perfs? It will be interesting to see how much the improvements are.

Gillibald:
I am fine with redoing this. Your suggestion reduces the code noise so I am fine with it.

I wonder if we should get rid of the delegate as well if we introduce a helper method on each type. 

Creating object before this change took 3% after this change it was 1%. Destruction of objects still takes about 4% of all processing.

In general SkiaSharp had less impact on the overall processing time. 

I can run some benchmarks if you want.

mattleibow:
Interesting that destruction takes more time... Do you have any idea where this overhead is? I will run some more profiling and see... Has there been any variance with types of objects/operations?

Gillibald:
I am calling SKFontManager.MatchCharacter quite often. That produces `SKFontStyle` on every call. It is really bad to cache these one time used instances.

mattleibow:
> That produces SKFontStyle on every call.

I saw that and now I cache that now: https://github.com/mono/SkiaSharp/blob/master/binding/Binding/SKFontStyle.cs#L55

But, did you try the overloads that take a parameter?
https://github.com/mono/SkiaSharp/blob/fc9f030a08eb158fb934774cdf54f8df939088c7/binding/Binding/SKFontManager.cs#L170

Gillibald:
Y I will cache them on my own in the future still I don't see why SKFontStyle isn't a struct.

Gillibald:
I will fix the unit tests tomorrow. When we agree on this solution I will fix coding style issues.

mattleibow:
OK, I'll review. Thanks for all the effort. It is much appreciated.

binding/Binding/HandleDictionary.cs
```diff
@@ -40,59 +35,38 @@ internal static bool GetInstance<TSkiaObject> (IntPtr handle, out TSkiaObject in
...
-		internal static TSkiaObject GetObject<TSkiaObject, TSkiaImplementation> (IntPtr handle, bool owns = true, bool unrefExisting = true, bool refNew = false)
+		internal static bool TryGetObject<TSkiaObject, TSkiaImplementation> (IntPtr handle, bool owns, bool unrefExisting, bool refNew, out TSkiaImplementation obj)
 			where TSkiaObject : SKObject
 			where TSkiaImplementation : SKObject, TSkiaObject
 		{
-			if (handle == IntPtr.Zero)
-				return null;
+			if (handle == IntPtr.Zero) {
+				obj = null;
+				return false;
```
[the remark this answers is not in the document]
Gillibald: Woops. You are right null is a valid value. I shouldn't rush such a refactoring. Good that we can sort out such errors in a review and the power of unit tests 😅

binding/Binding/HandleDictionary.cs
```diff
@@ -40,59 +35,37 @@ internal static bool GetInstance<TSkiaObject> (IntPtr handle, out TSkiaObject in
...
 
-					return instance;
+					return true;
 				}
-
-				var type = typeof (TSkiaImplementation);
-				var constructor = constructors.GetOrAdd (type, t => GetConstructor (t));
-
-				// we don't need to go into a writable here as the object will do it in the Handle property
-				var obj = (TSkiaObject)constructor.Invoke (new object[] { handle, owns });
```
mattleibow, line 76: Just noticed the construction is no longer inside the `instancesLock`. This might allow for concurrent construction of the same object. We might need that delegate 😢

Gillibald: Hmm

Gillibald: `internal static TSkiaObject GetOrAddObject<TSkiaObject> (IntPtr handle, bool owns, bool unrefExisting, bool refNew, Func<IntPtr, bool, TSkiaObject> objectFactory)`

Gillibald: Okay will adjust the implementation a bit

Gillibald: `internal static GRContext GetObject (IntPtr handle) => GetOrAddObject (handle, (h,o) => new GRContext (h, o));`

mattleibow: Yes, that looks like it should work now. And the delegate is reused instead of for each member.

Gillibald:
Okay. Next try 👍

mattleibow:
I did some benchmarks and things are looking like they are about 1.3x faster!

```
BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4150.0), X64 RyuJIT
  Job-VTGAFA : .NET Framework 4.8 (4.8.4150.0), X64 RyuJIT
  Job-MNUBEA : .NET Core 3.1.2 (CoreCLR 4.700.20.6602, CoreFX 4.700.20.6702), X64 RyuJIT
  Job-XIDDXQ : Mono 6.8.0 (Visual Studio), X64
```

**BEFORE**

```
|  Method |       Runtime |     Mean |     Error |    StdDev |
|-------- |-------------- |---------:|----------:|----------:|
| TheTest |    .NET 4.7.2 | 1.369 us | 0.0211 us | 0.0197 us |
| TheTest | .NET Core 3.1 | 1.178 us | 0.0233 us | 0.0229 us |
| TheTest |          Mono | 2.811 us | 0.0536 us | 0.0550 us |
```
**AFTER**

```
|  Method |       Runtime |     Mean |     Error |    StdDev |
|-------- |-------------- |---------:|----------:|----------:|
| TheTest |    .NET 4.7.2 | 1.051 us | 0.0172 us | 0.0143 us |
| TheTest | .NET Core 3.1 | 0.940 us | 0.0181 us | 0.0169 us |
| TheTest |          Mono | 1.912 us | 0.0186 us | 0.0156 us |
```

**CODE**

```csharp
[SimpleJob(RuntimeMoniker.Mono)]
[SimpleJob(RuntimeMoniker.Net472)]
[SimpleJob(RuntimeMoniker.NetCoreApp31)]
public class Benchmark {
    [Benchmark]
    public object TheTest() {
        // the actual test
        var data = SKData.Create(1);

        data.Dispose();
        return data;
    }
}

public class Program {
    public static void Main(string[] args) {
        var summary = BenchmarkRunner.Run<Benchmark>();
    }
}
```

Gillibald:
Hmm that isn't much but at least this will free us from reflection.

mattleibow:
Not much, but it might add up on larger drawings. This is a 1.3x speed up on almost every static construction and many members. 
And this is just .NET desktop. Who knows what actually happens on iOS or other AOT systems.

mattleibow:
Thanks for this PR, took a few runs but we got it all done and we now have a slightly faster SkiaSharp.

Gillibald:
I am fine with redoing things to get the best results. Will start experimenting with the V2 bits soon so we get that stable soon.

Issue #1227, opened 2020-04-11
[QUESTION]  Random unit test crashes in dev/update-skia?


Ziriax:
I am trying to run the units tests in the `dev/update-skia` branch, so I can continue working on PRs, but my `dotnet test` run always fails randomly, with native exceptions (access violations). Sometimes a dozens of tests succeed before crashing, sometimes 50, so this smells like a multi threading issues. I have 32 hyperthreads (AMD 3950x) all running at the same time, so this might reveal race conditions.

Is this a known issue? 

I have build both the native and managed code successfully, on Windows 10, using the latest version of VS2019.

Ziriax:
If I re-run only the failed tests, they eventually all pass.

So clearly some shared state is getting corrupted here.

mattleibow:
Yeah, I notice that as well. I have been working in master to fix tests so they all run locally. I will merge into update soon.

But that won't fix them all. In some cases, the rules have changed. For example, in one case the base types have changed in the native code (`SkColorSpace` went from virtual to non-virtual). In another case, the method return values change from a copy to a reference and vice versa. Only the native comments show this.

So I just have to run various combos until I find a repro crash and then try and figure it out.

What was crashing? What was the native pinvoke?

Another thing I have noticed that the `this` object sometimes gets collected while a method is running. The one I saw last night was:

```csharp
class SKTextBlobBuilder {
	public SKTextBlob Build () =>
		GetObject<SKTextBlob> (SkiaApi.sk_textblob_builder_make (Handle));
}
```

`SkiaApi.sk_textblob_builder_make` crashes because the object is collected and disposed while the native member is running.

There are 2 fixes for this type:
 - use `GC.KeepAlive(this)` at the end of the method
    - this is verbose 
    - most methods will have to be done
 - use a `SafeHandle`
    - this may have performance implications

Ziriax:
Just to be sure, the `dev/update-skia` branch is the current beta that will become the next release right?

I can reproduce one problem at least by putting a for loop around part of the test.

For example:
```cs
		[SkippableFact]
		public void StreamLosesOwnershipToCodecButIsNotForgotten()
		{
			var bytes = File.ReadAllBytes(Path.Combine(PathToImages, "color-wheel.png"));
			var stream = new SKMemoryStream(bytes);
			var handle = stream.Handle;

			Assert.True(stream.OwnsHandle);
			Assert.True(SKObject.GetInstance<SKMemoryStream>(handle, out _));

[...]

PR #1228, opened 2020-04-11
Fixed unit tests crashes and a WGL deadlock

Ziriax:
**Description of Change**

I wasn't able to run the unit tests in dev/update-skia dbfa12b644099e78cc4e46e226daba332f204f01

Got random crashes and deadlocks.

**Bugs Fixed**

Fixed the crashes and the deadlock I encountered.

- Related to issue #1227 

**API Changes**

No public changes, but I removed the `bool refNew = false` argument in the `GetObject` method, since it was always passed as `false`, so it felt redundant. This can of course change, but the `owns` and `refNew` arguments felt conflicting: should a reference ever be incremented if the wrapper doesn't own the object in the first place? Felt wrong.

**Behavioral Changes**

It is possible we now have memory leaks, I didn't check that...

**PR Checklist**

- [x] Rebased on top of dev/update-skia at time of PR
- [ ] Updated documentation


Ziriax:
My original patch doesn't seem to work anymore with the latest version

mattleibow:
Thanks for this!

It was merged in from #1237 and is now included in master (v1.68) and develop (v2.80).

PR #1237, opened 2020-04-17
Fixed unit tests crashes and a WGL deadlock (#1228 for master)

mattleibow:
**Description of Change**

Fixed unit tests crashes and a WGL deadlock

**Bugs Fixed**

- Related to issue #1227
- Related to PR #1228

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Updated documentation


mattleibow:
I am looking into the cause of the crash, I think they changed something between this and the next version. Your test case crashes every time on the m80, but not at all on the m68. Having a look at the native API now, but I do see a difference:

m68 (https://github.com/google/skia/blob/chrome/m68/include/codec/SkCodec.h#L174):

```cpp
const SkImageInfo& getInfo() const;
```
m80 (https://github.com/google/skia/blob/chrome/m80/include/codec/SkCodec.h#L194):

```cpp
SkImageInfo getInfo() const;
```

mattleibow:
Owns handle is just whether the object is responsible for deconstructing the reference/object. In the case of the statics, the initial reference is incremented, therefore we have to also decrement when it is collected.

But, you were almost on the exact right track. The object is reference counted, but in the case of the particular call of `getInfo`, I think it is not incremented. In that case, we need to let the managed side know that is should not decrement (using the `unrefExisting` param):

https://github.com/mono/SkiaSharp/blob/dev/update-skia/binding/Binding/SKImageInfo.cs#L27

But, I am just going to do some tests quick and confirm.

Ziriax:
> Owns handle is just whether the object is responsible for deconstructing the reference/object. In the case of the statics, the initial reference is incremented, therefore we have to also decrement when it is collected.
> 
> But, you were almost on the exact right track. The object is reference counted, but in the case of the particular call of `getInfo`, I think it is not incremented. In that case, we need to let the managed side know that is should not decrement (using the `unrefExisting` param):
> 
> https://github.com/mono/SkiaSharp/blob/dev/update-skia/binding/Binding/SKImageInfo.cs#L27
> 
> But, I am just going to do some tests quick and confirm.

Oh nice catch. So this is way more complex than I thought.

A look at the [implementation](https://source.chromium.org/chromium/chromium/src/+/master:third_party/skia/include/private/SkEncodedInfo.h;drc=74778f2d1a27e6349c10117ef6ba53adda8c6db1;l=159?originalUrl=https:%2F%2Fcs.chromium.org%2F) shows that a new copy is always returned by `SkCodec::getInfo()`:

```cpp
  */
    SkImageInfo makeImageInfo() const {
        auto ct =  kGray_Color == fColor ? kGray_8_SkColorType   :
                 kXAlpha_Color == fColor ? kAlpha_8_SkColorType  :
                    k565_Color == fColor ? kRGB_565_SkColorType  :
                                           kN32_SkColorType      ;
        auto alpha = kOpaque_Alpha == fAlpha ? kOpaque_SkAlphaType
                                             : kUnpremul_SkAlphaType;
        sk_sp<SkColorSpace> cs = fProfile ? SkColorSpace::Make(*fProfile->profile())
                                          : nullptr;
        if (!cs) {
            cs = SkColorSpace::MakeSRGB();
        }
        return SkImageInfo::Make(fWidth, fHeight, ct, alpha, std::move(cs));
    }
```

mattleibow:
Moving the conversation to your original issue #1227

PR #1242, opened 2020-04-19
Expose a generic, writable span for the pixels

### LINKED ISSUE #1240: [QUESTION] Is SKPixMap.GetPixelSpan supposed to be readonly?
`SKPixMap.GetPixelSpan` returns a  `ReadOnlySpan<byte>`, not a `Span<byte>`

So one can't use that for direct pixel manipulation.

Is this by design? I believe a pix-map is writable no?

Skia itself also has `void* SKPixMap::writable_addr() `.

It would also be useful to have a generic version, to avoid having to call `MemoryMarshal` methods

mattleibow:
**Description of Change**

Expose a generic, writable span for the pixels. Also "cache" the info object in size so we don' have to hop the interop.

**Bugs Fixed**

- Fixes #1240

**API Changes**

```csharp
class SKPixmap {
    public unsafe Span<T> GetPixelSpan<T> () where T : unmanaged;
}
```

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKPixmap.cs
+++ b/binding/Binding/SKPixmap.cs
@@ -95,9 +95,14 @@ namespace SkiaSharp
 
 		public int Height => Info.Height;
 
-		public SKSizeI Size => new SKSizeI (Width, Height);
+		public SKSizeI Size {
+			get {
+				var info = Info;
+				return new SKSizeI (info.Width, info.Height);
+			}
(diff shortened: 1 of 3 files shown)


Ziriax:
Seems perfect at first sight :)

PR #1254, opened 2020-04-27
Make the GL views a "zero allocation draw"

### LINKED ISSUE #1251: [BUG] (Performance) SkSwapChaninPanel makes unnecessary per-frame heap allocations
SkSwapChainPanel::OnRenderFrame(Rect) makes 2 heap allocations. It would be preferable to not have any allocations in the hot path.

1. SkAutoCanvasRestore(...), The RAII semantics are neat, but maybe this could be changed to one of the new disposable stack-scoped structs?
2. Can SKPaintGLSurfaceEventArgs be reused rather than allocated each frame? 

https://github.com/mono/SkiaSharp/blob/354d85d129e898071e002a3157e7fcba1d7830a7/source/SkiaSharp.Views/SkiaSharp.Views.UWP/SKSwapChainPanel.cs#L85

mattleibow:
**Description of Change**

Reduce the allocations in a draw loop of the GL views. 

Or at least see what we can do and provide options.

**Bugs Fixed**

- Fixes #1251 

**API Changes**

Added: 
 
- `void Class.Method();`

Changed:

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.UWP/SKSwapChainPanel.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.UWP/SKSwapChainPanel.cs
@@ -88,12 +88,15 @@ namespace SkiaSharp.Views.UWP
 				lastPaintEventArgs = null;
 			}
 
-			// start drawing
+			// create or recreate the event args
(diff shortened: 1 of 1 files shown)


mattleibow:
We can have a look at this later.

PR #1257, opened 2020-04-28
No need to allocate on disposal

mattleibow:
**Description of Change**

No need to allocate on disposal. The lazy property must not be used, but rather the field.

**Bugs Fixed**

None.

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKObject.cs
+++ b/binding/Binding/SKObject.cs
@@ -71,7 +71,9 @@ namespace SkiaSharp
 				}
 				dic.Clear ();
 			}
-			KeepAliveObjects?.Clear ();
+			if (keepAliveObjects is ConcurrentDictionary<IntPtr, SKObject> ka) {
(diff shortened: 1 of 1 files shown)


binding/Binding/SKObject.cs
```diff
@@ -71,7 +71,9 @@ protected override void DisposeManaged ()
 				}
 				dic.Clear ();
 			}
-			KeepAliveObjects?.Clear ();
+			if (keepAliveObjects is ConcurrentDictionary<IntPtr, SKObject> ka) {
+				ka.Clear ();
+			}
```
mattleibow, line 76: Using the property will always construct a fairly expensive object. And we are doing it in the Dispose. This means that we are ALWAYS constructing this for EVERY object.

PR #1259, opened 2020-04-28
First version of high quality sub-pixel precision image downsampler

Ziriax:
**Description of Change**

- Currently uses Skia for upsampling (IMHO upsampling always looks blurry anyway)
- Not optimized yet, just uses `Vector4` from the `System.Numerics` package.  
   - Most of the time is actually spent in Skia's color type conversion...
   - I have optimized AVX2 code ready that short circuits the most common case (speedup x5), but I first wanted to share a "minimal" working example, without unsafe pointers, to get feedback
- The code currently always uses a `Parallel.For`, but by default only uses a single thread. 
- I should add a lot more comments about how this all works, but I don't understand all the details myself, this is one week of pragmatic coding ;-)
- I struggled a lot to verify the results. I reported some bugs to Skia regarding PNG image encoding bugs. Thanks for the GIMP tip, that one always worked correctly.
- Since Skia doesn't have these methods, I decided to use extensions methods, and put the code in a separate directory.
- The `outputWidth` and `outputHeight` are `float`s, to allow fractional sizes (just like `SKCanvas.DrawImage`  can draw into a fractional rectangle). 
- The `SKImageSamplingOptions.OutputOffset` allows to control the sub-pixel offsets, if needed, and allows setting other options.
- I added an example in the gallery. The baboon image is actually a good image, clearly showing the difference in quality between Skia's native hybrid box-resampler/resizer and this Lanczos based resampler.
- Currently always uses a Lanczos filter (good average filter), but allows to provide external ones.
- [To improve quality](https://github.com/avaneev/avir), the image is first upscaled `x2`, but this is completely integrated in the filter weights, so has zero overhead. 
- I haven't done any testing on ARM devices yet (not sure how to do this). 
- I was unable to run all unit tests, but I didn't have to change any existing API...

**Bugs Fixed**

Skia's default image resizing can be blurry.

- Related to issue ##1127 

**API Changes**

Added: 

- `public static SKBitmap ResampledBy (this SKBitmap srcBitmap, float scaleFactor, SKImageSamplingOptions options = null)`
- `public static SKBitmap ResampledBy (this SKBitmap srcBitmap, float widthFactor, float heightFactor, SKImageSamplingOptions options = null)`
- `public static SKBitmap ResampledTo (this SKBitmap srcBitmap, float outputWidth, float outputHeight, SKImageSamplingOptions options = null)`
- `public static SKBitmap ResampledBy (this SKPixmap srcPixmap, float scaleFactor, SKImageSamplingOptions options = null)`
- `public static SKBitmap ResampledBy (this SKPixmap srcPixmap, float widthFactor, float heightFactor, SKImageSamplingOptions options = null)`
- `public static SKBitmap ResampledTo (this SKPixmap srcPixmap, float outputWidth, float outputHeight, SKImageSamplingOptions options = null)`

**Behavioral Changes**

None

**PR Checklist**

- [n] Has tests (if omitted, state reason in description) -> Added sample (should add many unit tests some day)
- [y] Rebased on top of `dev/update-skia` at time of PR
- [y] Changes adhere to coding standard
- [n] Updated documentation -> Not yet, until we agree on API, if this PR is accepted at all ;-)

### DIFF
--- a/binding/Binding.Resampler/FilterKernel.cs
+++ b/binding/Binding.Resampler/FilterKernel.cs
@@ -1,0 +1,144 @@
+﻿using System;
+using System.Numerics;
+
+namespace SkiaSharp
+{
+	internal sealed class FilterKernel
+	{
+		private readonly int stride;
+		private readonly float[] weights;
+		private readonly int[] offsets;
+		private readonly int[] counts;
+
+		public FilterKernel (SKSamplingFilterFunction filterFunction, int filterRadius, int srcLength, int dstLength, double outputShift, double outputLength, bool superSampling, bool wrapAtEdges)
+		{
+			double srcScaleFactor = outputLength / srcLength;
+			double dstScaleFactor = srcLength / outputLength;
+			double srcKernelRadius = filterRadius * dstScaleFactor;
+
+			// https://en.wikipedia.org/wiki/Lanczos_resampling
+			stride = 2 + (int)Math.Ceiling (2 * srcKernelRadius);
+			weights = new float[dstLength * stride];
+			offsets = new int[dstLength];
+			counts = new int[dstLength];
+
+			// To detect indices out of range, and for better cache locality, we allocate this on the stack
+			Span<float> localWeights = stackalloc float[stride];
+
+			int dstMax = dstLength - 1;
+
+			float startCoverage = (float)(1 - outputShift);
+			float endCoverage = (float)(1 - (dstLength - outputShift - outputLength));
+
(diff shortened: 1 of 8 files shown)


mattleibow:
Could you rebase on the new and shiny "develop" branch. That is where the v2 action is going. Or is this for v1 as well and then maybe base on master?

Ziriax:
Will do! No I need the 32-bit floating point pixel format, and I don't think that is already in master?

Ziriax:
Going to make another PR for this, cherry-picking my work.

PR #1276, opened 2020-05-07
Bindings for m80 Vulkan backend context.

MarchingCube:
**Description of Change**

Native changes: https://github.com/mono/skia/pull/78

This updates old m68 bindings to m80 which had a few changes (mostly in native).

- [ ] Updated documentation


binding/Binding/GRVkBackendContext.cs
```diff
@@ -1,37 +1,87 @@
...
+					getProcProxy = DelegateProxies.Create (
+						value,
+						DelegateProxies.GRVkGetProcDelegateProxy,
+						out var gch, out var ctx);
+					getProcHandle = gch;
+					getProcContext = ctx;
+					getProc = value;
+				}
+			}
+		}
```
mattleibow, line 57: I like this!

binding/Binding/GRVkBackendContext.cs
```diff
@@ -1,37 +1,87 @@
...
+
+		public UInt32 MaxAPIVersion { get; set; }
+
+		public GRVkExtensions Extensions { get; set; }
+
+		public IntPtr VkDeviceFeatures { get; set; }
+
+		public IntPtr VkDeviceFeatures2 { get; set; }
+
+		public GRVkGetProcDelegate GetProc {
```
mattleibow, line 35: The only thing I have a question about is what we should call this property... `GetProc` is fine, but is there a .NET-ism for this? I'm going to have to browse the reference sources...

MarchingCube: Well, this is how this is called in Skia. I would rather not diverge too much, especially in such niche API.

PR #1277, opened 2020-05-08
Performance Improvements

mattleibow:
**Description of Change**

Some work into improving performance of object creation and disposal. On average as simple create-dispose operation is 2.84x faster:

```cs
new SKPaint().Dispose()
```

This is almost 3x faster, and just using a single thread, one object at a time. The perf improvements will come for a few reasons:
 - some objects do not need to be added/located in a dictionary
 - reduced dictionary operations improves perf for other objects that are in the dictionary
 - the commonly used objects are not added

**Bugs Fixed**

None.

**API Changes**

None.

**Behavioral Changes**

Some objects no longer are stored in the the global dictionary. This is not a user-visible change.

**PR Checklist**

- [x] Updated documentation


binding/Binding/GRContext.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace SkiaSharp
 {
-	public unsafe class GRContext : SKObject, ISKReferenceCounted
+	public unsafe class GRContext : SKObject, ISKReferenceCounted, ISKSkipObjectRegistration
```
mattleibow, line 6: To mark an object as "do not add to dictionary", add the `ISKSkipObjectRegistration` interface.
 - `GetInstance` will always return `null`
 - `GetOrAddObject` will always return a new object
 - `RegisterHandle` is a no-op
 - `DeregisterHandle` is a no-op

binding/Binding/GRContext.cs
```diff
@@ -105,6 +105,6 @@ public void GetResourceCacheUsage (out int maxResources, out long maxResourceByt
 		public int GetRecommendedSampleCount (GRPixelConfig config, float dpi) => 0;
 
 		internal static GRContext GetObject (IntPtr handle) =>
-			GetOrAddObject (handle, (h, o) => new GRContext (h, o));
+			new GRContext (handle, true);
```
mattleibow: Because we _know_ the object will always be new and is never in the dictionary, we can further improve perf by not even using a "object factory". Just use the constructor.

binding/Binding/HandleDictionary.cs
```diff
@@ -27,6 +30,11 @@ internal static bool GetInstance<TSkiaObject> (IntPtr handle, out TSkiaObject in
 				return false;
 			}
 
+			if (SkipObjectRegistrationType.IsAssignableFrom (typeof (TSkiaObject))) {
+				instance = null;
+				return false;
+			}
```
mattleibow, line 36: This should not be hit as this is more used by unit tests.

binding/Binding/HandleDictionary.cs
```diff
@@ -45,6 +53,16 @@ internal static TSkiaObject GetOrAddObject<TSkiaObject> (IntPtr handle, bool own
...
 
+			if (SkipObjectRegistrationType.IsAssignableFrom (typeof (TSkiaObject))) {
+#if THROW_OBJECT_EXCEPTIONS
+				throw new InvalidOperationException (
+					$"For some reason, the object was constructed using a factory function instead of the constructor. " +
+					$"H: {handle.ToString ("x")} Type: {typeof (TSkiaObject)}");
+#else
+				return objectFactory.Invoke (handle, owns);
+#endif
+			}
```
mattleibow, line 64: Technically "slow" but should never be hit because the constructors are being used directly. But, just to be safe we include the code here. However, debug runs will actually throw as well.

binding/Binding/HandleDictionary.cs
```diff
@@ -111,6 +129,9 @@ internal static void RegisterHandle (IntPtr handle, SKObject instance)
 			if (handle == IntPtr.Zero || instance == null)
 				return;
 
+			if (instance is ISKSkipObjectRegistration)
+				return;
```
mattleibow, line 133: The no-op register.

source/SkiaSharp.Build.targets
```diff
@@ -48,7 +48,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Reference Include="netstandard" />
```
mattleibow, line 51: Too many warnings!

source/SkiaSharp.Build.targets
```diff
@@ -60,7 +59,12 @@
     Generate the assembly versioning attributes from the VERSIONS.txt file.
     ===================================================================================================================
   -->
-  <Target Name="_GenerateAssemblyVersionInfo" BeforeTargets="CoreCompile" Condition=" '$(SkipGenerateAssemblyVersionInfo)' != 'true' ">
+  <PropertyGroup Condition=" '$(SkipGenerateAssemblyVersionInfo)' != 'true' ">
+    <_GenerateAssemblyVersionInfoBeforeTargets>
+      CoreCompile
+    </_GenerateAssemblyVersionInfoBeforeTargets>
+  </PropertyGroup>
+  <Target Name="_GenerateAssemblyVersionInfo" BeforeTargets="$(_GenerateAssemblyVersionInfoBeforeTargets)" Condition=" '$(SkipGenerateAssemblyVersionInfo)' != 'true' ">
```
mattleibow, line 67: To avoid issues when setting up tasks that may not run, use a property group. This will allow for "skipped" targets to never reach the queue.

Gillibald:
This is doing what I wanted. Avoiding the redundant overhead. Some managed objects should be lightweight enough to always create a new instance instead of caching.

Ziriax:
This might be a silly question, but would be a possible to use a `struct` wrapper from some objects? E.g. would it make sense to make `SkPaint` etc an immutable struct, with operator overloading so that two structs are equal when their internal native pointer equals?

mattleibow:
@Ziriax could you maybe start a discussion on this and why it will help? Almost lost your comment as it was on a closed issue.

Ziriax:
@mattleibow Sorry, indeed, I shouldn't comment on closed issues. But no need to start a discussion, my question was silly. Using `struct` for wrappers will only work if the SkiaSharp user carefully manually manages ownership, calling `Dispose` as needed. Like in the "good" old days of C... It would be very fast, but not very .NET friendly; users expect garbage collection to take place. If C# had an automatic RAII pattern, like C++ scoped constructors/destructors, then it would be a nice option IMHO.

PR #1299, opened 2020-05-18
Make the SKFont's public API glyph-only and the others types use it under the hood

mattleibow:
**Description of Change**

Make `SKFont` glyph-only:

- hide the `SKFont` members that take strings/bytes
- hide the overloads that return arrays (promote the use of the "out span")
- expose string/byte members on `SKPaint`, and only expose the returning array (for consistency)
- use a single instance of `SKFont` for `SKTypeface` glyph members

The whole reason for this PR is to help define or direct type usage:

- `SKFont`  
   - This type is very low level and just works with glyphs and unicode (`ushort` and `int`).
   - To keep all the text members together, the other overloads are also defined, but are internal.
   - For now, only the overloads that take an "outwards span" are exposed: `void Get(ReadOnlySpan<T> input, Span<T> output)`
- `SKPaint`  
   - This type is for use with a canvas, so only exposes the string (and sadly, due to bad choices the byte arrays).
   - All members just redirect to the underlying `SKFont` internal members.
   - For now, only the overloads that return an array are exposed: `T[] Get(ReadOnlySpan<T> input)`
- `SKTypeface`
  - This type is used to get glyph information (in addition to the other members): count, get and contains
  - All members just redirect to an underlying `SKFont`
   - For now, only the overloads that return an array are exposed: `T[] Get(ReadOnlySpan<T> input)`

Typically, the members on `SKPaint` and `SKTypeface` should not really be used as the string logic is dumb and only really understands Latin text. However, it is useful still. The members on `SKFont` are all the things you would use when doing text shaping. And, that API strongly encourages the use of array pools or other non-allocating operations.

**Bugs Fixed**

None.

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKFont.cs
+++ b/binding/Binding/SKFont.cs
@@ -262,7 +262,7 @@ namespace SkiaSharp
 			return SkiaApi.sk_font_text_to_glyphs (Handle, text, (IntPtr)length, encoding, null, 0);
 		}
 
-		// MeasureText
+		// MeasureText (text)
 
 		internal float MeasureText (string text, SKPaint paint = null) =>
 			MeasureText (text.AsSpan (), paint);
@@ -316,6 +316,23 @@ namespace SkiaSharp
 				return 0;
 
 			return SkiaApi.sk_font_measure_text (Handle, text, (IntPtr)length, encoding, bounds, paint?.Handle ?? IntPtr.Zero);
+		}
+
+		// MeasureText (glyphs)
+
+		public float MeasureText (ReadOnlySpan<ushort> glyphs, SKPaint paint = null)
+		{
+			fixed (ushort* gp = glyphs) {
+				return MeasureText (gp, glyphs.Length * 2, SKTextEncoding.GlyphId, null, paint);
+			}
+		}
+
+		public float MeasureText (ReadOnlySpan<ushort> glyphs, out SKRect bounds, SKPaint paint = null)
+		{
+			fixed (ushort* gp = glyphs)
+			fixed (SKRect* b = &bounds) {
+				return MeasureText (gp, glyphs.Length * 2, SKTextEncoding.GlyphId, b, paint);
+			}
 		}
 
 		// BreakText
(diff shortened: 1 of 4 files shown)


binding/Binding/SKFont.cs
```diff
@@ -264,46 +264,46 @@ internal int CountGlyphs (void* text, int length, SKTextEncoding encoding)
...
-		public float MeasureText (ReadOnlySpan<byte> text, SKTextEncoding encoding, SKPaint paint = null)
+		internal float MeasureText (ReadOnlySpan<byte> text, SKTextEncoding encoding, SKPaint paint = null)
 		{
 			fixed (void* t = text) {
 				return MeasureText (t, text.Length, encoding, null, paint);
 			}
 		}
 
-		public float MeasureText (IntPtr text, int length, SKTextEncoding encoding, SKPaint paint = null) =>
+		internal float MeasureText (IntPtr text, int length, SKTextEncoding encoding, SKPaint paint = null) =>
```
Gillibald, line 284: I think there should be overloads that take a span of glyph indices. `ReadOnlySpan<ushort> glyphs`

Gillibald: So all methods that except `SKTextEncoding` will use a hardcoded `SKTextEncoding.GlyphId`

mattleibow: Very interesting... seems the C++ does not have that. But, I can add that. The interesting thing is that the current `measureText` method first gets the glyphs and measures that...

mattleibow: > So all methods that except `SKTextEncoding` will use a hardcoded `SKTextEncoding.GlyphId`

I see. Just had to check the code... Sometimes the void* is restricted to various UTF* types.

Gillibald: If it is not supported out of the box then it is fine.

mattleibow: I am going to add it as it is just a matter of passing the glyph IDs. Nice of them to support glyphs.

PR #1347, opened 2020-06-23
Add version numbers to native libraries

mattleibow:
**Description of Change**

Add a "suffix" to library names:

`libSkiaSharp_<milestone>_<skiasharp>.<ext>`

For example for 1.68.3:

`libSkiaSharp_68_3.dll`

And for 2.80.0:

`libSkiaSharp_80_0.dll`

This will avoid the cases when older versions are used with newer apps and causing crashes.

The reason for the 2 numbers only is the way SkiaSharp is versioned:

`<major-change-indicator>.<skia-version>.<skiasharp-version>.<bug-or-packaging-fix>`

So...
- Going from `1.*` to `2.*` is nothing special - except we may have re-written the library (which we haven't actually done because we like backwards compatibility)
- Going from `*.1` to `*.2` means that Google's code is new - and may have behavioral changes (which usually doesn't happen - except for bug fixes)
- Going from `*.*.1` to `*.*.2` means that we did a new API or changed some interop code (and thus ABI breaking)
- Going form `*.*.*.1` to `*.*.*.2` means that there may have been a docs/packaging/managed-only change (and the native library is still the same version)

### DIFF
--- a/VERSIONS.txt
+++ b/VERSIONS.txt
@@ -28,6 +28,9 @@ Xamarin.Forms                                   reference   4.4.0.991757
 libSkiaSharp            soname      80.0.0
 HarfBuzz                soname      0.20601.0
 
+# native suffix
+libSkiaSharp            suffix      _80_0
+
 # SkiaSharp.dll
 SkiaSharp               assembly    2.80.0.0
 SkiaSharp               file        2.80.0.0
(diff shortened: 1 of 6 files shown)


mattleibow:
Closing this in favor of adding a version API to the actual library in #1342

PR #1352, opened 2020-06-24
Update the version of Clang used to build for Tizen

mattleibow:
**Description of Change**

Clang v4.0 is too old to use, so update to a newer version.


mattleibow:
This PR was reverted and now in #1386

PR #1333, opened 2020-06-13
Add WASM Support for Uno Platform

mattleibow:
**Description of Change**

The title says it all...

Initial files that are produced by the build:
 - [NuGets zip](https://github.com/mono/SkiaSharp/files/4779378/NuGets.zip)
 - [libSkiaSharp.a zip](https://github.com/mono/SkiaSharp/files/4782725/libSkiaSharp.zip)

**Bugs Fixed**

- Related to #1219
- Related to #1194
- Waiting on https://github.com/dotnet/aspnetcore/issues/5466 for Blazor

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of master at time of PR
- [ ] Changes adhere to coding standard
- [ ] Updated documentation

Issue #1219, opened 2020-04-08
[FEATURE] WebAssembly Support


mattleibow:
The is the issue to track the main WebAssembly support. Several issues have already been opened on this topic in various parts:
 - [FEATURE] Support webassembly using CanvasKit (https://github.com/mono/SkiaSharp/issues/750)
 - [QUESTION] SkiaSharp wasm support (https://github.com/mono/SkiaSharp/issues/1132)
 - [FEATURE] Support SkiaSharp as a Blazor Extension (https://github.com/mono/SkiaSharp/issues/1194)
 - [FEATURE] Add a SkiaApi and SKAbstractManagedStream native abstraction (https://github.com/mono/SkiaSharp/issues/876)

RChrisCoble:
**Regarding:**

_[FEATURE] Support SkiaSharp as a Blazor Extension (#1194)
This is what we are going to be doing here._

When it's a **Blazor Extension** you can utilize the SkiaSharp API on the Client or Server Blazor deployment model.   I believe what you are doing here is akin to the forked Uno repo, allowing Skia/SkiaSharp to work properly when running under WASM in a client browser.  My specific feature request around making it a Blazor Extension was to also allow the API to run on the server with the rendering happening on the client.

This would most likely mean Skia (and maybe SkiaSharp) would need to be running in the Client Browser, with the SkiaSharp API calls being packaged up and marshaled over the Blazor SignalR connection when the Blazor Extension was running on the server.

The example of this was done with the HTML5 Canvas here:  [https://github.com/BlazorExtensions/Canvas](https://github.com/BlazorExtensions/Canvas)

If you look at that GitHub example, they wrote a wrapper around the HTML5 canvas API's which ended up doing one of the following:  (1) If the Blazor Extension was running in a client browser, issued the draw call directly or (2) If the Blazor Extension was running on the server, marshaled the call over the SignalR connection to the client to be executed.  

The Blazor server deployment model is valuable in situations where you can't deploy multiple MB's of DLL's down to the client just to draw anything.

hawkerm:
Any updates on this issue? I found this feed https://aka.ms/skiasharp-eap/index.json from another issue, assume that has the same bits called out above?

Is there any sample project in the `develop` branch that shows how to use the WASM bits yet?

I'm interested in testing out SkiaSharp for a Blazor WebAssembly Client App.

Thanks!

michaldobrodenka:
[...]

Issue #1194, opened 2020-03-27
[FEATURE] Support SkiaSharp as a Blazor Extension


RChrisCoble:
My understanding is SkiaSharp running on WebAssembly is in progress with a target release of May.  This would allow a "Blazor Client" application to use SkiaSharp and Skia fully running in a supported client browser.

One highly desired feature after that would be supporting the Blazor model of having the option to run the business logic calling SkiaSharp either client or server via a "Blazor Extension".  Running from the server is important for scenarios of smaller download size, thin client support, taking advantage of server capabilities, etc.  I understand the performance would be quite different in the server scenario, but it would still 'work'.

An example of this concept used on an HTML5 Canvas API is here:
[https://github.com/BlazorExtensions/Canvas](https://github.com/BlazorExtensions/Canvas)

This isn't rocket science, as the HTML5 canvas is a known API existing in the client browser, and the Blazor Extension wraps the canvas API so when running on the client, the extension simply calls the canvas API directly.  When running on the server, the Extension marshals the API call to the client over the Blazor SignalR connection.  There's some extra magic in that linked Extension to batch calls for performance, but you get the idea.

While I'm not an expert on the SkiaSharp architecture, my assumption is the only way this would work is have Skia still run in the client browser, and SkiaSharp (or a thin wrapper) run on the server to marshal the calls over SignalR.  Alternatively (but probably impossible for **this repo**) is have Skia marshall the calls over SignalR and have them execute using WebGL or Canvas.

So, ultimately the feature request is once SkiaSharp can run in Client Blazor, please support SkiaSharp as a Blazor Extension.

mattleibow:
Moving this discussion to a single location: https://github.com/mono/SkiaSharp/issues/1219

mattleibow:
https://github.com/mono/SkiaSharp/issues/1219#issuecomment-611157056

> **Regarding:**
> 
> _[FEATURE] Support SkiaSharp as a Blazor Extension (#1194)
> This is what we are going to be doing here._
> 
> When it's a **Blazor Extension** you can utilize the SkiaSharp API on the Client or Server Blazor deployment model.   I believe what you are doing here is akin to the forked Uno repo, allowing Skia/SkiaSharp to work properly when running under WASM in a client browser.  My specific feature request around making it a Blazor Extension was to also allow the API to run on the server with the rendering happening on the client.
> 
[...]

PR #1359, opened 2020-06-27
Build libSkiaSharp for WASM (only)

mattleibow:
**Description of Change**

In order to help the world get started with SkiaSharp on WASM, this PR will do the absolute minimum to get started. No views, no interop, no managed code at all. It just builds libSkiaSharp for WASM so that no one else has to deal with anything other than using it.

This artifact will be attached to the releases like the initial Linux builds were and can be pulled down and used. It will work with the code for the managed release. It is highly likely that there will be another intermediate PR that fixes the generate code (see #1333) so that it can actually work without modification, but small steps. 

> There is a tiny example that was thrown together, making use of @UnoPlatform's bootstrapper. Special thanks to them for all their work in this regard.

**Bugs Fixed**

- Main WASM PR is #1333 
- Relates to #1219 and #1194


mattleibow:
cc @jeromelaban

jeromelaban:
Very nice ;)

PR #1361, opened 2020-06-28
WASM Unit Tests

mattleibow:
**Description of Change**

Since we have the WASM library, and we do have the capability of running a headless browser on CI... Why not?

This PR does a few things ('cause, bugs) but the main idea is to set up a test platform. Right now, it just p/invokes into the library and checks the version, but that is just a start. In a separate PR, the full test suite can be added. (the main reason for not doing this now is that we need a few things from #1333 that aren't quite ready yet)

The next chunk of work will be to get a NuGet package that can be pulled in by any platform - such as Uno - and then be hooked up with whatever system/scripts/tools that are being used. For example, if Uno.SkiaSharp depended on `SkiaSharp.NativeAssets.Wasm`, all they will need is a custom build of the managed part that works in the way they need. When all the bits are aligned, we can just go ahead and release a `SkiaSharp.Wasm` - or even merge this if .NET 6 supports it.

So far, we have a WASM library building, and this PR spot checks it to make sure it is valid and can be used. The next PR will do full tests and/or NuGets and/or libraries.


mattleibow:
Moving this to v2.80 so we can confirm that the WASM library is actually working when we release :)

Issue #1048, opened 2019-12-03
[BUG] SKTouchAction.Cancelled not issued on iOS when SkCanvasView is in a ScrollView


MaurizioPanzica:
**Description**

[On iOS]
When a SkCanvasView is in a ScrollView, and the EnableTouchEvents property is set to 'true', the touch event is not cancelled (i.e.: SkTouchAction.Cancelled is not invoked) if the ScrollView is scrolled while touching the SkCanvasView. Neither is the SkTouchAction.Released event.
Everything works fine on Android.

**Code**

See attached sample project.

**Expected Behavior**

SKTouchAction.Cancelled event invoked when the container is scrolled while touching the SkCanvasView.

**Actual Behavior**

Event is not invoked on iOS. Works fine on Android. Not tested on other platforms.

**Basic Information**

- Version with issue:  1.68.1
- Last known good version:  unknown
- IDE:  VS2019 Mac/Windows
- Platform Target Frameworks: 
  - iOS:  

**Reproduction Link**

[TestSkiaSharp.zip](https://github.com/mono/SkiaSharp/files/3916183/TestSkiaSharp.zip)

MaurizioPanzica:
Well, after further investigations this behavior seems to be related to iOS and not to SkiaSharp itself.
In this particular scenario, the underlying scrollview, as soon as the scroll begins, invokes the method 'touchesCancelled' of the view and not the GestureRecognizer's.
This is why the SKCanvasView does not receive it. But the native view does. I guess something can be made at that level.

daltonks:
Hey @MaurizioPanzica,
I am currently running into this problem. I need a ListView that contains SKCanvasViews with touch handling.
Have you been able to find a workaround?

mpanzica:
@daltonks Unfortunately not.
Since then the project has been put on hold because of Covid and I've been working on another one. I guess it will be resumed in the near future. Rest assured that I will post the workaround, should I find one.

Cheers,
M.

daltonks:
@mpanzica Thanks for the reply :)
I'm kind of led to believe that Xamarin is mucking up the touches somehow, because as far as I can tell, TouchesCancelled should be propagated to gesture recognizers when a scroll view sends the event.

mpanzica:
[...]

PR #1367, opened 2020-06-29
Fire Cancelled events from SKTouchHandler.Reset() when needed

daltonks:
**Description of Change**

Touches are now tracked in `SKTouchHandler`. For all touches that aren't Ended or Cancelled by the time `Reset()` is called, the `Reset()` method fires `SKTouchAction.Cancelled` events. This fixes `SKTouchAction.Cancelled` not getting sent when the user scrolls.

https://developer.apple.com/documentation/uikit/uigesturerecognizer/1620004-reset

There could potentially be a better way of doing this without a `HashSet`, but this is the only method that I have found to work just like the other platforms.

**Bugs Fixed**

- #1048

**API Changes**

None

**Behavioral Changes**

When the iOS gesture recognizer is Reset (commonly due to the user scrolling a view), the `SKTouchAction.Cancelled` touch event will now be received.

**PR Checklist**

- [Let me know if there are tests to add] Has tests (if omitted, state reason in description)
- [✔] Rebased on top of master at time of PR
- [✔?] Changes adhere to coding standard
- [N/A I hope] Updated documentation


daltonks:
Hey @mattleibow, did you have any comments on this PR?

mattleibow:
Thanks for this PR and sorry for the very long wait. I'll think about that hash set since it is not the greatest to store things - unless there is a guarantee that all the touches will actually be released.

However, I'll do some testing and get this into the next release.

PR #1370, opened 2020-06-30
Handle a crash when the Apple views are 0x0 or 1x1

mattleibow:
**Description of Change**

Sometimes in Xamarin.Forms, the view is added to the hierarchy at a 1x1 or invisible. This results in the view trying to draw, but actually not a valid GL context yet.

Detect this when the GRContext cannot be created and wait for the next render. We are not losing any data, because the view is hidden or 1x1.


mattleibow:
This was not the issue. The iOS GL interface assembler was incorrect.

PR #1384, opened 2020-07-06
Do not use any locking

### LINKED ISSUE #1383: [BUG] SkiaSharp can call WM_PAINT handler and random COM callbacks when creating or obtaning objects
```
   Avalonia.Win32.dll!Avalonia.Win32.WindowImpl.WndProc(System.IntPtr hWnd, uint msg, System.IntPtr wParam, System.IntPtr lParam) Line 30  C#
   [Native to Managed Transition]  
   [Managed to Native Transition]  
   System.Private.CoreLib.dll!System.Threading.WaitHandle.WaitOneNoCheck(int millisecondsTimeout)  Unknown
   System.Private.CoreLib.dll!System.Threading.WaitHandle.WaitOne(int millisecondsTimeout)  Unknown
   System.Private.CoreLib.dll!System.Threading.ReaderWriterLockSlim.WaitOnEvent(System.Threading.EventWaitHandle waitEvent, ref uint numWaiters, System.Threading.ReaderWriterLockSlim.TimeoutTracker timeout, System.Threading.ReaderWriterLockSlim.EnterLockType enterLockType)  Unknown
   System.Private.CoreLib.dll!System.Threading.ReaderWriterLockSlim.TryEnterWriteLockCore(System.Threading.ReaderWriterLockSlim.TimeoutTracker timeout)  Unknown
   SkiaSharp.dll!SkiaSharp.HandleDictionary.RegisterHandle(System.IntPtr handle, SkiaSharp.SKObject instance)  Unknown
   SkiaSharp.dll!SkiaSharp.SKObject.RegisterHandle(System.IntPtr handle, SkiaSharp.SKObject instance)  Unknown
   SkiaSharp.dll!SkiaSharp.SKObject.Handle.set(System.IntPtr value)  Unknown
   SkiaSharp.dll!SkiaSharp.SKObject.SKObject(System.IntPtr handle, bool owns)  Unknown
   SkiaSharp.dll!SkiaSharp.SKPath.SKPath(System.IntPtr handle, bool owns)  Unknown
   SkiaSharp.dll!SkiaSharp.SKPath.SKPath()  Unknown
```

This is caused by all locks in .NET being alertable locks that [can still run the message pump when waiting for a lock](https://stackoverflow.com/a/4540745/2231814).

This behavior might lead to unexpected behavior in the user code or even deadlocks, since user code doesn't expect to receive WM_PAINT while calling SKPath constructor.

mattleibow:
**Description of Change**

It appears that using locks triggers a paint...

So, we can't do that. I have pulled in the code from https://github.com/VSadov/NonBlocking to test out. This will actually be collapsed into the minimum requirements for an IntPtr dictionary.

**Bugs Fixed**

- Fixes #1383

None.

**Behavioral Changes**

None.

**PR Checklist**

- [x] Updated documentation


binding/Binding/HandleDictionary.cs
```diff
@@ -170,45 +162,31 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
-			instancesLock.EnterWriteLock ();
-			try {
-				var existed = instances.TryGetValue (handle, out var weak);
-				if (existed && (!weak.IsAlive || weak.Target == instance)) {
-					instances.Remove (handle);
-				} else {
+			var existed = instances.TryUpdate (handle, emptyRef, instance.WeakHandle);
+			if (existed) {
+				// TODO: remove from the dictionary
+				//       or, better yet, do this in a single op
```
mattleibow, line 168: This is the reason I originally switched away from the dictionary. There is no easy way to have an atomic comparison between the key and value. I only want to remove if the key AND value match. 

The reason for this is just because the native side will have created a brand new object at the same location, and we mustn't remove it form the dictionary. We only do that if this is our handle AND our object.

binding/Binding.Shared/NonBlocking/ConcurrentDictionary/DictionaryImpl`3.cs
```diff
@@ -0,0 +1,1464 @@
...
+                entry = ref curTable._entries[idx];
+            }
+
+            // Found the proper Key slot, now update the Value.  
+            // We never put a null, so Value slots monotonically move from null to
+            // not-null (deleted Values use Tombstone).
+
+            // volatile read to make sure we read the element before we read the _newTable
+            // that would guarantee that as long as _newTable == null, entryValue cannot be forwarded.
+            var entryValue = Volatile.Read(ref entry.value);
```
mattleibow, line 346: We could check the value here, and then jump out if it didn't match.

mattleibow:
From @kekekeks in [gitter](https://xamarinhq.slack.com/archives/C9A0TTQFQ/p1594121735187500):

> Why are you using AddOrUpdate?
> I think that TryAdd is better
> since the thread that has lost the race will just destroy its newly created object and take one from the dictionary (edited) 
> it's way less bug prone to have a guarantee that a particular handle is only getting added to the table once

mattleibow:
from @kekekeks 
> @mattleibow I think we could use an alternative approach
> 1) implement some lock abstraction that would use EnterCriticalSection on win32 and Monitor.Enter otherwise
> 2) copy-paste ConcurrentDictionarty from dotnet/runtime and make it use said lock abstraction
> might be costly due to additional P/Invoke though

mattleibow:
This is probably not needed after #1817

Either way, this is not the right approach.

Issue #1403, opened 2020-07-14
[FEATURE] Allow building just a single external


Ziriax:
For testing, I often build just a single external, e.g. Windows.

Instead of listing all externals that should be skipped, it would be handy to specify just the externals to build.

E.g. in the `externals.cake`:

```CS
    if (JUST_EXTERNALS.Length > 0)
        return JUST_EXTERNALS.Contains(platform);
```

On could then build with 

`./bootstrapper.ps1 --target=libs --justExternals=windows`

Ziriax:
Oh btw, it seems `--skipExternals` doesn't accept multiple tokens anymore, this now gives the error:

```
 ./bootstrapper.ps1 --target=libs --skipExternals="linux,macos"
Preparing to run build script...
Running build script...
More than one build script specified.
For usage, use parameter --help
```

Changing the cake code to split by semicolon instead of comma, makes it work again.

mattleibow:
This is already there with the format `--target=externals-<native>`. Where `<native>` matches the name of the "native" folder.

For example, to build just iOS, do `externals-ios`. 

This is generated from the folder names:
https://github.com/mono/SkiaSharp/blob/master/cake/externals.cake#L7-L18

mattleibow:
But the semicolon issue might break macos. I would have it split on both `;` and `,`. Then i should work everywhere.

Ziriax:
Oh nice. Didn't think of that. Then please ignore my PR.

mattleibow:
No prob. Thanks for the issues and general seconds set of eyes.

PR #1406, opened 2020-07-14
Allow building just a single external

Ziriax:
**Description of Change**

Passing `-justExternals=windows` to the bootstrapper just builds that external.

**Bugs Fixed**

- Related to issue #1403 

This also replaces the `,` separator with `;` for `--skipExternals`, since `,` didn't work at all, at least on Windows. See note in issue #1403. Maybe I should have created another issue for that 😉 

**API Changes**

none

**Behavioral Changes**

Passing multiple externals to `--skipExternals` separated by semicolons now works.

**PR Checklist**

- [n] Has tests (if omitted, state reason in description)
- [y] Rebased on top of master at time of PR
- [y] Changes adhere to coding standard
- [n] Updated documentation

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -28,7 +28,9 @@ DirectoryPath ROOT_PATH = MakeAbsolute(Directory("."));
 #load "cake/native-shared.cake"
 
 var SKIP_EXTERNALS = Argument ("skipexternals", "")
-    .ToLower ().Split (new [] { ',' }, StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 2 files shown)


mattleibow:
This is already there with the format `--target=externals-<native>`. Where `<native>` matches the name of the "native" folder.

For example, to build just iOS, do `externals-ios`. 

This is generated from the folder names:
https://github.com/mono/SkiaSharp/blob/master/cake/externals.cake#L7-L18

mattleibow:
Closing this as you can pick a build using the targets `externals-*`

PR #1420, opened 2020-07-19
Add more Uno things

mattleibow:
**Description of Change**

More Uno things and some small tweaks


binding/Binding/DelegateProxies.cs
```diff
@@ -103,18 +103,18 @@ private static void SKSurfaceReleaseDelegateProxyImplementation (void* address,
 		}
 
 		[MonoPInvokeCallback (typeof (GRGlGetProcProxyDelegate))]
-		private static IntPtr GRGlGetProcDelegateProxyImplementation (void* context, string name)
+		private static IntPtr GRGlGetProcDelegateProxyImplementation (void* context, void* name)
 		{
 			var del = Get<GRGlGetProcedureAddressDelegate> ((IntPtr)context, out _);
-			return del.Invoke (name);
+			return del.Invoke (Marshal.PtrToStringAnsi ((IntPtr)name));
```
mattleibow, line 109: This is purely for WASM which is having issues with string on callbacks.

PR #1487, opened 2020-09-02
feat: Add support for Uno Platform 3.0

jeromelaban:
**Description of Change**

Updates to Uno Platform 3.0

**Bugs Fixed**

- Fixes the WebAssembly `SKXamlCanvas` not being cleared on creation.

**API Changes**

None.

**Behavioral Changes**

None.

**PR Checklist**

- [ ] Updated documentation


mattleibow:
Very nice! Thanks! 

What is the reason for switching from the Marshal to the managed array? I have no problem just wondering the reason.

jeromelaban:
> Very nice! Thanks!
> 
> What is the reason for switching from the Marshal to the managed array? I have no problem just wondering the reason.

This is for two reasons. The first is because the array needs to be memset to zero, and the second is looping in managed code can be quite slow with the interpreter, even when using pointers. Pinning the managed array and providing it to Skia is very fast.

Also the build is failing, but I cannot see the errors. What are they ?

mattleibow:
> Also the build is failing, but I cannot see the errors. What are they ?

Oh, nothing special... just the bot losing connection to the internet 😆 Nothing weird for a cloud machine. One might think these are real physical servers 🤔

mattleibow:
@jeromelaban thanks for this PR!

I have merged it into a local branch so that CI gets more power 😄 
Things are looking good and it should be all good to merge into master when CI finishes.

PR #1489, opened 2020-09-03
Update Uno Platform to v3.0

mattleibow:
**Description of Change**

 - Pull in the changes from Uno
 - Create a local branch so more things run
 - Add the new projects to all the solutions

**Bugs Fixed**

- Merging #1487

**API Changes**

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Reference/nuget/build/netstandard2.0/SkiaSharp.Views.Uno.targets
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project>
+
+  <ItemGroup>
+    <UnoRuntimeEnabledPackage Include="SkiaSharp.Views.Uno" PackageBasePath="$(MSBuildThisFileDirectory)" />
```
mattleibow, line 5: @jeromelaban How does the `PackageBasePath` work. If this is included into the nuget package under the `build/netstandard` folder, the value will be that folder. Is this correct?

source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Wasm/SKXamlCanvas.Wasm.cs
```diff
@@ -1,18 +1,20 @@
...
+using Uno.UI.Runtime.WebAssembly;
 using Windows.UI.Xaml;
 
 namespace SkiaSharp.Views.UWP
 {
+	[HtmlElement("canvas")]
 	public partial class SKXamlCanvas : FrameworkElement
 	{
-		private IntPtr pixels;
+		private byte[] pixels;
```
mattleibow, line 12: Bringing the comments forward from the previous PR:

@mattleibow: What is the reason for switching from the Marshal to the managed array? I have no problem just wondering the reason.

@jeromelaban: This is for two reasons. The first is because the array needs to be memset to zero, and the second is looping in managed code can be quite slow with the interpreter, even when using pointers. Pinning the managed array and providing it to Skia is very fast.

source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Reference/nuget/build/netstandard2.0/SkiaSharp.Views.Uno.targets
jeromelaban: It's [used here](https://github.com/unoplatform/uno/blob/d08c0d6891331edb97e0d40bb4d11a0a2a9734eb/build/uno.winui.targets#L84), I still have to document it. That's correct, there's a missing `../`/
```suggestion
    <UnoRuntimeEnabledPackage Include="SkiaSharp.Views.Uno" PackageBasePath="$(MSBuildThisFileDirectory)../" />
```

I could not find a way to determine the location of the msbuild file declaring the item, so this is pretty much required for now :)

jeromelaban:
@mattleibow what's failing in the CI, can I help ?

jeromelaban:
@mattleibow sweet ;) Thanks for the troubleshooting!

samples/Basic/Uno/SkiaSharpSample.Wasm/SkiaSharpSample.Wasm.csproj
```diff
@@ -23,10 +23,10 @@
...
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.1.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Filter" Version="1.1.1" />
-    <PackageReference Include="Uno.UI" Version="2.4.4" />
-    <PackageReference Include="Uno.UI.RemoteControl" Version="2.4.4" Condition="'$(Configuration)'=='Debug'" />
-    <PackageReference Include="Uno.Wasm.Bootstrap" Version="1.4.0-dev.12" />
-    <PackageReference Include="Uno.Wasm.Bootstrap.DevServer" Version="1.4.0-dev.12" />
+    <PackageReference Include="Uno.UI" Version="3.0.12" />
+    <PackageReference Include="Uno.UI.RemoteControl" Version="3.0.12" Condition="'$(Configuration)'=='Debug'" />
+    <PackageReference Include="Uno.Wasm.Bootstrap" Version="1.4.0-dev.32" />
+    <PackageReference Include="Uno.Wasm.Bootstrap.DevServer" Version="1.4.0-dev.32" />
```
mattleibow, line 29: What is the correct version of these packages? I notice that this tweaks the version of mono and emscripten.

scripts/azure-pipelines.yml
```diff
@@ -33,7 +33,7 @@ variables:
   THROW_ON_TEST_FAILURE: true
   NUGET_DIFF_PRERELEASE: false
   ENABLE_CODE_COVERAGE: true
-  EMSCRIPTEN_VERSION: 1.39.18
+  EMSCRIPTEN_VERSION: 1.40.0
```
mattleibow, line 36: What version of emscripten should and can be used? I know there is no guarantee of compatibility between versions... Does this change with 1.40.x?

mattleibow (commented):
@jeromelaban I think all is good, I just want to confirm all the version numbers...

samples/Basic/Uno/SkiaSharpSample.Wasm/SkiaSharpSample.Wasm.csproj
jeromelaban: Yes, 3.0.12 is the correct version of the package, it does not need to be changed, same for the 1.4.0-dev.32 of the bootstrapper.

scripts/azure-pipelines.yml
jeromelaban: At this point, it has to be synchronized with the emscripten version specified here: https://github.com/unoplatform/Uno.Wasm.Bootstrap/blob/3154b8ec7396c136c1b919922678c01b1e53cf92/src/Uno.Wasm.Bootstrap/Constants.cs#L20 That being said, the bootstrapper's latest build automatically uses its own copy of an embedded emscripten, so steps for installing emscripten should now be safely removed.

It seems that mono version used by 1.4-dev is stable enough, so we'll probably make a stable out of it. Next bits of mono will use emscripten 2.0.x.

Could be that at some point we'll need to make multiple versions of the binaries with different versions of emscripten...

jeromelaban:
For python, I changed it this way: https://github.com/unoplatform/docker/blob/394e55f5b3571466c2377e6ec50bb13a4f77bec8/wasm-build/Dockerfile#L7

It changes the default python to 3.x

mattleibow:
Very interesting... I didn't know about that.

But, I need v2 for gn. So this seems to work for now.

PR #1510, opened 2020-09-26
Don't allocate arrays when copying streams

### LINKED ISSUE #1509: [FEATURE] Reduce heap allocation on SKManagedStream.OnReadManagedStream
I've found a big heap allocation on images decode at `SKManagedStream.OnReadManagedStream`. 
Screenshot 2020-09-25 at 23 50 48
Maybe it would be better to use a 81920 bytes length buffer to copy a managed stream data to unmanaged pointer? I hope Marshal.Copy cost is not huge.
Screenshot 2020-09-25 at 23 54 03

mattleibow:
**Description of Change**

Don't allocate arrays when copying streams

**Bugs Fixed**

- Fixes #1509

**API Changes**

None.

**Behavioral Changes**

Reading from a wrapped managed stream will no longer allocate huge arrays. or any arrays. Using a native copy with array pools.

**PR Checklist**

- [ ] Updated documentation


mattleibow:
It is a pity that we have to copy twice, first to the managed array and then into the destination pointer.

binding/Binding/SKManagedStream.cs
```diff
@@ -90,18 +87,25 @@ protected override void DisposeManaged ()
...
+			if (buffer == IntPtr.Zero)
+				throw new ArgumentNullException (nameof (buffer));
+			if ((int)size < 0)
+				throw new ArgumentOutOfRangeException (nameof (size));
+
+			if (size == IntPtr.Zero)
+				return IntPtr.Zero;
+
+			using var managedBuffer = Utils.RentArray<byte> ((int)size);
+			var len = stream.Read (managedBuffer.Array, 0, managedBuffer.Length);
```
Ziriax, line 99: Shouldn't this be `var len = stream.Read (managedBuffer.Array, 0, size)`;?

The `managedBuffer.Length` is always the first power of two larger than `size` I believe

mattleibow: That would be the case if I was using `ArrayPool` directly. I am using my Utils version that makes sure to set the `Length` of the "wrapper" to what I requested.

mattleibow: Thanks for the review!

Ziriax: Oh right! I actually remember you making this utils method, I forgot about that.

PR #1524, opened 2020-10-08
Add all the tests to WASM runner

mattleibow:
**Description of Change**

Add all the tests to WASM runner, skipping the ones that can't build.

**Bugs Fixed**

- Related to issue #

**API Changes**

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


scripts/azure-templates-bootstrapper.yml
```diff
@@ -15,7 +15,7 @@ parameters:
   condition: succeeded()                        # whether or not to run this template
   shouldPublish: true                           # whether or not to publish the artifacts
   configuration: $(CONFIGURATION)               # the build configuration
-  buildExternals: ''                            # the build number to download externals from
+  buildExternals: '4127355'                            # the build number to download externals from
```
sampletext32, line 18: this looks like a broken indent

tests/SkiaSharp.Wasm.Tests.sln
```diff
@@ -1,19 +1,43 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
-VisualStudioVersion = 16.0.30212.25
+VisualStudioVersion = 16.0.29102.215
```
sampletext32, line 4: I might be wrong, but this looks like user-specific change

tests/SkiaSharp.Wasm.Tests/Program.cs
```diff
@@ -1,11 +1,60 @@
...
+					var dir = Path.GetDirectoryName(dest);
+
+					if (!Directory.Exists(dir))
+						Directory.CreateDirectory(dir);
+
+					using (var stream = assembly.GetManifestResourceStream(resource))
+					using (var file = File.Create(dest))
+						stream.CopyTo(file);
+
+					Console.WriteLine($"Saved file to {dest}: {File.ReadAllBytes(dest)?.Length}");
```
sampletext32, line 39: `File.ReadAllBytes` never returns null according to documentation, it throws an exception

sampletext32: And this is a total overhead - reading a file to just get it's length
```suggestion
					Console.WriteLine($"Saved file to {dest}: {new FileInfo(dest).Length}");
```

tests/SkiaSharp.Wasm.Tests/Program.cs
```diff
@@ -1,11 +1,60 @@
...
+
+			// copy resources out of assembly
+			var prefix = "SkiaSharp.Tests.";
+			var resources = assembly.GetManifestResourceNames();
+			foreach (var resource in resources)
+			{
+				if (!resource.StartsWith(prefix))
+					continue;
+
+				if (resource.StartsWith(prefix + "fonts.") || resource.StartsWith(prefix + "images."))
```
sampletext32, line 25: I might be wrong, but I think assemblies are in CamelCase notation, so this should be 
```suggestion
				if (resource.StartsWith(prefix + "Fonts.") || resource.StartsWith(prefix + "Images."))
```

scripts/azure-templates-bootstrapper.yml
mattleibow: This is a temporary hack to speed up the CI builds. Usually it takes over an hour to build all the native bits, but for this PR, we can just download the last build's artifacts.

PR #1575, opened 2021-01-05
add macos arm64 support

danwalmsley:
**Description of Change**
Adds arm64 as a target for libSkiaSharp and  Harfbuzz... may not work..

**Bugs Fixed**

- Related to issue #

**API Changes**

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation


danwalmsley:
@mattleibow trying to get the dylibs to be universal binaries. Iv simply added architectures... dont know if more needs to be done... on Avalonia we simply made it use xcode 11.1 and by default it spat out a compatible binary.

Avalonia is almost running natively on M1 make... just need libSkiaSharp and maybe harfbuzz and it might work.

native/macos/build.cake
```diff
@@ -47,7 +47,7 @@ Task("libHarfBuzzSharp")
     .WithCriteria(IsRunningOnMac())
     .Does(() =>
 {
-    Build("x86_64");
+    Build("x86_64", "arm64");
 
     RunLipo(OUTPUT_PATH, "libHarfBuzzSharp.dylib", new [] {
        (FilePath) "x86_64/libHarfBuzzSharp.dylib"
```
Gillibald, line 53: This should be run for the arm64 arch as well if I am not mistaken

mattleibow: That is correct, for both libs.

mattleibow:
Hopefully this works, I'm also trying to get another fairly major update to skia to make sure we get all the bits that has been added for this new platform.

mattleibow:
Thanks for this PR. I opened a new PR with what I think is the final bits needed for arm64: https://github.com/mono/SkiaSharp/pull/1627

I have some test binaries and hopefully someone can test it if they have an M1 device and a build of arm64 mono/dotnet. There are some instructions that will help you get started.

Issue #1308, opened 2020-05-25
[BUG] SIGABRT: Object reference not set to an instance of an object


arashcoder:
**Description**

Hi! I'm using Microcharts library, which uses SkiaSharp under the hood. Occasionally on iOS, I get null reference exception. Here's all the crash log I get:

**Code**

```
SKCGSurfaceFactory.CreateSurface (CoreGraphics.CGRect contentsBounds, System.nfloat scale, SkiaSharp.SKImageInfo& info)
SKCanvasView.Draw (CoreGraphics.CGRect rect)
(wrapper managed-to-native) 
UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate)
UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName)
Application.Main (System.String[] args)
```

**Expected Behavior**

It shouldn't crash.

**Actual Behavior**

It crashes occasionally.

**Basic Information**

- Version with issue:  1.68.0
- Last known good version:  
- IDE:  Visual Studio for Mac 
- Platform Target Frameworks: 
  - iOS:  13.4
- Target Devices:   iPhone 8 Plus, iPhone XR, iPad (5th gen, Wi-Fi)

Any ideas why this occurs? 
Thanks in advance.

mattleibow:
Does it still crash if you update to the latest v1.68.3?

arashcoder:
Thanks for the suggestion. I will try that, but since I can't reproduce that crash I can't be sure.

arashcoder:
@mattleibow I updated to v1.68.3 and I still get that crash occasionally. 
Any ideas? 
Thanks.

mattleibow:
Could you paste the full stack trace? Very weird.

arashcoder:
That's the whole stack trace that I get in AppCenter.

arashcoder:
Thanks for the explanations. Indeed it's strange.
 This is the whole log that I get from this exception. I don't think I'm doing anything special with the graphs, and it's working almost all the time. The problem is that I can't reproduce this, so I don't have a full stack trace.
Also could this line cause the crash?
`Info = info = CreateInfo((int)contentsBounds.Width, (int)contentsBounds.Height);`

mattleibow:
That is working with struts, which cannot be null. How rare is this exception? Maybe iOS just decides that it will return null when it feels like it.

[...]

PR #1784, opened 2021-08-23
Return an empty surface in case the allocation has failed

richirisu:
**Description of Change**

Return an empty surface in case the allocation has failed

**Bugs Fixed**
- Related to issue #1308

**API Changes**

None

**Behavioral Changes**

Prior to this change the application would crash with an unrecoverable exception. After the change the application does not crash anymore but the surface of the canvas view will remain blank, i.e. nothing will be drawn on the surface in case there is not sufficient memory to allocate.

**PR Checklist**

- [ ] Updated documentation


mattleibow:
Thanks for this PR.

I was playing around with the test app from the issue and think it lasts even longer if I dispose the data when it is removed from the window. It still crashes, but much later. Without the fix, it crashed after about 380 navigations on mu iPhone 6s. With the fixe and the release, it lasted to about 500. It still crashes, but this was the app. 

It started producing blank pages for a bit (basically ran out of memory to allocate the surface), and then crashed after a while - but not from our code. Seems the OS got tired of allocating view controllers. I suppose 500 is a bit extreme for an app to just keep navigating down while preserving the back stack. Some overhead might also be that Xamarin.Forms may not be disposing the actual UIViews.

At least now, we are making sure to release the data when it is removed from the view hierarchy.

I think this is better? We don't need to keep all the data in memory because we can totally re-allocate. And, the OS still has the last image in the graphics context. Our drawing is basically a second buffer. SkiaSharp draws to that, then blits it to the screen. If we dispose our buffer, the OS still has the copy.

richirisu:
Thanks for having looked into it this thoroughly.

I think releasing the pixel data when the canvas view is being removed from the view hierarchy is a good move. After all, it can be recreated. Back in the good old days of iOS 1, it was even required to free up all data that could be recreated when switching view controllers. Of course, my test application is just an extreme example, and since NSData objects share the memory with all the rest, there can be a multitude of reasons for running out of memory.

The only thing I am not really sure about is that in most cases willMoveToWindow will only be called when the view controller is being pushed and when it is being popped again, but not when another view controller is being pushed on top. So all intermediate view controllers will still keep their views alive. But I might be mistaken.

mattleibow:
> willMoveToWindow will only be called when the view controller is being pushed and when it is being popped again, but not when another view controller is being pushed on top

I tested this and it seems that when we push another VC on top, it removes the previous one from the window. I did not actually expect that as I assumed it would still live on. However, it seems as soon as a view is no longer in the view hierarchy, it is removed from the window.

Maybe this was sort of a hint but not explicitly stated:

> **newWindow**  
> The window object that will be at the root of the receiver's new _view hierarchy_. This parameter may be nil.

richirisu:
I could confirm in Xcode that you are right. Using the “Debug View Hierarchy” feature, only the top-most view controller is attached. All intermediate view controllers are detached for the time being.

I could also confirm that for all view controllers that are being detached willMoveToWindow will be called with ’null’ (which is in accordance with the specification as the view is being detached from the window). However, willMoveToSuperview will not be called with ‘null’ because the view is still attached to its superview. Only for the view controller's root view willMoveToSuperview will be called with ‘null’ because the root view is indeed detached from its parent view (which is a view belonging to the navigation controller). Note, that you can assign any custom view as the root view by overriding the loadView method of a view controller.

I think I had confused it with the search view controller. Displaying the search view controller does not cause the actual view controller underneath to be removed from the view hierarchy.

mattleibow:
Fixes #1308

PR #1804, opened 2021-09-10
Fix the what the `IgnorePixelScaling` property works

### LINKED ISSUE #1177: [FEATURE] Rework the IgnorePixelScaling property
**Is your feature request related to a problem? Please describe.**
Currently, the `IgnorePixelScaling` property works backwards. It draws the image and then scales it up.

**Describe the solution you'd like**
The property should create the large canvas, but then apply scaling so that further drawing operations are scaled automatically.

**Additional context**
Related: https://github.com/mono/SkiaSharp/issues/527

mattleibow:
**Description of Change**

Previously the `IgnorePixelScaling` allowed the drawing code to work in density-independent UI units, however it did so in a poor manner. Instead of scaling the canvas in such a way that it preserved the high density, it stretched and reduced the quality of the image.

This PR makes sure to always draw in high resolution, and then pre-scale the canvas before invoking the paint events.

**Bugs Fixed**

- Fixes #1177 
- Avoids the need for https://github.com/dotnet/Microsoft.Maui.Graphics/pull/155

**API Changes**

None.

**Behavioral Changes**

A demo of the change. 

| Before | After |
| :-: | :-: |
| image | image | 

**PR Checklist**

- [ ] Updated documentation


myroot:
Hi, this PR meet the our requirements except one thing

We hope to able to update ScalingFactor
https://github.com/mono/SkiaSharp/blob/2ad29861d5a40d3bf78c28ab0a9cb02a8f0fe437/source/SkiaSharp.Views/SkiaSharp.Views.Tizen/ScalingInfo.cs#L30
Because, in several reasons, we need to support a virtual scaling factor.
Could you make a setter on ScalingInfo.ScalingFactor?

mattleibow:
> Because, in several reasons, we need to support a virtual scaling factor.

@myroot I can do that. Do we need to have different scaling factors on different canvases? Or can it all be in one place?

For now I will add a setter to the `ScalingInfo` type. But we can easily move it to the canvas before we merge this.

source/SkiaSharp.Views/SkiaSharp.Views.Tizen/ScalingInfo.cs
```diff
@@ -23,14 +23,21 @@ public static class ScalingInfo
...
+		public static double ScalingFactor => scalingFactorOverride ?? scalingFactor.Value;
 
 		public static double FromPixel(double v) => v / ScalingFactor;
 
 		public static double ToPixel(double v) => v * ScalingFactor;
+
+		public static void SetScalingFactor(double? scalingFactor)
+		{
+			scalingFactorOverride = scalingFactor;
+		}
```
mattleibow, line 41: Adding this as a separate method so it can be a null that can allow for reverting back to the system value.

Is this useful in this global context? Should it be on the individual canvas views?

tipa:
@mattleibow I just stumbled over this and how the `IgnorePixelScaling` works.
From what I understand the new behavior is different from the document behavior, which was the old one.
The [docs](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.views.ios.skcanvasview.ignorepixelscaling) still say `When true, the canvas is resized to device independent pixels, and then stretched to fill the view`.
This isn't the case any more with v2.88.0, is it?

PR #1811, opened 2021-09-20
Support ASP.NET Blazor Components

mattleibow:
**Description of Change**

Add support for Blazor on Web Assembly.

**Bugs Fixed**

- Related to issue #1219
- Related to issue #1194

**API Changes**

**Behavioral Changes**

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of main at time of PR
- [ ] Changes adhere to coding standard
- [ ] Updated documentation

PR #1815, opened 2021-09-23
Bump Maui to RC 2 / P9

mattleibow:
**Description of Change**

Update the things.


artemiusgreat:
Does anybody have any idea why in Mac OS I'm getting this error? 
The method `OnAfterRenderAsync` complains that there is no suitable method to override. 
If I remove `override` from this method, it says that it hides existing method in this class.
This error seems to contradict itself, the method exists but can't be overridden. 
Is it a Mono issue? 

image

mattleibow:
@artemiusgreat  I moved your comment to https://github.com/mono/SkiaSharp/pull/1811#issuecomment-932817230

PR #1817, opened 2021-09-26
Replace HandleDictionary lock with a critical section under Windows

toptensoftware:
**Description of Change**

Fix for deadlock issue related to the lock implementation used by HandleDictionary under Windows. See issue #1383 for full description of issue.

* The decision of which lock implementation to use is made at runtime to save compile time build directives
* The lock is abstracted behind a `IPlatformLock` interface
* On Windows, `IPlatformLock` maps to a Win32 Critical Section via PInvoke
* On all other platforms, `IPlatformLock` maps to a `ReaderWriterLockSlim` as before

Tested on Windows 10 under a heavy load scenario where this issue was previously easily reproducible. After the fix issue couldn't be reproduced.

**Bugs Fixed**

-Related to issue: #1383 

**API Changes**

None

**Behavioral Changes**

* Besides fixing deadlock, functionality remains the same.
* No tests are included because the issue is timing and platform specific and can only be reliably reproduced under load and can depend on GC interactions.
* Not sure about coding standards, but tried to match existing style.
* No documentation updates applicable


binding/Binding/HandleDictionary.cs
```diff
@@ -212,4 +213,115 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
+	{
+		public NonAlertableWin32Lock ()
+		{
+			_cs = Marshal.AllocHGlobal (Marshal.SizeOf<CRITICAL_SECTION> ());
+			InitializeCriticalSectionEx (_cs, 4000, 0);
+		}
+
+		~NonAlertableWin32Lock ()
+		{
+			Marshal.FreeHGlobal (_cs);
```
mattleibow, line 286: Should this first delete the critical section: `DeleteCriticalSection`?

binding/Binding/HandleDictionary.cs
```diff
@@ -212,4 +213,115 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
+	 * This can lead to re-entrancy and a deadlock on the HandleDictionary lock. 
+	 * 
+	 * This fix replaces the ReaderWriteLockSlim instance on Windows with a native Win32 
+	 * CRITICAL_SECTION.
+	 */
+
+	/// <summary>
+	/// Abstracts a platform dependant lock implementation
+	/// </summary>
+	interface IPlatformLock
```
mattleibow, line 234: Would this be worth making into a public (maybe hidden) API such that if you only ever have a single threaded app that never needs to look things up, you can just skip locks altogether and don't pay the price of locking?

binding/Binding/HandleDictionary.cs
```diff
@@ -212,4 +213,115 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
+
+	/// <summary>
+	/// Windows platform lock uses Win32 CRITICAL_SECTION
+	/// </summary>
+	class NonAlertableWin32Lock : IPlatformLock
+	{
+		public NonAlertableWin32Lock ()
+		{
+			_cs = Marshal.AllocHGlobal (Marshal.SizeOf<CRITICAL_SECTION> ());
+			InitializeCriticalSectionEx (_cs, 4000, 0);
```
mattleibow, line 281: Should these calls check the result and throw using `Marshal.GetLastWin32Error`? At least we can catch an error and not break later on in a random place.

binding/Binding/HandleDictionary.cs
```diff
@@ -212,4 +213,115 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
+			public IntPtr DebugInfo;
+			public int LockCount;
+			public int RecursionCount;
+			public IntPtr OwningThread;
+			public IntPtr LockSemaphore;
+			public UIntPtr SpinCount;
+		}
+
+		[DllImport ("Kernel32.dll")]
+		static extern bool InitializeCriticalSectionEx (IntPtr lpCriticalSection, uint dwSpinCount, uint Flags);
```
mattleibow, line 320: Should this have a `[return: MarshalAs(UnmanagedType.Bool)]` attribute as well?

binding/Binding/HandleDictionary.cs
```diff
@@ -212,4 +213,115 @@ internal static void DeregisterHandle (IntPtr handle, SKObject instance)
...
+		[DllImport ("Kernel32.dll")]
+		static extern bool EnterCriticalSection (IntPtr lpCriticalSection);
+		[DllImport ("Kernel32.dll")]
+		static extern bool LeaveCriticalSection (IntPtr lpCriticalSection);
```
mattleibow, line 324: These have a void return in native code.

binding/Binding/HandleDictionary.cs
toptensoftware: Yes, second commit includes fix for this

binding/Binding/HandleDictionary.cs
toptensoftware: I've put in a check on the AllocHGlobal to throw an out of memory exception.  I'm not sure what conditions InitCritSec would fail (I've never seen one fail) so not sure what exception to throw.

binding/Binding/HandleDictionary.cs
toptensoftware: Added

binding/Binding/HandleDictionary.cs
toptensoftware: Oops.  Late Sunday night copy/paste error.  Fixed

binding/Binding/HandleDictionary.cs
toptensoftware: You still need some sort of locking because the HandleDictionary can be called from the GC thread.

toptensoftware: On further thought,  having this as a public API might be a really good idea.  It could also be a good alternative to including the critical section implementation at all.  If I could simply plugin my own sync primitive that'd solve everything and wouldn't require SkiaSharp to include this platform specific stuff.

That said, anyone using SkiaSharp in an STA app is probably going to want the critical section solution (whether they realize it or not) and IMO is a better and safer default implementation for Windows.

binding/Binding/HandleDictionary.cs
mattleibow: I mean the InitializeCriticalSectionEx may return false and have an error.

toptensoftware: Yes, I understand that... I just wasn't sure what exception to throw? Suggestions?

mattleibow: Looking at people's code online, seems that they just use void and thus never care. I wonder why. Maybe it never fails?

mattleibow: I looked in CoreRT and Sqlite. So I guess just ignore for now... I saw a comment on `InitializeCriticalSection` and `InitializeCriticalSectionAndSpinCount` so maybe it also applies:

> This function always succeeds and returns a nonzero value.
>
> Windows Server 2003 and Windows XP:  If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0). To get extended error information, call GetLastError. Starting with Windows Vista, the InitializeCriticalSectionAndSpinCount function always succeeds, even in low memory situations.

toptensoftware: Not 100% sure, but I suspect that InitializeCriticalSection just initializes the fields of the CRITICAL_SECTION structure... so it could be that it only fails if you pass it bad parameters like a null pointer for example.

mattleibow (approved):

mattleibow (changes requested):
All right, not sure if you are still wanting to work on this after all this time, but I have been thinking about how I can have this as well as not add things to the global namespace as well as keep things neat and tidy.

I sort of decided to make this like - a bit inspired by discussions here on making this configuration point externally usable: https://github.com/mono/SkiaSharp/pull/1856

So I think we can embrace this and maybe move the locking code into the internals namespace - probably can be in a separate file too.

We could have a `public interface IPlatformLock` and the have a static one-time configuration for now. Unless you feel inspired to have some way of making sure that we don't do something stupid and allow a lock system to be swapped out while things are mid lock.

```cs
public static class PlatformLock {
    static Func<IPlatformLock>? _currentFactory;
    public static void SetFactory (Func<IPlatformLock> factory) => _currentFactory = factory;

    public static IPlatformLock Create() {
        return _currentFactory?.Invoke() ?? new ReadWriteLock();
    }

    public class ReadWriteLock {}
    public class NonAlertableWin32Lock {}
}
```

And then usage from a library could be semi-discoverable:

```cs
PlatformLock.SetFactory(() => new PlatformLock.NonAlertableWin32Lock());
```

or even generics...


```cs
PlatformLock.SetFactory<PlatformLock.NonAlertableWin32Lock>();
```

Regardless, I think making it a feature and configurable will help for both your case and maybe Avalonia. Maybe there is some special platform lock on a unix machine that can do special things. Who knows.

toptensoftware:
Hi @mattleibow,  

I'm perfectly happy with whatever you think is the best solution - just so long as there's a way to get a non-alertable lock in there somehow.

I think the biggest question is what lock it uses on Windows by default?  Personally I think it should use the non-alertable lock because of reasons [mentioned here](https://github.com/mono/SkiaSharp/issues/1383#issuecomment-927409389).  This decision doesn't really affect me because my toolkit will configure as appropriate but for general SkiaSharp users you need to make a call on that.

I can make the changes if you let me know exactly how you want it to work (but probably won't get to it until Monday).  If you prefer make the changes yourself I'm fine with that too.

Brad

mattleibow:
Good point. Maybe we can change the default now in the previews and see what breaks. We can always change it back in a new preview.

toptensoftware:
Hey Matt,

I've updated the PR as follows:

* Moved `PlatformLock` to `SkiaSharp.Internals` namespace
* Moved `PlatformLock`, `IPlatformLock` and interface implementations it to a new file `PlatformLock.cs`
* The implementations of `IPlatformLock` are now nested classes in `PlatformLock`
* `PlatformLock` class has a new static property `Factory` that lets you change the implementation
* `Factory` defaults to new static function `PlatformLock.DefaultFactory`
* `PlatformLock.DefaultFactory` does as per before, creating platform specific lock (non-alertable on Windows).

So by default it uses non-alertable lock on Windows.  

If you want to change it...

```cs
SkiaSharp.Internals.PlatformLock.Factory = () => new MyFunkyLock();
```

Does that cover it?

mattleibow (approved):

mattleibow:
This is amazing. Thanks. Once the PR builds then it is ready to go!

mattleibow (dismissed):

mattleibow:
Just adding this for laters... I noticed this on the test run. Might be some bad code with the dictionary or lifetime things, but I will have a closer look after a nap:

```
Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SkiaSharp.Internals.PlatformLock.NonAlertableWin32Lock.EnterCriticalSection(IntPtr lpCriticalSection)
   at SkiaSharp.Internals.PlatformLock.NonAlertableWin32Lock.EnterWriteLock()
   at SkiaSharp.HandleDictionary.DeregisterHandle(IntPtr handle, SKObject instance)
   at SkiaSharp.SKObject.DeregisterHandle(IntPtr handle, SKObject instance)
   at SkiaSharp.SKObject.set_Handle(IntPtr value)
   at SkiaSharp.SKNativeObject.Dispose(Boolean disposing)
   at SkiaSharp.SKColorSpace.Dispose(Boolean disposing)
   at SkiaSharp.SKNativeObject.Finalize()
```

Looks like the GC is collecting things and the critical section is done. Not sure why since it is static, so maybe it is the GC that has first collected the lock and is now collecting the object.

mattleibow:
3 runs later and it is still the same error. I'll have a look and see if we are missing some lifetime thing. Hopefully a missing semicolon :)

mattleibow:
I can repro this with this command:

```
dotnet cake --target=tests-netfx --skipExternals=all --unsupportedTests=Category=API --supportVulkan=false --buildarch=x64
```

If I remove the if statement and always use the slim lock, it works. It might be because the slim lock is a framework think and that outlives the tests/app.

Time for some learning!

mattleibow:
I added some code to basically make every object keep a reference to the platform lock instance - to prevent the GC from collecting too soon.

But, this seemed to throw another type of exception:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SkiaSharp.SkiaApi.sk_font_measure_text_no_return(IntPtr font, Void* text, IntPtr byteLength, SKTextEncoding encoding, SKRect* bounds, IntPtr paint, Single* measuredWidth) in C:\Projects\SkiaSharp\binding\Binding\SkiaApi.generated.cs:line 4106
   at SkiaSharp.SKFont.MeasureText(Void* text, Int32 length, SKTextEncoding encoding, SKRect* bounds, SKPaint paint) in C:\Projects\SkiaSharp\binding\Binding\SKFont.cs:line 320
   at SkiaSharp.SKFont.MeasureText(ReadOnlySpan`1 text, SKPaint paint) in C:\Projects\SkiaSharp\binding\Binding\SKFont.cs:line 272
   at SkiaSharp.Tests.SKFontTest.MeasureTextMeasuresTheText() in C:\Projects\SkiaSharp\tests\Tests\SKFontTest.cs:line 144
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Xunit.Sdk.TestInvoker`1.CallTestMethod(Object testClassInstance) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestInvoker.cs:line 150
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass48_1.<<InvokeTestMethodAsync>b__1>d.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestInvoker.cs:line 257
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass48_1.<InvokeTestMethodAsync>b__1()
   at Xunit.Sdk.ExecutionTimer.<AggregateAsync>d__4.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\ExecutionTimer.cs:line 48
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.ExecutionTimer.AggregateAsync(Func`1 asyncAction)
   at Xunit.Sdk.ExceptionAggregator.<RunAsync>d__9.MoveNext() in C:\Dev\xunit\xunit\src\xunit.core\Sdk\ExceptionAggregator.cs:line 90
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.ExceptionAggregator.RunAsync(Func`1 code)
   at Xunit.Sdk.TestInvoker`1.<InvokeTestMethodAsync>d__48.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestInvoker.cs:line 239
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestInvoker`1.InvokeTestMethodAsync(Object testClassInstance)
   at Xunit.Sdk.TestInvoker`1.<<RunAsync>b__47_0>d.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestInvoker.cs:line 206
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestInvoker`1.<RunAsync>b__47_0()
   at Xunit.Sdk.ExceptionAggregator.<RunAsync>d__10`1.MoveNext() in C:\Dev\xunit\xunit\src\xunit.core\Sdk\ExceptionAggregator.cs:line 107
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.ExceptionAggregator.RunAsync[T](Func`1 code)
   at Xunit.Sdk.XunitTestRunner.<InvokeTestAsync>d__4.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\XunitTestRunner.cs:line 67
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.XunitTestRunner.InvokeTestAsync(ExceptionAggregator aggregator)
   at Xunit.Sdk.ExceptionAggregator.<RunAsync>d__10`1.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.ExceptionAggregator.RunAsync[T](Func`1 code)
   at Xunit.Sdk.TestRunner`1.<RunAsync>d__43.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestRunner`1.RunAsync()
   at Xunit.Sdk.TestCaseRunner`1.<RunAsync>d__19.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestCaseRunner.cs:line 82
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestCaseRunner`1.RunAsync()
   at Xunit.Sdk.SkippableFactDiscoverer.SkippableFactTestCase.<RunAsync>d__6.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.SkippableFactDiscoverer.SkippableFactTestCase.RunAsync(IMessageSink diagnosticMessageSink, IMessageBus messageBus, Object[] constructorArguments, ExceptionAggregator aggregator, CancellationTokenSource cancellationTokenSource)
   at Xunit.Sdk.XunitTestMethodRunner.RunTestCaseAsync(IXunitTestCase testCase) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\XunitTestMethodRunner.cs:line 45
   at Xunit.Sdk.TestMethodRunner`1.<RunTestCasesAsync>d__32.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestMethodRunner.cs:line 136
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestMethodRunner`1.RunTestCasesAsync()
   at Xunit.Sdk.TestMethodRunner`1.<RunAsync>d__31.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestMethodRunner.cs:line 106
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestMethodRunner`1.RunAsync()
   at Xunit.Sdk.TestClassRunner`1.<RunTestMethodsAsync>d__38.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestClassRunner.cs:line 213
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestClassRunner`1.RunTestMethodsAsync()
   at Xunit.Sdk.TestClassRunner`1.<RunAsync>d__37.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestClassRunner.cs:line 171
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestClassRunner`1.RunAsync()
   at SkiaSharp.Tests.CustomTestFramework.CollectionRunner.RunTestClassAsync(ITestClass testClass, IReflectionTypeInfo class, IEnumerable`1 testCases) in C:\Projects\SkiaSharp\tests\Tests\Xunit\CustomTestFramework.cs:line 138
   at Xunit.Sdk.TestCollectionRunner`1.<RunTestClassesAsync>d__28.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestCollectionRunner.cs:line 130
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestCollectionRunner`1.RunTestClassesAsync()
   at Xunit.Sdk.TestCollectionRunner`1.<RunAsync>d__27.MoveNext() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\TestCollectionRunner.cs:line 101
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Xunit.Sdk.TestCollectionRunner`1.RunAsync()
   at Xunit.Sdk.XunitTestAssemblyRunner.<>c__DisplayClass14_2.<RunTestCollectionsAsync>b__2() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\Runners\XunitTestAssemblyRunner.cs:line 184
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
   at System.Threading.Tasks.Task.ExecuteEntry(Boolean bPreventDoubleExecution)
   at Xunit.Sdk.MaxConcurrencySyncContext.RunOnSyncContext(SendOrPostCallback callback, Object state) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\MaxConcurrencySyncContext.cs:line 107
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at Xunit.Sdk.ExecutionContextHelper.Run(Object context, Action`1 action) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Utility\ExecutionContextHelper.cs:line 22
   at Xunit.Sdk.MaxConcurrencySyncContext.WorkerThreadProc() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\MaxConcurrencySyncContext.cs:line 89
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart(Object obj)
```

mattleibow:
Doing a bit more things and my logging shows that the lock is collected before all the objects are collected. No idea why since the objects invoke methods on the handle dictionary which reference the lock. Seems that the collection order is not correct.

And the reason why there are no issues with .net core is because none of the static objects are collected. It probably has to do with AppDomains because I don't think .net core has them and the netfx unloads them at the end of all the tests - which triggers the GC.

mattleibow:
I see I do have a slight bad case where I load the statics from the C++ library and wrap them. I have code to avoid disposing them, but this is still incorrect as I just prevent the user disposing it. If the GC collects them and disposes of the native side, we have corrupted the native library.

I fixed this by making sure I prevent all managed disposal since this is up to the native library.

toptensoftware:
Just having a look at this and a less impactful fix might be to just remove the finalizer from `NonAlertableWin32Lock`.  It's a tiny allocation and will get cleaned up when the process ends anyway.  Given there's only one lock for the entire process can't see it being an issue (unless somebody re-uses that lock class for something else).

mattleibow:
It seems to have fixed itself now after I stopped collecting the things I should not be collecting. The old code was disposing of things that I did not create. And I think the fails only come up if I am doing it incorrectly.

I'll rebuild and merge this if it is all green now. Then I will release a preview so we can test in the wild. If people get weird crashes, they can use the old lock and we can also change the default before we go stable.

toptensoftware:
Thanks Matt, sounds good... I'm hoping to put up an alpha build of the new version of my app with this early next week so seems like the timing might work out nicely.

mattleibow:
Thanks for this, merging with happy thoughts. Hopefully all is well and we just fixed other folks' weird errors.

toptensoftware:
Hey Matt,  just wondering if there's preview build of this available yet somewhere?

mattleibow:
Yep, on the preview feed. https://aka.ms/skiasharp-eap/index.json

RichardD2:
I'm still seeing the `AccessViolationException` on the finalizer thread using v2.88.0 (via QuestPDF v2022.6.3) in a .NET Framework 4.8 / ASP.NET MVC 5 application.

```
Application: w3wp.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AccessViolationException
   at SkiaSharp.Internals.PlatformLock+NonAlertableWin32Lock.EnterCriticalSection(IntPtr)
   at SkiaSharp.Internals.PlatformLock+NonAlertableWin32Lock.EnterWriteLock()
   at SkiaSharp.HandleDictionary.DeregisterHandle(IntPtr, SkiaSharp.SKObject)
   at SkiaSharp.SKObject.DeregisterHandle(IntPtr, SkiaSharp.SKObject)
   at SkiaSharp.SKObject.set_Handle(IntPtr)
   at SkiaSharp.SKNativeObject.Dispose(Boolean)
   at SkiaSharp.SKNativeObject.Finalize()
```

toptensoftware:
> I'm still seeing the `AccessViolationException` on the finalizer thread using v2.88.0 (via QuestPDF v2022.6.3) in a .NET Framework 4.8 / ASP.NET MVC 5 application.
> 

Is this happening while the app is running or during process shutdown. 

My guess is the PlatformLock's finalizer has run before the last native object is collected and deleting the critical section here:

https://github.com/mono/SkiaSharp/blob/e603133c3548bbd3835cf1012eac80ce5640f209/binding/Binding/PlatformLock.cs#L107

Not sure what other reason EnterCriticalSection would crash.

RichardD2:
@toptensoftware I suspect you're right - it'll be when the AppPool is recycled.

mattleibow:
I merged this PR that basically turns a collected lock into a no-op and I assume all the objects will be collected in parallel - potentially: https://github.com/mono/SkiaSharp/pull/2195

Is that worse?

RichardD2:
If it happens, it's possible that the `instances` dictionary's internal state could become corrupted, as the `Dictionary<,>` class is not thread-safe.

Since [the `DeregisterHandle` method](https://github.com/mono/SkiaSharp/blob/dc54fa5cddc2f2f8089c851e1752a490a9072d7e/binding/Binding/HandleDictionary.cs#L167) just seems to be removing references from the dictionary, and the error only seems to occur when the AppDomain is terminating, that *might* not be a problem. The only issue would be if the dictionary state gets so corrupted that it throws an exception from the `Remove` method.

toptensoftware:
That'll should fix the crash, but I agree with @RichardD2 - it could cause the dictionary to become corrupt.

Another safer but kludgier way to fix it might be to just never delete the critical section - it's a tiny blob of memory, there's only one instance.  Just leave Windows to clean it up when the process exits.

What I don't understand about this is there's methods being called on that lock instance after it's been GCed.  Is that possible and/or a known edge case during shutdown?

mattleibow:
It may be calling the destruction in a random order. So first takes out the dictionary and then an object.

But yeah, I agree for safety to not actually clear the critical section on the GC of the handle.

Issue #1853, opened 2021-11-05
[BUG] macOS: When display scale changes SKGLView does not adjust appropriately


HarlanHugh:
**Description**

Draw a circle (or anything) on a SKGLView. Move the parent window such that Window.BackingScaleFactor changes (for example moving from a retina display to a non-retina display ). The circle will be displayed at double or half size, depending on the change.

**Expected Behavior**

SKGLView should behave just like SKCanvasView, where the output should scales appropriately to the display's scale. 

**Actual Behavior**

Moving from a retina display to a non-retina display causes everything to be displayed at double size. Moving from a non-retina display to a retina display causes everything to be displayed at half size. 

**Basic Information**

- Version with issue:  2.80.2
- Last known good version:  None
- IDE: Visual Studio for Mac
- Platform Target Frameworks:
  - macOS:  12.0
- Target Devices: 
  - MacBook Pro (14" 2021)

PR #1854, opened 2021-11-05
Reshape SKGLView when Window.BackingScaleFactor changes (#1853)

HarlanHugh:
**Description of Change**

Reshape SKGLView when Window.BackingScaleFactor changes, ensuring that the output is displayed at the proper scale. If it is possible to create a test for this type of visual change, I don't know how.

**Bugs Fixed**

[BUG] macOS: When display scale changes SKGLView does not adjust appropriately

https://github.com/mono/SkiaSharp/issues/1853

**API Changes**

None

**Behavioral Changes**

None

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.Mac/SKGLView.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.Mac/SKGLView.cs
@@ -158,8 +158,28 @@ namespace SkiaSharp.Views.Mac
 
 		public event EventHandler<SKPaintGLSurfaceEventArgs> PaintSurface;
 
+		private nfloat lastBackingScaleFactor = 0;
+
 		protected virtual void OnPaintSurface(SKPaintGLSurfaceEventArgs e)
 		{
+			// Track if the scale of the display has changed and if so force the SKGLView to reshape itself.
+			// If this is not done, the output will scale correctly when the window is dragged from a non-retina to a retina display.
+			if (lastBackingScaleFactor != Window.BackingScaleFactor)
+			{
+				bool isFirstPaint = lastBackingScaleFactor == 0;
+				lastBackingScaleFactor = Window.BackingScaleFactor;
+				if (!isFirstPaint)
+				{
+					Reshape();
+					// A redraw will also be necessary. Invoke later or the request will be ignored
+					Invoke(() => {
+						NeedsDisplay = true;
+					}, 0);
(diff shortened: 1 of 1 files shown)


mattleibow:
Is this the only way to check? Or rather, is this sufficient?

I am thinking if you have moved the window, does it always trigger a paint - even without a render loop? Is there some window based event that we should check? What happens if you change the display scale in the system preferences?

The code looks good, I am just wondering if we should be listening to some event...

HarlanHugh:
Yes, moving the window between two displays of different scales triggers a paint even without a render loop. A paint is also triggered when then the display scale is changed via System Preferences.

There is an event that is fired when display scale is changed via System Preferences (NSApplication.Notifications.ObserveDidChangeScreenParameters) but it does not fire when moving the window between displays. If there is another event that fires when moving the window, I am not aware of it.

mattleibow (commented):
Could this change be moved to `DrawRect` instead? Then we don't have to do any of the GL logic and basically have a free re-render.

Other than that, this looks good to merge. 

Thanks!

HarlanHugh:
I moved the code as suggested. Great change. The result is visually perfect now, eliminated a brief flash that happened before.

HarlanHugh:
I may have just discovered a better way to do this... don't merge yet.

HarlanHugh:
Nevermind... There is an event that fires - NSView.DidChangeBackingProperties and the fix can be implemented there however doing so results in the content being redrawn at the wrong scale momentarily. Handling it in DrawRect is much better.

HarlanHugh:
I believe this code is ready to merge.

mattleibow:
OMG, 7 months. Sorry! Let me re review and merge. Thanks for fixing this.

mattleibow:
Actually, I am looking at this and see the logic in draw rect basically cancels and triggers a redraw.

What happens with the reshape event?

Is the chance basically to not have a flash?

HarlanHugh:
If we don't track whether the scale has been changed and call Reshape directly, it is never called unless the view is resized.

When tracking the scale from inside of OnPaintSurface where I had it originally, there was a flash. Moving it to DrawRect and bailing out when a changed is detected prevents a flash entirely.

FWIW, we have already been using this in production with thousands of users for the last 6 months.

mattleibow (approved):

PR #1864, opened 2021-11-15
Wrap and use SKMatrix4x4

mattleibow:
Right now just SKCanvas

**Description of Change**

**Bugs Fixed**

- Related to issue #

**API Changes**

**Behavioral Changes**

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/binding/Binding/MathTypes.cs
+++ b/binding/Binding/MathTypes.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+#if !NETSTANDARD2_1_OR_GREATER && !NET5_0_OR_GREATER
+using MathF = System.Math;
+#endif
 
 namespace SkiaSharp
 {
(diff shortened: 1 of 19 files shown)


binding/Binding/MathTypes.cs
```diff
@@ -89,6 +91,18 @@ public static SKPoint Reflect (SKPoint point, SKPoint normal)
 			new SKPoint (pt.X - sz.X, pt.Y - sz.Y);
 		public static SKPoint operator - (SKPoint pt, SKPoint sz) =>
 			new SKPoint (pt.X - sz.X, pt.Y - sz.Y);
+
+		public static SKPoint operator * (SKPoint pt, SKPoint sz) =>
+			new SKPoint (pt.X * sz.X, pt.Y * sz.Y);
+		public static SKPoint operator * (SKPoint pt, float sz) =>
+			pt * new SKPoint (sz, sz);
+		public static SKPoint operator * (float sz, SKPoint pt) =>
+			new SKPoint (sz, sz) + pt;
```
OneEyeMaker, line 100:
```			// multiply, not plus
			new SKPoint (sz, sz) * pt;
```

mattleibow:
No longer needed as in skia v3 we are using SKM44 everywhere and I took a shortcut and it is a wrapper for Matrix4x4 from System.Numerics.

Probably losing out with the conversion, but we can do something later - maybe copy the implementation and use the same intrinsics. But with implicit converters, it is far less wasteful than interop with C...

Issue #1838, opened 2021-10-23
[BUG] Blazor Wasm runs out of memory trying to resize HTML canvas


artemiusgreat:
Trying to create a new Blazor Wasm app using Skia similar to this sample. 
https://github.com/mattleibow/SkiaSharpBlazorWebAssembly

The result is here. 
https://github.com/artemiusgreat/SkiaBlazorWasmIssue 

The sample app is running fine, but the new app keeps reloading page (notice the scrollbars) until it runs out of memory. There might be an issue with Blazor template, because new Blazor app created in VS specified port 5060 and 5061 as target but was still running on 5000 and 5001, so I copied files like `launch.json`, `nuget` packages in `proj` file and `index.razor` component from the sample app into new app.  I tried both, new app and the sample in the same environment, the same nuget version, the same .NET version, but only new app is failing.

* Windows 10 x64 
* .NET 6.0.100-rc.2.21505.57 
* VS 2022 Preview 3 
* SkiaSharp 2.88.0-preview.152

![Error](https://user-images.githubusercontent.com/5140879/138571526-5fd7fe83-96a7-4d8a-b0dc-a083a35fb71d.png)

I think I should add some config or additional dependency in the new app to make it work, but can't understand which exactly. 
Could somebody possibly take a look at the second link above and tell what is missing?

artemiusgreat:
Not sure why this issue was happening and how it was fixed, but everything is working now. 

**Razor Library - CrossCanvas.Web**

```XML
<Project Sdk="Microsoft.NET.Sdk.Razor">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>disable</Nullable>
    <ImplicitUsings>disable</ImplicitUsings>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Components.Web" Version="6.0.0-rc.2.21480.10" />
    <PackageReference Include="SkiaSharp.NativeAssets.WebAssembly" Version="2.88.0-preview.152" />
    <PackageReference Include="SkiaSharp.Views.Blazor" Version="2.88.0-preview.152" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\CrossCanvas\CrossCanvas.csproj" />
  </ItemGroup>

</Project>
```

**Actual Blazor app**

```XML
<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">

  <PropertyGroup>
[...]

PR #1889, opened 2021-12-12
[Blazor] prevent infinite loop in SKCanvasView

JensKrumsieck:
**Description of Change**

Prevents infinite resize loop - i don't know if this is the most elegant solution, but it works.

Added the following line to `OnSizeChanged` Event of Blazor `SKCanvasView`
```csharp
if (Math.Round(canvasSize.Width * dpi) == newSize.Width && Math.Round(canvasSize.Height * dpi) == newSize.Height)
    return;
```
If  Monitor scaling is applied in Windows Settings App, the view keeps multiplying the canvas size with the DPI (in my case being 1.25 for 125% scaling) in the `Invalidate`-Method. 
The Canvas will be resized to the new size which results in the `SizeWatcherInterop` to fire its `OnSizeChanged` Event which will set `canvasSize` to `newSize` and calls `Invalidate` again to start infinite loop, also mentioned here:
https://github.com/mono/SkiaSharp/issues/1838#issuecomment-991914840

**Bugs Fixed**
https://github.com/mono/SkiaSharp/issues/1838

- Related to issue #1838

**API Changes**
None

**Behavioral Changes**
None

**PR Checklist**

- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/SKCanvasView.razor.cs
+++ b/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/SKCanvasView.razor.cs
@@ -157,7 +157,7 @@ namespace SkiaSharp.Views.Blazor
 
 		private void OnSizeChanged(SKSize newSize)
 		{
-			if (Math.Round(canvasSize.Width * dpi) == newSize.Width && Math.Round(canvasSize.Height * dpi) == newSize.Height)
+			if ((int)(canvasSize.Width * dpi) == newSize.Width && (int)(canvasSize.Height * dpi) == newSize.Height)
 				return;
 			canvasSize = newSize;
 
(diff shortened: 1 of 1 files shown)


mattleibow (commented):
This makes sense. Just wondering if we should just be casting to `int` instead of rounding? That is what is happening on line 80.

Not sure if that was causing an issue and you already tried that. Feels like it should work.

JensKrumsieck:
You are right. I looked back at the console output from https://github.com/mono/SkiaSharp/issues/1838#issuecomment-991913527
```
DPI: 1,25 -> predicting: 1952,5: 1463,75
1952 : 1463
```
It cuts the digits after the comma instead of rounding.
I appended https://github.com/mono/SkiaSharp/pull/1889/commits/5385f9dd623a26c251ed510297deb9e4c80aca84

mattleibow (approved):

Issue #1896, opened 2021-12-17
[FEATURE] Disable automatic resizing of the canvas


Code-Chops:
At the moment it's not possible to disable automatic resizing of the canvas. Besides that there is a bug in the automatic resizing, I don't want the functionality at all for my project.

Would it be a good idea to add a parameter 'IgnoreResizing' / 'DisableAutomaticResizing' to the canvas? Setting the value to 'true' will not load the SizeWatcher JavasScript at all: https://github.com/mono/SkiaSharp/blob/3c721d250969ba6d0617827a3aa1d4f15fe95747/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/SKCanvasView.razor.cs#L70

I could add the functionality if you are open for it? Thanks in advance!

robloo:
`AutoResize` might be a better parameter name. Usually names are such that true is the enabled state.

mgood7123:
why would you want to disable automatic resizing?

ViRuSTriNiTy:
> why would you want to disable automatic resizing?

Because sometimes you have the requirement of a fixed width and height for the canvas. I stumpled upon this issue searching for options on how to set width and height to a specific value but did not find any. Then I had a look at the source code just to see that JSInterop is overriding my initially set width and height settings. So, I am in the same boat as @Code-Chops requiring the auto size thing to be controllable.

mattleibow:
You should not set width/height of the canvas directly, and instead use the styles:

```razer
<SKGLView OnPaintSurface="OnPaintSurface" style="width:100px; height:100px;" />
```

In order to not have a blurry canvas, we need to change the stile to be `<style-size> * <screen-density>`:

https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas#scaling_for_high_resolution_displays

ViRuSTriNiTy:
@mattleibow Yep, the inline style solution is also discussed in https://github.com/mono/SkiaSharp/issues/2038 and specifically comment https://github.com/mono/SkiaSharp/issues/2038#issuecomment-1129008602 provides a ruleset that avoids the automatic resizing, at least visually, the internal services are still running. I think the issue can be closed.

PR #1897, opened 2021-12-19
Option to disable automatic resizing of the canvas

Code-Chops:
At the moment it's not possible to disable automatic resizing of the canvas. Besides that there is a bug in the automatic resizing, I don't want the functionality at all for my project.

I added parameter 'AutoResize' to SKGLView and SKCanvasView (which is true by default). When it is disabled, the SizeWatcherInterop won't be used.

See issue:
https://github.com/mono/SkiaSharp/issues/1896

Added: 
 
```
/// <summary>
/// If true (default), the canvas will be automatically resized to fit the client width and height. 
/// </summary>
[Parameter]
public bool AutoResize { get; set; } = true;
```

### DIFF
--- a/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/Internal/SizeWatcherInterop.cs
+++ b/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/Internal/SizeWatcherInterop.cs
@@ -17,11 +17,13 @@ namespace SkiaSharp.Views.Blazor.Internal
 
 		private DotNetObjectReference<FloatFloatActionHelper>? callbackReference;
 
+		/// <summary>
+		/// Imports the SizeWatcher script, but doesn't start observing.
+		/// </summary>
 		public static async Task<SizeWatcherInterop> ImportAsync(IJSRuntime js, ElementReference element, Action<SKSize> callback)
 		{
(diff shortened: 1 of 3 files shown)


robloo:
It looks like if you set AutoResize from false to true nothing will happen until the next paint. It might be a good idea, in the AutoResize property setter, to handle changes and then refresh things.

Daendel:
I changed the code so that the AutoResize script will always be imported. Observing is not started anymore when the script is loaded, but will start when the AutoResize-proprety is set to true. If the property is set to false, the element will not be observed anymore.

mattleibow:
Are you talking about this bug: https://github.com/mono/SkiaSharp/pull/1889

Daendel:
Yes :)

mgood7123:
why would you want to disable automatic resizing?

mattleibow:
I think if you want to disable auto-resizing maybe you just need to set a fixed with/height?

mgood7123:
> I think if you want to disable auto-resizing maybe you just need to set a fixed with/height?

heavily agreed

ViRuSTriNiTy:
Fixed with and height can be enforced via CSS, see solution posted here: https://github.com/mono/SkiaSharp/issues/2038#issuecomment-1129008602

Issue #1909, opened 2021-12-28
[BUG] SKShaper/DrawShapedText ignores SKPaint.TextAlign property


koolkdev:
**Description**
When drawing text with DrawShapedText, the specified TextAlign in SKPaint is ignored.

**Code**
```
using var paint= new SKPaint { ... TextAlign = SKTextAlign.Center ...};
canvas.DrawShapedText(shaper, "test", width / 2f, (height / 2f) + (paint.FontMetrics.XHeight / 2), paint);
```

Would result in:
<img src="https://user-images.githubusercontent.com/5482304/147596629-82728dd2-8682-4a1c-9cac-45212e14eab3.png" width=25% height=25%>
<img src="https://user-images.githubusercontent.com/5482304/147596632-bec64e06-4ab7-451e-b65a-58bcd2f22cf9.png" width=25% height=25%>

When I would expect the same behaviour as DrawText as for the alignment.
<img src="https://user-images.githubusercontent.com/5482304/147596410-f897c031-6f9e-40c9-94b2-716d05339a00.png" width=25% height=25%>
<img src="https://user-images.githubusercontent.com/5482304/147596412-27fc2ce5-795f-4dcf-a036-8adbf78adcdd.png" width=25% height=25%>

I am opening a PR to fix it,

trisolari:
This should be closed yeah?

PR #1910, opened 2021-12-28
SKShaper: Support SKPaint.TextAlign property

koolkdev:
**Description of Change**

Added handling of the SKPaint.TextAlign property in SKShaper.Shape.
The width is calculated by the sum of XAdvance

Befpre;
DrawText with TextAlign = Center:
 

DrawShapedText with TextAlign = Center
 

After this PR:
DrawShapedText with TextAlign = Center
 

**Bugs Fixed**

- Related to issue #1909

**Behavioral Changes**

**PR Checklist**

- [X] Has tests (if omitted, state reason in description)
- [X] Rebased on top of main at time of PR
- [ ] Updated documentation


source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz.Shared/SKShaper.cs
```diff
@@ -67,6 +68,14 @@ public Result Shape(Buffer buffer, float xOffset, float yOffset, SKPaint paint)
 			float textSizeY = paint.TextSize / FONT_SIZE_SCALE;
 			float textSizeX = textSizeY * paint.TextScaleX;
 
+			// adjust alignment
+			if (paint.TextAlign != SKTextAlign.Left) {
+				var width = pos.Sum(p => p.XAdvance) * textSizeX;
```
Gillibald, line 73: This might be a huge performance hit. I think text alignment should be evaluated when the text blob is being drawn onto a canvas. Better implement this in the DrawShapedText extension method. SKTextBlob.Bounds should give you needed Size for your calculations.

koolkdev: Hi, thanks for your suggestion. I tried to implement it, but Bounds return much larger box than there is actually rendered.

For example:
`SKTextBlob.Create("test", font).Width`
returns 157.3125
while 
`paint.MeasureText("test")`
returns 67.125

According to the documentation of bounds:
```
    /** Returns conservative bounding box. Uses SkPaint associated with each glyph to
        determine glyph bounds, and unions all bounds. Returned bounds may be
        larger than the bounds of all glyphs in runs.

        @return  conservative bounding box
    */
```

When diving deeper into the code of `SkTextBlobBuilder`, there are two functions to calculate the bounds, one is called `TightRunBounds` and the other `ConservativeRunBounds`, where `ConservativeRunBounds` is used for `sk_textblob_builder_alloc_run_pos` and `sk_textblob_builder_alloc_run_pos_h`, and `TightRunBounds` is only used for `sk_textblob_builder_alloc_run` (and there is a FIXME comment to not use tight run in this case too).
So it seems that bounds is pretty useless for this use case.

Another option is to use paint.MeasureText but it would not be accurate with things like Arabic text (which is the point of SKShaper)

Another option to that I can think of is to calculate the width during the loop of Shape (so there won't be any performance hit), and return it as part of the Result and than use it in DrawShapedText. What do you think?
Thanks.

koolkdev: https://github.com/koolkdev/SkiaSharp/commit/17da8ad14534a9630fd3e192187d2edf9cbde971

Gillibald: DrawPositionedText should use https://github.com/mono/SkiaSharp/blob/fffae82cf6ec1c4a6f490ddb780709ce9211725e/binding/Binding/SKTextBlob.cs#L806 if that doesn't work there is something wrong with the binding

Gillibald: Hmm, I think `bounds` is misused here and should be an out parameter. Only when a nullptr is used Skia calculates the bounds.

koolkdev: It seems that bounds is an input parameter, according to documentation:
```
        bounds defines an optional bounding box, used to suppress drawing when SkTextBlob
        bounds does not intersect SkSurface bounds. If bounds is nullptr, SkTextBlob bounds
        is computed from RunBuffer::pos, and RunBuffer::glyphs metrics.
```
Skia calculate the bounds anyway, but that parameter is just the initial value which it does SkRect.join to 
https://github.com/mono/skia/blob/916a1e9721de74a05c75391f52393972ed7f736c/src/core/SkTextBlob.cpp#L514
https://github.com/mono/skia/blob/916a1e9721de74a05c75391f52393972ed7f736c/src/core/SkTextBlob.cpp#L383
At least if I am understanding the code correctly....

Anyway, the issue is the that Skia uses conservative way to calculate the bounds (probably for performance) which result in a much larger bounds than the actual bounds. (like twice as big or more in some cases), so it doesn't really help in this case.

What do you say about the current solution?

source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz.Shared/SKShaper.cs
```diff
@@ -129,20 +132,24 @@ public Result()
 				Codepoints = new uint[0];
 				Clusters = new uint[0];
 				Points = new SKPoint[0];
+				Width = 0f;
 			}
 
-			public Result(uint[] codepoints, uint[] clusters, SKPoint[] points)
+			public Result(uint[] codepoints, uint[] clusters, SKPoint[] points, float width)
```
mattleibow, line 138: This is an API break, so is it possible to rather add an overload to this?

mattleibow (commented):
Thanks for this PR. It has been so long since I reviewed a PR on this repo 😢  Thanks for this, but I just got 1 suggestion about API/ABI compatibility.

mattleibow (approved):

Issue #1580, opened 2021-01-15
[FEATURE] Binding Skottie


PieroCastillo:
Hello, I see on Skia Official Docs there is an Lottie Animation Player, [Skottie](https://skia.org/user/modules/skottie), and I wanna use this feature, but I saw the SkiaSharp API and this doesn't exist, do you think implement this module some day? or never? if so I would like to know why.

mattleibow:
Duplicate of #661

PR #1987, opened 2022-03-31
feat: Add basic Skottie support

jeromelaban:
**Description of Change**

Adds basic support for Skottie (dependent on https://github.com/mono/skia/pull/84)

Related to issue #1580

**Bugs Fixed**

None

**API Changes**
Added: 
- `SkiaSharp.Skottie.Animation`
- `SkiaSharp.Skottie.InvlidationController`

**PR Checklist**

- [ ] Updated documentation


jeromelaban:
Uno 4.1.9 is binary compatible with .NET 6 Mobile RC1, so given that RC1 is used by the repository, the build should be now fine.

binding/Binding/Skottie/InvalidationController.cs
```diff
@@ -0,0 +1,55 @@
...
+			Handle = SkiaApi.sksg_invalidation_controller_new ();
+		}
+
+		internal InvalidationController (IntPtr handle, bool owns)
+			: base (handle, owns)
+		{
+		}
+
+		// SK_C_API void sksg_invalidation_controller_inval (sksg_invalidation_controller_t* instance, sk_rect_t* rect, sk_matrix_t* matrix);
+		public unsafe void Inval(SKRect rect, SKMatrix matrix)
```
mattleibow, line 22: Can we use the full Invalidate here since we are not saving that much for the very non-.NET style.

binding/Binding/Skottie/InvalidationController.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.ComponentModel;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace SkiaSharp.Skottie
+{
+	public unsafe class InvalidationController : SKObject
```
mattleibow, line 8: If objects are only ever created by users, we can also add the `ISKSkipObjectRegistration` interface to speed up and also reduce locks since we will not need to keep a reference to the handle.

binding/Binding/Skottie/Animation.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.ComponentModel;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace SkiaSharp.Skottie
+{
+	public unsafe class Animation : SKObject
```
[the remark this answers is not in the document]
mattleibow: Similar to the invalidation, we can add `ISKSkipObjectRegistration` to prevent the locking of the handle dictionary and storage of the object if the instance is never returned by an API. It is via Create, but that is more of a constructor. If the object cannot be obtained at a later stage, then it is free to skip registration.

binding/Binding/Skottie/Animation.cs
```diff
@@ -0,0 +1,91 @@
...
+		//SK_C_API void skottie_animation_render_with_flags(skottie_animation_t *instance, sk_canvas_t *canvas, sk_rect_t *dst, skottie_animation_renderflags_t flags);
+		public void Render (SKCanvas canvas, SKRect dst, RenderFlags flags)
+			=> SkiaApi.skottie_animation_render_with_flags (Handle, canvas.Handle, &dst, flags);
+
+		//SK_C_API void skottie_animation_seek(skottie_animation_t *instance, SkScalar t, sksg_invalidation_controller_t *ic);
+		public void Seek (float t, InvalidationController ic)			=> SkiaApi.skottie_animation_seek (Handle, t, ic?.Handle ?? IntPtr.Zero);
+
+		//SK_C_API void skottie_animation_seek_frame(skottie_animation_t *instance, double t, sksg_invalidation_controller_t *ic);
+		public void SeekFrame(float t, InvalidationController ic)
+			=> SkiaApi.skottie_animation_seek_frame (Handle, t, ic.Handle);
```
mattleibow, line 46: handle ?? zero

binding/Binding/Skottie/InvalidationController.cs
```diff
@@ -0,0 +1,55 @@
...
+	{
+		public InvalidationController ()
+			: this (IntPtr.Zero, true)
+		{
+			Handle = SkiaApi.sksg_invalidation_controller_new ();
```
mattleibow, line 13: This can be assigned in the ctor directly. Not sure if there was an optimisation, but it skips a few checks.

binding/Binding/Skottie/Animation.cs
```diff
@@ -0,0 +1,91 @@
...
+			: base (handle, owns)
+		{
+		}
+
+		// SK_C_API skottie_animation_t* skottie_animation_make_from_string(const char* data, size_t length);
+		public static Animation Make (string data)
+			=> GetObject (SkiaApi.skottie_animation_make_from_string (data, data.Length));
+
+		// SK_C_API skottie_animation_t* skottie_animation_make_from_stream(sk_stream_filestream_t* stream);
+		public static Animation Make (SKStream stream)
```
mattleibow, line 26: We can add the .NET stream overload here too and make things easier.

tests/Tests/Skottie/AnimationTest.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using Xunit;
+using System.IO;
+
+namespace SkiaSharp.Tests
+{
+	public class AnimationTest : SKTest
+	{
+		[Trait(CategoryKey, GpuCategory)]
```
mattleibow, line 10: The GPU category is probably not needed here?

tests/Tests/Skottie/AnimationTest.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using Xunit;
+using System.IO;
+
+namespace SkiaSharp.Tests
+{
+	public class AnimationTest : SKTest
```
mattleibow, line 8: Could we add a test to call the various properties to make sure that all the marshalling is working correctly?

mattleibow (commented):
Mostly hidden interfaces and naming in a pool of bad names...

Otherwise, this is a nice addition!

binding/Binding/Skottie/Animation.cs
```diff
@@ -0,0 +1,91 @@
...
+		//SK_C_API void skottie_animation_render(skottie_animation_t *instance, sk_canvas_t *canvas, sk_rect_t *dst);
+		public unsafe void Render(SKCanvas canvas, SKRect dst)
+			=> SkiaApi.skottie_animation_render (Handle, canvas.Handle, &dst);
+
+		//SK_C_API void skottie_animation_render_with_flags(skottie_animation_t *instance, sk_canvas_t *canvas, sk_rect_t *dst, skottie_animation_renderflags_t flags);
+		public void Render (SKCanvas canvas, SKRect dst, RenderFlags flags)
+			=> SkiaApi.skottie_animation_render_with_flags (Handle, canvas.Handle, &dst, flags);
+
+		//SK_C_API void skottie_animation_seek(skottie_animation_t *instance, SkScalar t, sksg_invalidation_controller_t *ic);
+		public void Seek (float t, InvalidationController ic)			=> SkiaApi.skottie_animation_seek (Handle, t, ic?.Handle ?? IntPtr.Zero);
```
mattleibow, line 42: I see the duration is double and the time is float. Can/should these params also be doubles so that someone can do animation.Duration * 0.3 and not have to cast?

binding/Binding/Skottie/Animation.cs
```diff
@@ -0,0 +1,91 @@
...
+			DisableTopLevelClipping = 0x02,
+		}
+
+		internal Animation (IntPtr handle, bool owns)
+			: base (handle, owns)
+		{
+		}
+
+		// SK_C_API skottie_animation_t* skottie_animation_make_from_string(const char* data, size_t length);
+		public static Animation Make (string data)
```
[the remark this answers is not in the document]
jeromelaban: I've changed all of them to `TryCreate` or `TryParse`. The native implementation returns null when the parsing failed, so the only .NET semantic that makes sense here would be a `TryParse`, or `Parse` with a nullable return (which is not current available in this solution).

binding/Binding/Skottie/Animation.cs
jeromelaban: Makes sense, it'll be easier to use for sure.

jeromelaban:
I added https://github.com/mono/SkiaSharp/pull/1987/commits/21e285dc6f5e843806e7ba482c493cec80d3d6c3 to validate that the build passes fully, will remove once it does.

mattleibow (approved):

mattleibow (dismissed):

amirvenus:
How would we specify the Assets folder for png files?

PR #2119, opened 2022-06-18
Add nullable attributes, TimeSpan and overloads to Skottie

mattleibow:
**Description of Change**

This PR just adds some overloads and nullable annotations to the Skottie `Animation` type.

While working with the new type, I was unsure what the `t` in all the seek methods were, so I renamed the parameters to be more explicit: percent, frame, seconds. I also added an overload for `SeekFrameTime` to take a `TimeSpan` because that was useful.

I also made a change to make `Duration` a `TimeSpan`. This was again to reduce confusion about the actual unit, but also keeps all the time concepts as `TimeSpan` because it feels more accurate?

I added some "overloads" for the `TryCreate` and `TryParse` to just return null instead (`Create` and `Parse`). The reason for null and not an exception is that for some reason, null was used to when there was an error reading things. It should have been an exception, but here we are. Skia does not really have "errors" but more it tries and then says nah, here is a big fat null for you. At least it is mostly consistent.

Is the SkiaSharp API perfect? Yes. There have been no regrets in the design and naming of this perfect library. _\*looks for tissue to dry eyes\*_


binding/SkiaSharp.Skottie/Animation.cs
```diff
@@ -22,27 +21,54 @@ void ISKNonVirtualReferenceCounted.UnreferenceNative ()
 		protected override void DisposeNative ()
 			=> SkottieApi.skottie_animation_delete (Handle);
 
-		public static bool TryParse (string data, out Animation animation)
+		public static Animation? Parse (string json) =>
+			TryParse (json, out var animation)
+				? animation
+				: null;
```
mattleibow, line 27: Adding "overloads" to return the object, or null on some error that we will never know about.

binding/SkiaSharp.Skottie/Animation.cs
```diff
@@ -53,17 +79,20 @@ public unsafe void Render (SKCanvas canvas, SKRect dst)
 		public void Render (SKCanvas canvas, SKRect dst, AnimationRenderFlags flags)
 			=> SkottieApi.skottie_animation_render_with_flags (Handle, canvas.Handle, &dst, flags);
 
-		public void Seek (double t, InvalidationController ic = null)
-			=> SkottieApi.skottie_animation_seek (Handle, (float)t, ic?.Handle ?? IntPtr.Zero);
+		public void Seek (double percent, InvalidationController? ic = null)
+			=> SkottieApi.skottie_animation_seek (Handle, (float)percent, ic?.Handle ?? IntPtr.Zero);
```
mattleibow, line 83: Renamed the parameters.

binding/SkiaSharp.Skottie/Animation.cs
```diff
@@ -53,17 +79,20 @@ public unsafe void Render (SKCanvas canvas, SKRect dst)
...
+		public void SeekFrameTime (double seconds, InvalidationController? ic = null)
+			=> SkottieApi.skottie_animation_seek_frame_time (Handle, (float)seconds, ic?.Handle ?? IntPtr.Zero);
 
-		public double Duration
-			=> SkottieApi.skottie_animation_get_duration (Handle);
+		public void SeekFrameTime (TimeSpan time, InvalidationController? ic = null)
+			=> SeekFrameTime (time.TotalSeconds, ic);
+
+		public TimeSpan Duration
+			=> TimeSpan.FromSeconds(SkottieApi.skottie_animation_get_duration (Handle));
```
mattleibow, line 95: Added the new `SeekFrameTime(TimeSpan)` and changed `Duration` to be a `TimeSpan`.

source/SkiaSharp.Build.props
```diff
@@ -47,16 +47,16 @@
 
   <!-- .NET Standard, .NET FX and .NET 6 -->
   <PropertyGroup>
-    <BasicTargetFrameworks>netstandard1.3;netstandard2.0;net462</BasicTargetFrameworks>
+    <BasicTargetFrameworks>netstandard1.3;netstandard2.0;netstandard2.1;net462;netcoreapp3.1;net5.0</BasicTargetFrameworks>
+    <BasicTargetFrameworks Condition="$(BuildingForNet6)">$(BasicTargetFrameworks);net6.0</BasicTargetFrameworks>
```
mattleibow, line 51: Also building for netcoreapp3.1 and net5 by default, and on platforms that support net6, build net6.

This is not added to any NuGets, but I want to do that in a new PR so we get all the new code features.

source/SkiaSharp.Build.targets
```diff
@@ -54,6 +54,10 @@
     <PackageReference Include="mdoc" Version="5.8.9" PrivateAssets="All" GeneratePathProperty="true" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Nullable" Version="1.3.0" PrivateAssets="all" />
+  </ItemGroup>
```
mattleibow, line 59: Thanks OSS community! No need for me to copy attributes into my codebase!

PR #2149, opened 2022-07-07
Png chunk reader codec bitmap

mgood7123:
**Description of Change**

expose more api's

**Bugs Fixed**

- Fixes #

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKBitmap.cs
+++ b/binding/Binding/SKBitmap.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.IO;
+using System.Threading;
 
 namespace SkiaSharp
 {
@@ -37,8 +38,6 @@ namespace SkiaSharp
 	}
 
 	// TODO: keep in mind SKBitmap may be going away (according to Google)
-	// TODO: `ComputeIsOpaque` may be useful
-	// TODO: `GenerationID` may be useful
 	// TODO: `GetAddr` and `GetPixel` are confusing
 
 	public unsafe class SKBitmap : SKObject, ISKSkipObjectRegistration
@@ -114,6 +113,76 @@ namespace SkiaSharp
 
 		protected override void DisposeNative () =>
 			SkiaApi.sk_bitmap_destructor (Handle);
+
+		// Other
+
+		internal static SKBitmap GetObject(IntPtr handle, bool owns = true, bool unrefExisting = true) =>
+			GetOrAddObject(handle, owns, unrefExisting, (h, o) => new SKBitmap(h, o));
+
+		public bool SetInfo(SKImageInfo info)
+		{
+			// bool 	setInfo (const SkImageInfo &imageInfo, size_t rowBytes=0)
+			return SetInfo(info, 0);
+		}
+
+		public bool SetInfo(SKImageInfo info, int rowBytes)
+		{
+			var cinfo = SKImageInfoNative.FromManaged(ref info);
+			return SkiaApi.sk_bitmap_set_info(Handle, &cinfo, (IntPtr)rowBytes);
+		}
+
+		public bool ComputeIsOpaque()
+		{
+			return SkiaApi.sk_bitmap_compute_is_opaque(Handle);
+		}
+
+		// AllocPixels
+
+		public void AllocPixels()
+        {
+			AllocPixels((Allocator)null);
+		}
+
+		public void AllocPixels(Allocator allocator)
+		{
+			if (!TryAllocPixels(allocator))
+			{
+				SKImageInfo i = Info;
+				throw new OutOfMemoryException("SkBitmap::tryAllocPixels failed "
+					+ "ColorType:" + i.ColorType + "AlphaType:" + i.AlphaType +
+					"[w:" + i.Width + " h:" + i.Height + "] rb:" + RowBytes
+				);
+			}
+		}
+
+		public void AllocPixels(SKImageInfo info)
+		{
+			AllocPixels(info, info.RowBytes);
+		}
+
+		public void AllocPixels(SKImageInfo info, int rowBytes)
+		{
+			if (!TryAllocPixels(info, rowBytes))
+			{
+				SKImageInfo i = Info;
+				throw new OutOfMemoryException("SkBitmap::tryAllocPixels failed "
+					+ "ColorType:" + i.ColorType + "AlphaType:" + i.AlphaType +
+					"[w:" + i.Width + " h:" + i.Height + "] rb:" + RowBytes
+				);
+			}
+		}
+
+		public void AllocPixels(SKImageInfo info, SKBitmapAllocFlags flags)
+		{
+			if (!TryAllocPixels(info, flags))
+			{
+				SKImageInfo i = Info;
+				throw new OutOfMemoryException("SkBitmap::tryAllocPixels failed "
+					+ "ColorType:" + i.ColorType + "AlphaType:" + i.AlphaType +
+					"[w:" + i.Width + " h:" + i.Height + "] rb:" + RowBytes
+				);
+			}
+		}
 
 		// TryAllocPixels
 
@@ -317,6 +386,8 @@ namespace SkiaSharp
 			}
 		}
 
+		public SKImage AsImage () => SKImage.FromBitmap (this);
+
 		// properties
 
 		public bool ReadyToDraw => SkiaApi.sk_bitmap_ready_to_draw (Handle);
@@ -361,6 +432,10 @@ namespace SkiaSharp
 			get { return (int)SkiaApi.sk_bitmap_get_byte_count (Handle); }
 		}
 
+		public uint GenerationId {
+			get { return (uint)SkiaApi.sk_bitmap_get_generation_id(Handle); }
+		}
+
 		// *Pixels*
 
 		public IntPtr GetPixels () =>
@@ -400,6 +475,14 @@ namespace SkiaSharp
 		}
 
 		// more properties
+
+		public SKPixmap Pixmap {
+			get {
+				var pixmap = new SKPixmap (SkiaApi.sk_bitmap_get_pixmap (Handle), false); // pixmap is owned by this
+				pixmap.pixelSource = this;
+				return pixmap;
+			}
+		}
 
 		public byte[] Bytes {
 			get {
@@ -721,6 +804,27 @@ namespace SkiaSharp
 				: releaseProc;
 			var proxy = DelegateProxies.Create (del, DelegateProxies.SKBitmapReleaseDelegateProxy, out _, out var ctx);
 			return SkiaApi.sk_bitmap_install_pixels (Handle, &cinfo, (void*)pixels, (IntPtr)rowBytes, proxy, (void*)ctx);
+		}
+
+		public bool ReadPixels (SKImageInfo info, IntPtr dstpixels, int rowBytes, int x, int y)
+		{
+			if (GetPixels() == null) return false;
+			return Pixmap.ReadPixels(info, dstPixels, rowBytes, x, y);
+		}
+
+		public bool ReadPixels(SKPixmap dstPixmap) => ReadPixels(dstPixmap, 0, 0);
+
+		public bool ReadPixels(SKPixmap pixmap, int x, int y)
+		{
+			if (GetPixels() == null) return false;
+    		return Pixmap.ReadPixels(dst.Info, dst.Pixels, dst.RowBytes, srcX, srcY);
+		}
+
+		public bool WritePixels(SKPixmap pixmap) => WritePixels(pixmap, 0, 0);
+
+		public bool WritePixels(SKPixmap dstPixmap, int x, int y)
+		{
+			return SkiaApi.sk_bitmap_write_pixels_at_location(Handle, dstPixmap.Handle, x, y);
 		}
 
 		public bool InstallPixels (SKPixmap pixmap)
--- a/binding/Binding/SKCodec.cs
+++ b/binding/Binding/SKCodec.cs
@@ -4,7 +4,6 @@ using System.IO;
 
 namespace SkiaSharp
 {
-	// TODO: `Create(...)` should have overloads that accept a SKPngChunkReader
 	// TODO: missing the `QueryYuv8` and `GetYuv8Planes` members
 
 	public unsafe class SKCodec : SKObject, ISKSkipObjectRegistration
@@ -12,6 +11,25 @@ namespace SkiaSharp
 		internal SKCodec (IntPtr handle, bool owns)
 			: base (handle, owns)
 		{
+		}
+
+		/// <summary>
+		/// For container formats that contain both still images and image sequences,
+		/// <br></br> instruct the decoder how the output should be selected. (Refer to comments
+		/// <br></br> for each value for more details.)
+		/// </summary>
+		public enum SelectionPolicy
+		{
+			/// <summary>
+			/// If the container format contains both still images and image sequences,
+			/// <br></br> SKCodec should choose one of the still images. This is the default.
+			/// </summary>
+			preferStillImage,
+			/// <summary>
+			/// If the container format contains both still images and image sequences,
+			/// <br></br> SKCodec should choose one of the image sequences for animation.
+			/// </summary>
+			preferAnimation
 		}
 
 		protected override void Dispose (bool disposing) =>
@@ -41,6 +59,13 @@ namespace SkiaSharp
 
 		public SKEncodedImageFormat EncodedFormat =>
 			SkiaApi.sk_codec_get_encoded_format (Handle);
+
+		public SKSizeI Dimensions
+		{
+			get {
+				return Info.Size;
+			}
+		}
 
 		public SKSizeI GetScaledDimensions (float desiredScale)
 		{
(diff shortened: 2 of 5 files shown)


binding/Binding/SKBitmap.cs
```diff
@@ -401,6 +456,14 @@ public void SetColorTable (SKColorTable ct)
 
 		// more properties
 
+		public SKPixmap Pixmap {
+			get {
+				var pixmap = new SKPixmap (SkiaApi.sk_bitmap_get_pixmap (Handle), false); // pixmap is owned by this
+				pixmap.pixelSource = this;
+				return pixmap;
+			}
+		}
```
mattleibow, line 465: What happens if there is no pixels? Will the handle be null? Can you write a test for the cases where there are no pixels?

binding/Binding/SKBitmap.cs
```diff
@@ -723,6 +786,27 @@ public bool InstallPixels (SKImageInfo info, IntPtr pixels, int rowBytes, SKBitm
 			return SkiaApi.sk_bitmap_install_pixels (Handle, &cinfo, (void*)pixels, (IntPtr)rowBytes, proxy, (void*)ctx);
 		}
 
+		public bool ReadPixels (SKImageInfo info, IntPtr dstpixels, int rowBytes, int x, int y)
+		{
+			if (GetPixels() == null) return false;
```
mattleibow, line 791: Do we need this check? What does ReadPixels do when there are no pixels?

binding/Binding/SKPngChunkReader.cs
```diff
@@ -0,0 +1,159 @@
...
+		{
+			userData = DelegateProxies.CreateUserData(this, true);
+			Handle = SkiaApi.sk_managed_png_chunk_reader_new((void*)userData);
+
+			if (Handle == IntPtr.Zero)
+				throw new InvalidOperationException("Unable to create a new SKManagedAllocator instance.");
+		}
+
+		protected override void Dispose(bool disposing) =>
+			base.Dispose(disposing);
```
mattleibow, line 42: This is not needed. It exists in other places because I have a dumb API checker that thinks overrides are "breaking".

binding/Binding/SKPngChunkReader.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace SkiaSharp
+{
+	/// <summary>
+	/// Base class for optional callbacks to retrieve meta/chunk data out of a PNG
+	/// <br></br> encoded image as it is being decoded.
+	/// <br></br> Used by SkCodec.
+	/// </summary>
```
mattleibow, line 13: All docs unfortunately need to be done in the separate repo: https://github.com/mono/SkiaSharp-API-docs

However, this repo is only updated AFTER we merge the PR here... so yeah,  🐔 ♻️🐣 ♻️ 🍗 ♻️ 🍳

binding/Binding/SKPngChunkReader.cs
```diff
@@ -0,0 +1,159 @@
...
+		/// <br></br> @param data Data to be interpreted by the subclass.
+		/// <br></br> @param length Number of bytes of data in the chunk.
+		/// <br></br> @return true to continue decoding, or false to indicate an error, which
+		/// <br></br>     will cause the decoder to not return the image.
+		/// </summary>
+		/// <param name="tag">Name for this type of chunk.</param>
+		/// <param name="data">Data to be interpreted by the subclass.</param>
+		/// <param name="length">Number of bytes of data in the chunk.</param>
+		/// <returns>true to continue decoding, or false to indicate an error, which will cause the decoder to not return the image.</returns>
+		public abstract bool ReadChunk(string tag, void* data, IntPtr length);
```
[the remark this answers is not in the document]
mattleibow: If tag is always a string, then that should be public and the void* one should remain private. If it can be some random data, then we should have both overloads. If the tag is 90% string, then we can make it abstract and the IntPtr version virtual so people that want/need the IntPtr version can overide that and throw in this one.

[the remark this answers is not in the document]
mattleibow: ReadChunk appears to be protected in C++, so we should do that as well - for consistency as well as to prevent random callers using this method.

mattleibow (commented):
Very nice!

All we need are some API tweaks and then some tests! The main areas for tests are around the pixmap values depending on whether bitmaps have pixel data or not. Also we need tests for various permutations of the chunk reader with some png files with actual tags in it.

binding/Binding/SKCodec.cs
```diff
@@ -14,6 +13,25 @@ internal SKCodec (IntPtr handle, bool owns)
 		{
 		}
 
+		/// <summary>
+		/// For container formats that contain both still images and image sequences,
+		/// <br></br> instruct the decoder how the output should be selected. (Refer to comments
+		/// <br></br> for each value for more details.)
+		/// </summary>
```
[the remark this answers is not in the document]
mgood7123: alright

binding/Binding/SKBitmap.cs
mgood7123: if there are no pixels then i assume pixmap will be empty

binding/Binding/SKBitmap.cs
mgood7123: it avoids allocating a SKPixmap object if there are no pixels

binding/Binding/SKPngChunkReader.cs
mgood7123: alright

native/wasm/build.cake
```diff
@@ -52,6 +52,7 @@ Task("libSkiaSharp")
         $"extra_cflags=[ " +
         $"  '-DSKIA_C_DLL', '-DXML_POOR_ENTROPY', " + 
         $"  '-DSKNX_NO_SIMD', '-DSK_DISABLE_AAA', '-DGR_GL_CHECK_ALLOC_WITH_GET_ERROR=0', " +
+        $"  '-DPNG_READ_UNKNOWN_CHUNKS_SUPPORTED', " +
```
mattleibow, line 55: This define is required to enable chunk reading.

tests/Tests/SKCodecTest.cs
```diff
@@ -491,5 +493,31 @@ public void CanDecodeData(string image)
...
+			var path = Path.Combine(PathToImages, "png_chunks/good_itxt.png");
+
+			using var chunkReader = new TestPngChunkReader();
+			using var codec = SKCodec.Create(path, chunkReader, SKCodecSelectionPolicy.PreferStillImage, out var codecResult);
+			using var bmp = SKBitmap.Decode(codec);
+
+			var chunks = chunkReader.Chunks;
+
+			Assert.NotEmpty(chunks);
+		}
```
mattleibow, line 509: This test is failing because chunks are not being read still.

mattleibow (commented):
@mgood7123  I see chunking wasn't working in 2 ways, so I opened this discussion: https://groups.google.com/g/skia-discuss/c/zHVOjwoQ-jQ

mattleibow:
PNG chunk images are from https://www.nayuki.io/page/png-file-chunk-inspector

tests/Tests/SKCodecTest.cs
mgood7123: could you try the following?

```cs
internal class Reader : SKPngChunkReader
{
    public bool r = false;

    protected override bool ReadChunk(string tag, IntPtr data, IntPtr length)
    {
        r = true;
        return true;
    }
}
public void CanInvokePngReadChunks(
{
    using Reader chunkReader = new();

    var path = Path.Combine(PathToImages, "png_chunks/good_itxt.png");

    using var codec = SKCodec.Create(path, chunkReader, SKCodecSelectionPolicy.PreferStillImage, out var codecResult);

    Assert.True(r.r, "failed to invoke chunk reader");
}
```

with `SkiaSharp\externals\skia\BUILD.gn`

```
optional("png_decode") {
  enabled = skia_use_libpng_decode
  public_defines = [ "SK_CODEC_DECODES_PNG", "PNG_READ_UNKNOWN_CHUNKS_SUPPORTED" ]

  deps = [ "//third_party/libpng" ]
  sources = [
    "src/codec/SkIcoCodec.cpp",
    "src/codec/SkPngCodec.cpp",
  ]
}
```

mattleibow: I already set the flag here https://github.com/mono/SkiaSharp/pull/2149#discussion_r946167028

I can actually debug and see it reading the chunk but it seems to decide it is not worth it and skips.

mgood7123: oof

mgood7123: https://groups.google.com/g/skia-discuss/c/zHVOjwoQ-jQ/m/aKQ-EIZ_AQAJ

it looks like we will need to test it with UNKNOWN chunks instead of KNOWN chunks

mgood7123: coincidentally this `[an example implementation here](https://android.googlesource.com/platform/frameworks/base/+/master/libs/hwui/jni/NinePatchPeeker.cpp).` is exactly what i have ported https://github.com/mgood7123/AndroidUI/blob/master/AndroidUI/Graphics/NinePatchPeeker.cs

mgood7123: but i currently dont have any NinePatch images to test with, however the NinePatch tests all pass https://github.com/mgood7123/android_9_patch/blob/31931f9910f428207ee482a808cc09195a985c87/9patch_tests.cpp

mgood7123: and https://github.com/mgood7123/AndroidUI/blob/f7746e01b17bd728f94937da5d96c34c5d6ea024/AndroidUI/Graphics/BitmapFactory.cs#L1254

mattleibow: I wonder if we can borrow some images from android in the GitHub.com/aosp_mirror

mgood7123:
> Based on the google discussion:
> 
> > I think this only supports unknown chunks - iTXt and pHYs are standard chunks, which are handled directly by libpng.
> > And
> 
> > The Skia API doesn't give you a way to get those chunks (or their data), though, and I'm not sure it should.
> 
> I am not sure this adds value since there is no way to read those values in the PNG.

i think they meant skia does not give you a way to get the standard chunks (nor does libpng unless you manually decode by yourself)

mattleibow:
I added a test, so before we merge I think we should add a test proving this works. Are you able to find an image that causes the chunks to be read?

I can't merge unless we know for sure this is working.

mgood7123:
> I added a test, so before we merge I think we should add a test proving this works. Are you able to find an image that causes the chunks to be read?
> 
> I can't merge unless we know for sure this is working.

ill see what i can find

mgood7123:
i tried it with
```cs
var bm = BitmapFactory.decodeFile(Context, image_path_NPatch);
var chunks = bm.getNinePatchChunk();
if (NinePatch.isNinePatchChunk(chunks))
{
    Log.d("NINEPATCH", "IS CHUNK YES");
    //image.setImageDrawable(new NinePatchDrawable(bm, chunks, new Rect(), null));
} else
{
    Log.d("NINEPATCH", "IS CHUNK NO");
}
```
but i get `IS CHUNK NO`

additionally i added

```cs
        protected override bool ReadChunk(string tag, IntPtr data, IntPtr length)
        {
            // NPatch
            System.Console.WriteLine("READ NINEPATCH CHUNK");
```

but i dont see it anywhere

mgood7123:
circle 9

mgood7123:
in android this works
```java
Bitmap bm = decodeResource(getResources(), R.drawable.circle); // same as DecodeFile
byte[] ninePatchChunk = bm.getNinePatchChunk();
if (NinePatch.isNinePatchChunk(ninePatchChunk)) {
    image.setImageDrawable(new NinePatchDrawable(bm, ninePatchChunk, new Rect(), null));
```

mgood7123:
ok i build a debug native skia lib :)

mgood7123:
hmmm it seems that Android's AAPT creates the 9-patch chunks itself, and the image does not actually contain these chunks (just the pixel info for these chunks)

mgood7123:
```java
// Bitmap bm = decodeResource(getResources(), R.drawable.circle); AAPT packaged
Bitmap bm = decodeFile("/data/local/tmp/circle.9.png"); // RAW FILE pushed via adb
byte[] ninePatchChunk = bm.getNinePatchChunk();
```

mgood7123:
so we would need to find such a file with these chunks already present

mgood7123:
and it can read it :)

image


```cs
internal class _10_test_png_chunk_reader : Test
{
    internal class Reader : SKPngChunkReader
    {
        public bool r = false;

        protected override bool ReadChunk(string tag, IntPtr data, IntPtr length)
        {
            r = true;
            Console.WriteLine("READ CHUNK");
            return true;
        }
    }
    public override void Run(TestGroup nullableInstance)
    {
        using Reader r = new();
        using FileStream stream = new("C:\\Users\\AndroidUI\\Desktop\\AndroidUI\\AndroidUITest\\png_chunks\\good_itxt.png", FileMode.Open);
        SKCodecResult result_;
        using SKCodec c = SKCodec.Create(stream, out result_, r);
        Tools.ExpectFalse(r.r, "known chunks are not meant to invoke chunk reader");

        using Reader r2 = new();
        using FileStream np = new("K:/Images/9PNG/circle.9.png", FileMode.Open);
        SKCodecResult result2_;
        using SKCodec c2 = SKCodec.Create(np, out result2_, r2);
        Tools.ExpectTrue(r2.r, "unknown chunks are meant to invoke chunk reader");
    }
}
```

mgood7123:
image

mgood7123:
so now we can read a 9patch image

circle 9

mgood7123:
also plugging this file into the png chunk inspector gives

image

mattleibow:
Can you add this as a test?

mgood7123:
this should work now :)

mgood7123:
> Based on the google discussion:
> 
> > I think this only supports unknown chunks - iTXt and pHYs are standard chunks, which are handled directly by libpng.
> > And
> 
> > The Skia API doesn't give you a way to get those chunks (or their data), though, and I'm not sure it should.
> 
> I am not sure this adds value since there is no way to read those values in the PNG.

it seems so

basically...
```c
   chunk_name = png_ptr->chunk_name;

   if (chunk_name == png_IDAT)
   {
        // ...
   }

   if (chunk_name == png_IHDR)
   {
        // ...
   }

   else if (chunk_name == png_IEND)
   {
        // ...
   }

#ifdef PNG_HANDLE_AS_UNKNOWN_SUPPORTED
   else if ((keep = png_chunk_unknown_handling(png_ptr, chunk_name)) != 0)
   {
      PNG_PUSH_SAVE_BUFFER_IF_FULL
      png_handle_unknown(png_ptr, info_ptr, png_ptr->push_length, keep);

      if (chunk_name == png_PLTE)
         png_ptr->mode |= PNG_HAVE_PLTE;
   }
#endif

   else if (chunk_name == png_PLTE)
   {
        // ...
   }

   else if (chunk_name == png_IDAT)
   {
        // ...
      return;
   }

#ifdef PNG_READ_gAMA_SUPPORTED
   else if (png_ptr->chunk_name == png_gAMA)
   {
        // ...
   }

#endif
#ifdef PNG_READ_sBIT_SUPPORTED
   else if (png_ptr->chunk_name == png_sBIT)
   {
        // ...
   }

#endif
#ifdef PNG_READ_cHRM_SUPPORTED
   else if (png_ptr->chunk_name == png_cHRM)
   {
        // ...
   }

#endif
#ifdef PNG_READ_sRGB_SUPPORTED
   else if (chunk_name == png_sRGB)
   {
        // ...
   }

#endif
#ifdef PNG_READ_iCCP_SUPPORTED
   else if (png_ptr->chunk_name == png_iCCP)
   {
        // ...
   }

#endif
#ifdef PNG_READ_sPLT_SUPPORTED
   else if (chunk_name == png_sPLT)
   {
        // ...
   }

#endif
#ifdef PNG_READ_tRNS_SUPPORTED
   else if (chunk_name == png_tRNS)
   {
        // ...
   }

#endif
#ifdef PNG_READ_bKGD_SUPPORTED
   else if (chunk_name == png_bKGD)
   {
        // ...
   }

#endif
#ifdef PNG_READ_hIST_SUPPORTED
   else if (chunk_name == png_hIST)
   {
        // ...
   }

#endif
#ifdef PNG_READ_pHYs_SUPPORTED
   else if (chunk_name == png_pHYs)
   {
        // ...
   }

#endif
#ifdef PNG_READ_oFFs_SUPPORTED
   else if (chunk_name == png_oFFs)
   {
        // ...
   }
#endif

#ifdef PNG_READ_pCAL_SUPPORTED
   else if (chunk_name == png_pCAL)
   {
        // ...
   }

#endif
#ifdef PNG_READ_sCAL_SUPPORTED
   else if (chunk_name == png_sCAL)
   {
        // ...
   }

#endif
#ifdef PNG_READ_tIME_SUPPORTED
   else if (chunk_name == png_tIME)
   {
        // ...
   }

#endif
#ifdef PNG_READ_tEXt_SUPPORTED
   else if (chunk_name == png_tEXt)
   {
        // ...
   }

#endif
#ifdef PNG_READ_zTXt_SUPPORTED
   else if (chunk_name == png_zTXt)
   {
        // ...
   }

#endif
#ifdef PNG_READ_iTXt_SUPPORTED
   else if (chunk_name == png_iTXt)
   {
        // ...
   }
#endif

   else
   {
      PNG_PUSH_SAVE_BUFFER_IF_FULL
      png_handle_unknown(png_ptr, info_ptr, png_ptr->push_length,
          PNG_HANDLE_CHUNK_AS_DEFAULT);
   }

   png_ptr->mode &= ~PNG_HAVE_CHUNK_HEADER;
```

PR #2153, opened 2022-07-07
expose more of SKMatrix

mgood7123:
**Description of Change**

**Bugs Fixed**

- Fixes #

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/SkiaSharp/pull/2153

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKMatrix.cs
+++ b/binding/Binding/SKMatrix.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.CompilerServices; // inlining
 using System.ComponentModel;
 
 namespace SkiaSharp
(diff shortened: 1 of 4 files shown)


binding/Binding/SKMatrix.cs
```diff
@@ -444,6 +535,78 @@ public static void PostConcat (ref SKMatrix target, ref SKMatrix matrix)
...
+			setFrom(ref o);
+		}
+
+
+		public void PreTranslate(float dx, float dy)
+		{
+			SKMatrix o;
+			fixed (SKMatrix* t = &this)
+			{
+				SkiaApi.sk_matrix_post_translate(&o, t, dx, dy);
```
sebllll, line 585: shouldn't it read `sk_matrix_pre_translate()` here?

mgood7123: Not sure, havnt worked on it in ages

mattleibow:
Thanks for this PR! However, I am not sure the current `SKMatrix` approach is the way we want to go for 3.0.

I have been investigating this from a perf perspective and right now `SKMatrix` has terrible performance compared to `SKMatrix44`. It is about 200% slower because it has to hop the interop, and it also is deceiving because using a struct is supposed to be light, however it does expensive things. 

 - See https://github.com/mono/SkiaSharp/issues/2779

It may even be cheaper to use `Matrix4x4` under the hood. In addition to the current interop, the managed struct does not have the matrix type mask which skia uses to take shortcuts. As a result, even just passing the struct between C++ and C# causes skia to have to recalculate it. It may be better to just re-implement the whole thing in C#.

mgood7123:
alright

PR #2157, opened 2022-07-07
add SKNativeObject allocation logging

mgood7123:
**Description of Change**

can be toggled on/off via `SKNativeObject.LOG_ALLOCATIONS` (default is false)

**Bugs Fixed**

None.

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/SkiaSharp/pull/2157

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKObject.cs
+++ b/binding/Binding/SKObject.cs
@@ -229,9 +229,29 @@ namespace SkiaSharp
 
 		// stack trace
 		private string s;
-		private static readonly object L = new();
 
 		public static bool LOG_ALLOCATIONS = false;
+		private static Action<string> c1;
+		private static Action<string> c2;
+		private static Action<string> c3;
+		private static Action<string> c4;
+
+		public static Action<string> LOG_ALLOCATION_CONSTRUCTOR_ENTER_CALLBACK {
+			get { return c1; }
+			set { c1 = value; }
+		}
+		public static Action<string> LOG_ALLOCATION_CONSTRUCTOR_EXIT_CALLBACK {
+			get { return c2; }
+			set { c2 = value; }
+		}
+		public static Action<string> LOG_ALLOCATION_DESTRUCTOR_ENTER_CALLBACK {
+			get { return c3; }
+			set { c3 = value; }
+		}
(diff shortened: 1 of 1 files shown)


mattleibow (changes requested):
Thanks for this PR. We should add some logging, but this must be a Debug-only feature, internal and opt-in by setting some property.

If you have a look at the way I set up `THROW_OBJECT_EXCEPTIONS`. It is a define that was added in debug builds: 
 - https://github.com/mono/SkiaSharp/blob/cb424e5230471ed300319e97cc7cb4fd9f771e03/source/SkiaSharp.Build.props#L88
 - https://github.com/mono/SkiaSharp/blob/e603133c3548bbd3835cf1012eac80ce5640f209/binding/Binding/HandleDictionary.cs#L16

The reason for internal is to avoid any accidental APIS leaking, and then when it is in the conditions, we can do checks with and without the logging - maybe for perf or to determine things are working correctly for tests.

mgood7123:
> Thanks for this PR. We should add some logging, but this must be a Debug-only feature, internal and opt-in by setting some property.
> 
> If you have a look at the way I set up `THROW_OBJECT_EXCEPTIONS`. It is a define that was added in debug builds:
> 
>     * https://github.com/mono/SkiaSharp/blob/cb424e5230471ed300319e97cc7cb4fd9f771e03/source/SkiaSharp.Build.props#L88
> 
>     * https://github.com/mono/SkiaSharp/blob/e603133c3548bbd3835cf1012eac80ce5640f209/binding/Binding/HandleDictionary.cs#L16
> 
> 
> The reason for internal is to avoid any accidental APIS leaking, and then when it is in the conditions, we can do checks with and without the logging - maybe for perf or to determine things are working correctly for tests.

hmmmm i still like that it is toggle-able and public, as it may help end users track down allocations that they forget to dispose (which will add memory pressure until GC occurs)

making it togglable vis building skiasharp itself (and via debug build) would likely be combersome since the end user will likely be debug building their app and expect skiasharp to be a release build

maybe we could offer both a debug nuget and a release nuget?

PR #2167, opened 2022-07-14
Support loading animations with BOM

### LINKED ISSUE #2166: [BUG] Skottie cannot load animations with a BOM
**Description**

If you try load an animation from a stream/file that has a BOM, the loader just returns null.

**Code**

```cs
var stream = await FileSystem.OpenAppPackageFileAsync("Lottie/trophy.json");

var animation = SkiaSharp.Skottie.Animation.Create(stream);
```

**Expected Behavior**

Animation laods.

**Actual Behavior**

Null result.

**Workarounds**

If the data is a stream, we can use the stream reader to remove the preamble:

```cs
var reader = new StreamReader(stream);
var len = reader.CurrentEncoding.Preamble.Length;
var pre = new byte[len];
stream.Read(pre, 0, len);
```

> It is more complicated the just setting the `Position = len` because the stream may be non-seekable or part of a larger stream.

mattleibow:
**Description of Change**

The JSON parser inside skia does not support BOMs, so we have to do it. 

**Bugs Fixed**

- Fixes #2166

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


binding/Binding/SKData.cs
```diff
@@ -103,7 +103,7 @@ public static SKData Create (Stream stream)
 			if (stream == null)
 				throw new ArgumentNullException (nameof (stream));
 			if (stream.CanSeek) {
-				return Create (stream, stream.Length);
+				return Create (stream, stream.Length - stream.Position);
```
mattleibow, line 106: This was a bug I found when you pass a stream that was not at 0. This can be used to create a data object from a section of a stream.

binding/Binding/Util.cs
```diff
@@ -12,6 +12,25 @@ internal unsafe static class Utils
...
+			var len = buffer.Length;
+
+			if (len >= 2 && buffer[0] == 0xfe && buffer[1] == 0xff)
+				return 2;
+			else if (len >= 3 && buffer[0] == 0xef && buffer[1] == 0xbb && buffer[2] == 0xbf)
+				return 3;
+			else if (len >= 3 && buffer[0] == 0x2b && buffer[1] == 0x2f && buffer[2] == 0x76)
+				return 3;
+			else if (len >= 4 && buffer[0] == 0 && buffer[1] == 0 && buffer[2] == 0xfe && buffer[3] == 0xff)
+				return 4;
```
mattleibow, line 29: BOM detection. Right now just the basic UTF-7/8/16/32

binding/SkiaSharp.Skottie/Animation.cs
```diff
@@ -56,8 +60,8 @@ public static bool TryCreate (SKStream stream, [System.Diagnostics.CodeAnalysis.
 		{
 			_ = stream ?? throw new ArgumentNullException (nameof (stream));
 
-			animation = GetObject (SkottieApi.skottie_animation_make_from_stream (stream.Handle));
-			return animation != null;
+			using var data = SKData.Create (stream);
+			return TryCreate (data, out animation);
```
mattleibow, line 64: All animations are now created through the SKData overload, which we then use to detect the BOM.

We _could_ use the stream, but then we have to use buffering and all other manner of complex mechanisms. This way we load the full data, and then optionally skip 4 bytes before passing it to C++.

PR #2254, opened 2022-09-15
Add .NET 7 Support for Blazor views

mattleibow:
**Description of Change**

Make sure to use the new build of the wasm binary for .NET 7 Blazor Apps.

**TODO**

 - [ ] Make use of SIMD flags


bricelam (approved):

danwalmsley:
@mattleibow just a heads up... that Blazor skiasharp views are broken under dotnet 7.

https://github.com/mono/SkiaSharp/blob/main/source/SkiaSharp.Views.Blazor/SkiaSharp.Views.Blazor/wwwroot/SKHtmlCanvas.ts#L114

This GL symbol is not in scope.

Dont quote me on this, but my conclusion from spending a day investigating is:

net 6, ES6 modules are not used: GL is a global variable.

net7 dotnet.js is using ES6 modules, and GL is now private and isnt exported.

something generates the js code for the GL stuff and stuffs it in (dotnet.js). I am unable to determine the thing that generates this source thus far.

for now ill continue to find a workaround, but maybe you know something ?

danwalmsley:
@mattleibow added an issue to runtime repo:

https://github.com/dotnet/runtime/issues/76077

PR #2387, opened 2023-02-15
[ci] Add PoliCheckExclusions.xml

pjcollins:
Adds an exclusions file for policheck to allow it to ignore the sources
in the `third_party` folder that we do not own.


pjcollins:
@mattleibow This should resolve ~1776 work items, though I am unable to test it.

pjcollins:
Closing this to test via https://github.com/mono/SkiaSharp/pull/2389

PR #2389, opened 2023-02-16
[ci] Add exclusions for PoliCheck and CredScan

pjcollins:
Adds a exclusion files for CredScan and PoliCheck to ignore third party
sources and a mathematical term used as a variable name in number of
places. The versions of the compliance and reporting tasks used have
also been updated.

The new `source.gdnsuppress` file is used to exclude specific terms from
PoliCheck, as the `PoliCheckExclusions.xml` file does not allow very
granular control. This file is generated and uploaded by the build, and
the generated file can be used to update the in-tree file to exclude
more terms in the future as needed.


pjcollins:
This cuts the issue set down dramatically (from `PoliCheck failed with 2617 issues found` to `PoliCheck failed with 112 issues found`), but there are still some components we will want to try to address.

pjcollins:
The new versions of the tools have uncovered some more issues, but we've gone from 2617 down to 169 in the [latest run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7345539&view=logs&j=20168050-bfe6-5e92-c15c-a5eae27a73d9&t=0200cdd6-bc1a-5e6a-3793-2f798d73fdd9&l=91) and now have a better starting point to resolve the rest. This should be ready to go.

Redth (approved):

PR #2393, opened 2023-02-17
[ci] Use default feed to install the .NET SDK

pjcollins:
Attempts to install .NET 6.0.403 have been failing on various jobs:

    Error: Unable to download https://dotnetcli.blob.core.windows.net/public/Sdk/6.0.403/dotnet-dev-win-x64.6.0.403.zip. Returned HTTP status code: 404
    Could not find ".NET Core SDK" with version = 6.0.403

Fix this by preferring the default feed found in the dotnet-install.ps1 script.

### DIFF
--- a/scripts/install-dotnet.ps1
+++ b/scripts/install-dotnet.ps1
@@ -1,7 +1,7 @@
 Param(
   [string] $Version,
   [string] $InstallDir,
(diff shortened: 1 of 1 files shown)


pjcollins:
@mattleibow it looks like this fixes most of the failing jobs, but can not be auto-merged due to another issue with expired Apple certs.

PR #2398, opened 2023-02-28
[Uno] Default Opaque to false for SKXamlCanvas on MacOS and iOS

roubachof:
**Description of Change**

See canvas default opacity to none on mac and iOS to align on other platforms.

**Bugs Fixed**

**API Changes**

None.

**Behavioral Changes**

By default, canvas will be transparent instead of black.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Mac/SKXamlCanvas.macOS.cs
+++ b/source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Mac/SKXamlCanvas.macOS.cs
@@ -20,6 +20,9 @@ namespace SkiaSharp.Views.UWP
 		public SKXamlCanvas()
 		{
 			Initialize();
+#if MACOS
+			Opaque = false;
+#endif
 		}
 
 		partial void DoLoaded() =>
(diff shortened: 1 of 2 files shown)


MartinZikmund (approved):

mattleibow (approved):

source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.Mac/SKXamlCanvas.macOS.cs
```diff
@@ -20,6 +20,9 @@ public partial class SKXamlCanvas
 		public SKXamlCanvas()
 		{
 			Initialize();
+#if MACOS
+			Opaque = false;
+#endif
```
mattleibow, line 25: This is actually incorrect for macOS. There is no `Opaque` property, but rather `IsOpaque`. However, this other property is read only and according to the docs you should override the property to return the value you want:

> The default value of this property is false to reflect the fact that views do no drawing by default. Subclasses can override this property and return true to indicate that the view completely covers its frame rectangle with opaque content. Doing so can improve performance during drawing operations by eliminating the need to render content behind the view.
> https://developer.apple.com/documentation/appkit/nsview/1483558-isopaque

mattleibow: Fixed in https://github.com/mono/SkiaSharp/pull/2477

roubachof: oops sorry about that ! I thought I tested it :/

PR #2443, opened 2023-04-18
Update bitmap SKAlphaType in SKXamlCanvas.Wasm.cs

roubachof:
To obtain the expected built bitmap by skia, we need to change the SKAlphaType of the build image. Opaque doesn't give correct results when bluring none opaque colors for example. Premul doesn't work either for a particular reason: In javascript we use the ImageData API.
The ImageData API store the pixels as un-premultiplied. So if we use the premultiplied enum value we end up with wrong pixel colors (same result as Opaque in fact). I tested unpremul value in a project that is using blur, and it gives the correct results.


mattleibow:
@roubachof is there a way to create a test? Right now we have no way to capture the image, but some drawing code that I can run against this PR to test?

Is it just a batter of drawing two different blocks, or is it blending with other elements?

roubachof:
When this PR will be merged (very soon I hope :) You will be able to test it  with the WASM target of the Uno toolkit sample => Control => ShadowContainer
You should see this : 
shadows-neumorphism
If it's in premul, the shadows will look "dirty"

PR #2476, opened 2023-06-06
Switch to Debian archives

mattleibow:
**Description of Change**

Update the Docker files to use the now-archived Debian 9 sources.


mattleibow:
Merging this even though the build is failing. This PR does fix the Linux build issues. However, CI is still not green and that is in progress over at https://github.com/mono/SkiaSharp/pull/2425

PR #2505, opened 2023-06-22
Modernize the repository and prepare SkiaSharp 3.0

mattleibow:
**Description of Change**

This PR is a massive jump for SkiaSharp. Not sure of the full scope just yet, but SkiaSharp is far too huge to be maintainable as well as it holds onto the past. Some platforms are no longer relevant/supported and the tooling is ancient.

The changes so far:
 - Dropping support for old platforms:
    - watchOS as this is not a platform that is useful to have SkiaSharp on - SkiaSharp is larger than the allowed app size
    - UWP as this platform is not so active - we can revisit this later and it appears Win32 dlls work in UWP now
 - Dropping support for old frameworks:
    - Classic Xamarin (the non .NET version) as this is going out of support in May 2024
    - Classic Tizen (the non .NET version)
    - Xamarin.Forms as this is no longer actively developed and requires Classic Xamarin
    - GTK 2 as this is super ancient
    - netcoreapp3.1 - moving to net6 and net7
 - Switching the build system to .NET 7 since this is far superior
 - Switching to the new xharness and device tests


dbriard:
Hi @mattleibow 
Could you have a look at this https://github.com/mono/SkiaSharp/issues/2530 for 3.0 please? (or an earlier preview builds)

ahmed605:
> UWP as this platform is not supported anymore



https://learn.microsoft.com/en-us/windows/apps/get-started/windows-developer-faq#deprecations-and-migrations

HEIC-to-JPEG-Dev:
> > UWP as this platform is not supported anymore
> 
> 
> 
> https://learn.microsoft.com/en-us/windows/apps/get-started/windows-developer-faq#deprecations-and-migrations

I think that's the actual definition of "deprecated" - nothing new, they'll keep the lights on

ahmed605:
> > > UWP as this platform is not supported anymore
> > 
> > 
> > 
> > https://learn.microsoft.com/en-us/windows/apps/get-started/windows-developer-faq#deprecations-and-migrations
> 
> I think that's the actual definition of "deprecated" - nothing new, they'll keep the lights on

my point is that it's still fully supported, even .NET Native is still fully supported and in fact there's still a dedicated support team for it

HEIC-to-JPEG-Dev:
> > > > UWP as this platform is not supported anymore
> > > 
> > > 
> > > 
> > > https://learn.microsoft.com/en-us/windows/apps/get-started/windows-developer-faq#deprecations-and-migrations
> > 
> > 
> > I think that's the actual definition of "deprecated" - nothing new, they'll keep the lights on
> 
> my point is that it's still fully supported, even .NET Native is still fully supported and in fact there's still a dedicated support team for it

My analogy is, if you're buying a car and they said it would only get services and fixes but no new upgrades, and there are no new models - you'd buy a different car.

Gillibald:
Dropping platforms makes it possible to move SkiaSharp further. It is desperately needed to modernize this project. Breaking changes are also welcome to increase maintainability.

ahmed605:
> Dropping platforms makes it possible to move SkiaSharp further. It is desperately needed to modernize this project. Breaking changes are also welcome to increase maintainability.

What's so hard about adding `uap` to `TargetFrameworks` tag though? especially that most SkiaSharp projects already support .NET Standard 2.0 (`uap` target has a similar BCL as the `netstandard2.0` target so there should be no compatibility issues) and Uno Platform (the API is (superficially) the exact same as UWP's) so it's (almost?) no work

michael-hawker:
I'm also worried about dropping UWP as a platform as the WASDK isn't a complete replacement for it yet. We're still planning on supporting UWP in the Toolkit. One of our components (Rive) uses ShiaSharp, so this would prevent us from updating to the latest version or creating other components using SkiaSharp.

mattleibow:
@ahmed605

> What's so hard about adding uap to TargetFrameworks tag though? 

The old way of building required a mac and windows build of separate project and then combining the artifacts in a separate step using the old nuget pack.

The new way can build and pack in a single project.

The main issue is that dotnet core does not support uwp. However, no uwp support in this PR does not mean that we can't get it back. As far as testing goes, we no longer need a special uwp build as the old win32 dlls now work and so does netstandard.

The only thing that could be counted as missing is the uwp views which I can either add back or put is a separate repository. This PR is trying to improve maintainability. 

However, this PR is still a ways away from a final release. Once I merge and then clean up the api and do the skia update, we can see the demand for uwp and we can add back. All the code is just a commit away.

mattleibow:
> I'm also worried about dropping UWP as a platform as the WASDK isn't a complete replacement for it yet. 

This is a valid case and maybe we need to add uwp back. However, I am hoping to get a first public preview in time for net8 and maybe stable for build. So this PR is more a remove all the bits that potentially are old and will cause issues. Then we are going to clean up all the obsolete apis and then update native skia.

Once we are new and clean, we can see if we still have the right tfm selection.

mattleibow:
Thanks folks for the comments. All the talk around UWP is good and shows there is still some need for that.

mattleibow:
Here is an example of the pain when building the repo. You basically have to check all the boxes in VS and then install some old components.

- https://github.com/mono/SkiaSharp/issues/2534

Once we merge, all that is required is a dotnet build/pack and the IDE will also just work.

Not only is this PR trying to improve my dev turnaround but also make this project something that at least some other collection of people can build.

PR #2664, opened 2023-11-03
Provide allocate-less overloads of methods in SKTextBlobBuilder

Youssef1313:
In Uno Platform, we are using `AllocatePositionedRun`, which in some scenarios can be problematic as it's allocating and it's in a hot path. See for example this is a case where this was responsible for 11.3% of all allocations.

image

The idea here is to introduce a new API that won't allocate `SKPositionedRunBuffer` and instead give us the `Span`s that we need right away. I have made the same change to APIs other than `AllocatePositionedRun` for parity.

- Fixes https://github.com/mono/SkiaSharp/issues/2776

### DIFF
--- a/binding/SkiaSharp/SKTextBlob.cs
+++ b/binding/SkiaSharp/SKTextBlob.cs
@@ -61,9 +61,9 @@ namespace SkiaSharp
 				return null;
 
 			using var builder = new SKTextBlobBuilder ();
-			var buffer = builder.AllocatePositionedRun (font, count);
-			font.GetGlyphs (text, length, encoding, buffer.GetGlyphSpan ());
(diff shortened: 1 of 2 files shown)


Gillibald:
Why aren't you changing SKRunBuffer to a readonly struct?

Youssef1313:
@Gillibald Avoiding breaking changes

Gillibald:
This can be a breaking change for V3

mattleibow:
This sort of came to me https://github.com/mono/SkiaSharp/pull/2775

Let me know if this is useful and leave comments/suggestions.

mattleibow:
Closing this in favour of https://github.com/mono/SkiaSharp/pull/2775

Since this API will be 3.0+ and the whole way the build works is different, the current API was mostly wrong. Hopefully the 3.x APIs are better to use.

PR #2697, opened 2024-01-08
Fix Android NDK extraction

mattleibow:
**Description of Change**

No idea why, but adding `-verbose` makes the image attach, at least in my CI testing...


mattleibow:
Seems to no longer work. Very sad.

PR #2756, opened 2024-02-15
[release/2.x] Backport some new APIs from 3.0 to 2.x allowing smoother migration.

maxkatz6:
**Description of Change**

With these changes Avalonia should be able to work with both 2.x and 3.0 referenced (need to double check once nightly builds are there).

**API Changes**

Added: 
 
- SKCanvas.SetMatrix(in SKMatrix);
- SKPath.Transform(in SKMatrix, SKPath destination);
- SKImageFilter methods without SKImageFilter and SKImageFilter.CropRect overloads

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/Binding/SKCanvas.cs
+++ b/binding/Binding/SKCanvas.cs
@@ -952,13 +952,6 @@ namespace SkiaSharp
 			SkiaApi.sk_canvas_set_matrix (Handle, &matrix);
 		}
 
-		public void SetMatrix (in SKMatrix matrix)
-		{
-			fixed (SKMatrix* ptr = &matrix) {
-				SkiaApi.sk_canvas_set_matrix (Handle, ptr);
-			}
-		}
-
 		public SKMatrix TotalMatrix {
 			get {
 				SKMatrix matrix;
--- a/binding/Binding/SKCanvas.cs
+++ b/binding/Binding/SKCanvas.cs
@@ -952,6 +952,13 @@ namespace SkiaSharp
 			SkiaApi.sk_canvas_set_matrix (Handle, &matrix);
 		}
 
+		public void SetMatrix (in SKMatrix matrix)
+		{
+			fixed (SKMatrix* ptr = &matrix) {
+				SkiaApi.sk_canvas_set_matrix (Handle, ptr);
+			}
+		}
+
 		public SKMatrix TotalMatrix {
 			get {
 				SKMatrix matrix;
--- a/binding/Binding/SKImageFilter.cs
+++ b/binding/Binding/SKImageFilter.cs
@@ -74,12 +74,21 @@ namespace SkiaSharp
 
 		// CreateMatrix
 
+		public static SKImageFilter CreateMatrix(ref SKMatrix matrix)
+		{
+			fixed (SKMatrix* m = &matrix)
+				return GetObject(SkiaApi.sk_imagefilter_new_matrix(m, default, IntPtr.Zero));
+		}
+
 		public static SKImageFilter CreateMatrix(SKMatrix matrix, SKFilterQuality quality, SKImageFilter input = null)
 		{
 			return GetObject(SkiaApi.sk_imagefilter_new_matrix(&matrix, quality, input == null ? IntPtr.Zero : input.Handle));
 		}
 
 		// CreateAlphaThreshold
+
+		public static SKImageFilter CreateAlphaThreshold (SKRectI region, float innerThreshold, float outerThreshold) =>
+			CreateAlphaThreshold (region, innerThreshold, outerThreshold, null);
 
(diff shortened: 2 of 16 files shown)


maxkatz6:
@mattleibow I went through breaking changes page, and these changes were the low hanging fruits of what can be backported from 3.0 to 2.x. Which is also enough for Avalonia to run both versions. 

Initially I also planned to backport some new SKRuntimeEffect methods (which we don't need), but it feels a bit intrusive, as I am not sure what to pass as isOpaque parameter.

binding/Binding/SKCanvas.cs
```diff
@@ -952,6 +952,13 @@ public void SetMatrix (SKMatrix matrix)
 			SkiaApi.sk_canvas_set_matrix (Handle, &matrix);
 		}
 
+		public void SetMatrix (in SKMatrix matrix)
```
maxkatz6, line 955: Note: C# will always prefer more specific overload. Which means for 2.x builds old code will still use old `SetMatrix(SKMatrix)`, unless they specify `in` explicitly.

workgroupengineering:
What do you think if instead of adding members to existing classes you added them as extension methods in a class called `CompatibilyExtensions`?

maxkatz6:
@workgroupengineering this extension method would have to be added to 3.0 as well. So, it can be binary compatible. Imo, it would add more unwanted API surface to maintain.

binding/Binding/SKCanvas.cs
mattleibow: One annoying thing is that this will make this method ambiguous to visual basic... VB appears not to be able to distinguish between in/ref and nothing. I know we have a scattering of ref and nothing in 2.x and I tried to clean this all up. However, maybe this is making migrations a bit more annoying.

binding/Binding/SKImageFilter.cs
```diff
@@ -74,13 +74,22 @@ internal SKImageFilter(IntPtr handle, bool owns)
 
 		// CreateMatrix
 
+		public static SKImageFilter CreateMatrix(ref SKMatrix matrix)
```
mattleibow, line 77: I think this is `in` in 3.x.

binding/Binding/SKCanvas.cs
maxkatz6: Unfortunately, you are right. Just tested locally, and found this issue confirming it https://github.com/dotnet/vblang/issues/505

maxkatz6: I guess the only choice would be to use reflection/UnsafeAccessor then. I don't really want to modify SkiaSharp 3.0 API surface.

binding/Binding/SKImageFilter.cs
maxkatz6: Fixed

maxkatz6:
@mattleibow update the PR, removed "in" overloads that will break VB. Current PR changes will already help with some migration, though not ideal.

mattleibow:
This was idea and my testing of prototypes seemed to work nicely.

https://github.com/mono/SkiaSharp/pull/2789

maxkatz6:
@mattleibow I like your idea. Closing this PR then.

maxkatz6:
Although, wait, that kind of API changes of this PR cannot be added to 3.0, as it already has these methods :D 

Unless we add back overloads with SKImageFilter and SKImageFilter.CropRect: 
https://github.com/mono/SkiaSharp/blob/v2.88.7/binding/Binding/SKImageFilter.cs#L101

Imo, for these methods specifically, it would be simpler to get this PR to 2.88 with overloads without optional parameters, keeping common API.

mattleibow:
Ah yeah. Both. We can backport anything too ugly to do - such as croprect. I'm working on making 2.x shippable again.

mattleibow:
I did some looking around and I had to also remove the default args from the existing ones. https://github.com/mono/SkiaSharp/pull/2810

For example, if there is this:

```cs
public static SKImageFilter CreatePaint(SKPaint paint, SKImageFilter.CropRect cropRect = null)
```

And we just add:

```cs
public static SKImageFilter CreatePaint (SKPaint paint)
```

People that are using it like this:

```cs
SKImageFilter CreatePaint(mypaint);
```

Will now get an ambiguous overload as both match. So I removed the ` = null` value so if you are not specifying a crop rect, then you are going to use the new overload. If you are, well you have specified a value.

mattleibow:
Thanks for this PR. I am going to close this one as I think I got all the APIs from 3.x in my PR and a few more things: https://github.com/mono/SkiaSharp/pull/2810

PR #2786, opened 2024-03-05
[CI] Use the APIScan dedicated pool.

mandel-macaque:
APIScan calculates the number of CPUs of the vm and throttles accordingly, using a better image results in a huge improvement.


mattleibow:
Thanks for the PR and the help! Switching the entire build to the new monsta masheens https://github.com/mono/SkiaSharp/pull/2785

PR #2815, opened 2024-03-28
Installing Java is troublesome, so avoid it

mattleibow:
**Description of Change**

There are some sort of weird corruption issues on macOS


mattleibow:
I think it is actually the agents as installing python is giving similar errors. But, this PR still reduces the things to install - saving time and also taking out a potential failure point later on.

PR #2854, opened 2024-05-09
fix: XamlRoot may be null when the SKXamlCanvas is unloaded

jeromelaban:
**Description of Change**

Avoids an NRE when using a null `XamlRoot` on WinUI when the control is unloading.

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.


jeromelaban:
@mattleibow could it be possible to backport this one to 2.88? Thanks!

jdobelstein (commented):
Is a null check in on the [root.Invalidate](https://github.com/mono/SkiaSharp/blob/0c0d311bda230ca450a70bd105fab9b4a2c895c8/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs#L104) call needed for the same reason? Couldn't hurt.

jeromelaban:
> [jeromelaban](https://github.com/mono/SkiaSharp/commits?author=jeromelaban) is a null check in on the [root.Invalidate](https://github.com/mono/SkiaSharp/blob/0c0d311bda230ca450a70bd105fab9b4a2c895c8/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs#L104) call needed for the same reason? Couldn't hurt.

That's fair. I ignored that particular site because the `XamlRoot.Changed` event is called only when `XamlRoot` is set, but indeed, it does not hurt.

mattleibow (approved):

PR #2917, opened 2024-07-03
Function pointers and LibraryImport

### LINKED ISSUE #1928: [QUESTION] Minimal WebAssembly example with NativeAOT-LLVM
Hi,

I'm trying to see if its possible to use SkiaSharp with NativeAOT-LLVM (https://github.com/dotnet/runtimelab/tree/feature/NativeAOT-LLVM).  I've created what I think is a pretty small example (https://github.com/yowl/wskia) .  You can build it with the command line (assuming you have SkiaSharp 2.0.12)
```
dotnet publish /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false  --self-contained /p:EmccExtraArgs=.packages\skiasharp.nativeassets.webassembly\2.80.3\build\netstandard1.0\libSkiaSharp.a\2.0.12\libSkiaSharp.a
```
It attempts to fill the canvas with white and draws some rectangles, there's no errors but nothing appears on the canvas, it stays black.  Is the set up code ok?   Copied below (and removed the debug that's in the original for clarity):

```cs
        [UnmanagedCallersOnly(EntryPoint = "MainLoop")]
        static void MainLoop()
        {
            SKCanvas canvas = surface.Canvas;

            canvas.Clear(SKColors.White);

            // configure our brush
            var redBrush = new SKPaint
            {
                Color = new SKColor(0xff, 0, 0),
                IsStroke = true
            };
            var blueBrush = new SKPaint
            {
                Color = new SKColor(0, 0, 0xff),
                IsStroke = true
            };

            for (int i = 0; i < 64; i += 8)
            {
                var rect = new SKRect(i, i, 256 - i - 1, 256 - i - 1);
                canvas.DrawRect(rect, (i % 16 == 0) ? redBrush : blueBrush);
            }
        }
        static SKSurface surface;

        static void Main()
        {
            EmscriptenWebGLContextAttributes attrs;
            emscripten_webgl_init_context_attributes(&attrs);

            attrs.stencil = 8;
            // attrs.majorVersion = 2;
            fixed (byte* n = &(s[0]))
            {
                glContext = emscripten_webgl_create_context(n, &attrs); // this is "#canvas"
            }

            emscripten_webgl_make_context_current(glContext);

            GRGlInterface grGlInterface = GRGlInterface.Create();

            GRContext grContext = GRContext.CreateGl(grGlInterface);

            emscripten_set_main_loop(&MainLoop, 0, 0);

            const SKColorType colorType = SKColorType.Rgba8888;

            int width, height;
            fixed (byte* n = &(s[0]))
            {
                emscripten_get_canvas_element_size(n, &width, &height);
            }

            var info = new GRGlFramebufferInfo(0, colorType.ToGlSizedFormat());

            surface = SKSurface.Create(grContext, new GRBackendRenderTarget(width, height, 0, 8, info),
                colorType);
            if (surface != null)
            {
                Console.WriteLine("got a surface");
            }
            else
            {
                Console.WriteLine("Failed to create surface");
            }

            surface.Canvas.DrawColor(SKColors.White);
    }
```

### LINKED ISSUE #1931: [BUG] flushClearColor  throws an error
**Description**

Attempting to `surface.Canvas.DrawColor(SKColors.White);` and flush causes an exception

**Code**

```
            surface.Canvas.DrawColor(SKColors.White);

            surface.Canvas.Flush();
```

**Expected Behavior**

The canvas goes white

**Actual Behavior**

Exception occurs, 
```
"RuntimeError: null function or function signature mismatch
    at GrGLFunction<void (float, float, float, float)>::GrGLFunction(void (*)(float, float, float, float))::'lambda'(void const*, float, float, float, float)::__invoke(void const*, float, float, float, float) (http://localhost:8000/wskia.wasm:wasm-function[42972]:0x153e1ca)
    at GrGLGpu::flushClearColor(SkRGBA4f<(SkAlphaType)2> const&) (http://localhost:8000/wskia.wasm:wasm-function[43322]:0x155382e)
    at GrGLGpu::beginCommandBuffer(GrRenderTarget*, SkIRect const&, GrSurfaceOrigin, GrOpsRenderPass::LoadAndStoreInfo const&, GrOpsRenderPass::StencilLoadAndStoreInfo const&) (http://localhost:8000/wskia.wasm:wasm-function[43337]:0x155583a)
    at GrGLOpsRenderPass::begin() (http://localhost:8000/wskia.wasm:wasm-function[43382]:0x155b466)
    at GrOpsTask::onExecute(GrOpFlushState*) (http://localhost:8000/wskia.wasm:wasm-function[43522]:0x1562580)
    at GrDrawingManager::executeRenderTasks(int, int, GrOpFlushState*, int*) (http://localhost:8000/wskia.wasm:wasm-function[42907]:0x1537823)
    at GrDrawingManager::flush(GrSurfaceProxy**, int, SkSurface::BackendSurfaceAccess, GrFlushInfo const&, GrPrepareForExternalIORequests const&) (http://localhost:8000/wskia.wasm:wasm-function[42906]:0x15373aa)
    at GrDrawingManager::flushSurfaces(GrSurfaceProxy**, int, SkSurface::BackendSurfaceAccess, GrFlushInfo const&) (http://localhost:8000/wskia.wasm:wasm-function[42908]:0x15379d8)
    at GrRenderTargetContext::flush(SkSurface::BackendSurfaceAccess, GrFlushInfo const&) (http://localhost:8000/wskia.wasm:wasm-function[32186]:0x1180ca5)
    at SkGpuDevice::flush() (http://localhost:8000/wskia.wasm:wasm-function[36719]:0x12ef302)"
```

**Basic Information**

- Version with issue:  WebAssembly 2.0.12
- Last known good version:  
- IDE:  Visual Studio 
- Platform Target Frameworks: WebAssembly/emscripten
- Target Devices:   Chrome Canary
  
<details>
  <summary>Detailed IDE/OS information (click to expand)</summary>
  
```

PASTE ANY DETAILED VERSION INFO HERE

```
</details>

**Screenshots**

**Reproduction Link**

https://github.com/yowl/wskia/tree/flush

and build with 

```
dotnet publish /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false  --self-contained /p:EmccExtraArgs=.packages\skiasharp.nativeassets.webassembly\2.80.3\build\netstandard1.0\libSkiaSharp.a\2.0.12\libSkiaSharp.a
```
Then server using some webserver and open in Chrome canary

### LINKED ISSUE #2615: [Epic] Use newer/better interop primitives
## Description

The SkiaSharp codebase is really old and is using many old primitives. Some such things are arrays instead of renting/span or MonoTouch attributes.

This issue is an epic issue to track what we would like. If there is something I missed or some new feature that will improve performance, please add a comment or even open a PR.

### Tasks

- [ ] #2616  
     https://learn.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines
- [ ] Switch to native function pointers  
     https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers
     https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#function-pointers
- [ ] Remove `[MonoPInvokeCallback]` attributes  
- [ ] Use Intrinsics where we can/should

### LINKED ISSUE #2916: [BUG] Delegate proxies crash with NRE when user delegate is null (only if rewrite to function pointers)
### Description

~This is a weird issue, as I wasn't able to reproduce it with a WASM Mono runtime or normal desktop application. But only with NativeAOT-LLVM.~
It seems this issue is only reproducible after I replaced managed delegates with function pointers as part of https://github.com/mono/SkiaSharp/issues/2615

But I decided to open this issue here instead, as it actually looks like a bug in SkiaSharp code, and exception is somehow swallowed outside of it.

### Code

SkiaSharp has many methods accepting callbacks. Like:
```
public bool InstallPixels (SKImageInfo info, IntPtr pixels, int rowBytes, SKBitmapReleaseDelegate releaseProc, object context)
{
	var cinfo = SKImageInfoNative.FromManaged (ref info);
	var del = releaseProc != null && context != null
		? new SKBitmapReleaseDelegate ((addr, _) => releaseProc (addr, context))
		: releaseProc;
	DelegateProxies.Create (del, out _, out var ctx);
	return SkiaApi.sk_bitmap_install_pixels (Handle, &cinfo, (void*)pixels, (IntPtr)rowBytes, DelegateProxies.SKBitmapReleaseDelegateProxy, (void*)ctx);
}
```

The problem with this callback, that it passes proxy function to the DllImport even when releaseProc is null.

And from proxy implementation, it will always crash with NRE, as `context` is an empty pointer.
```
private static void SKSurfaceReleaseDelegateProxyImplementation (void* address, void* context)
{
	var del = Get<SKSurfaceReleaseDelegate> ((IntPtr)context, out var gch);
	try {
		del.Invoke ((IntPtr)address, null);
	} finally {
		gch.Free ();
	}
}
```

### Expected Behavior

No exception.

### Actual Behavior

Exception.

### Version of SkiaSharp

3.x (Alpha)

### IDE / Editor

Visual Studio (Windows)

Reproduced on WASM (both Mono and NativeAOT)
_

### Relevant Log Output

<div aria-expanded="true" tabindex="-1" style="box-sizing: border-box; min-width: 0px; min-height: 0px; flex: 0 0 auto; color: rgb(255, 255, 255); font-family: consolas, &quot;lucida console&quot;, &quot;courier new&quot;, monospace; font-size: 12px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(0, 74, 119); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="source-code" style="box-sizing: border-box; min-width: 0px; min-height: 0px; font-family: var(--source-code-font-family); white-space: pre-wrap; font-size: var(--source-code-font-size) !important; line-height: 1.2;"><span class="console-message-anchor" style="box-sizing: border-box; min-width: 0px; min-height: 0px; float: right; text-align: right; max-width: 100%; margin-left: 4px;"><button class="devtools-link text-button link-style" jslog="Link; track: click" role="link" tabindex="-1" title="http://localhost:49512/dotnet.native.js:1553" style="box-sizing: border-box; min-width: 0px; min-height: 0px; font: inherit; color: var(--sys-color-primary); border: none; border-radius: 2px; background: none; margin: 0px; height: unset; padding: 0px !important; flex: 0 0 auto; white-space: nowrap; text-decoration: underline; outline-offset: 2px; cursor: pointer; outline: none; max-width: 100%; overflow: hidden; text-overflow: ellipsis; word-break: break-all;"><br class="Apple-interchange-newline">dotnet.native.js:1553</button> <devtools-icon role="button" class="resource-links console-search-icon" name="edge-search-info" title="Search for this message on the Web" aria-label="Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object." tabindex="0" style="flex-grow: 0; flex-shrink: 0; display: inline-block; width: 16px; height: 16px; color: var(--color-link); vertical-align: sub; position: relative; box-sizing: border-box; min-width: 0px; min-height: 0px; margin-top: -1px; margin-bottom: -4px; cursor: pointer;"><span style="display: block; width: 16px; height: 16px; background-color: currentcolor; mask: var(--icon-url) center / contain no-repeat; --icon-url: var(--image-file-edge-search-info);"></span></devtools-icon> </span><span class="console-message-text" style="box-sizing: border-box; min-width: 0px; min-height: 0px; color: var(--override-error-text-color) !important;">Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.</span></span></div><div class="" role="group" style="box-sizing: border-box; min-width: 0px; min-height: 0px; flex: 0 0 auto; color: rgb(255, 255, 255); font-family: consolas, &quot;lucida console&quot;, &quot;courier new&quot;, monospace; font-size: 12px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(0, 74, 119); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="monospace stack-preview-container width-constrained" style="display: inline-block; --monospace-font-size: 12px; --monospace-font-family: consolas, lucida console, courier new, monospace; --source-code-font-size: 12px; --source-code-font-family: consolas, lucida console, courier new, monospace; width: 768px; box-sizing: border-box; min-width: 0px; min-height: 0px; font-family: var(--monospace-font-family); font-size: var(--monospace-font-size) !important;">
  | put_char | @ | dotnet.native.js:1553
-- | -- | -- | --
  | $SystemNative_LogError | @ | dotnet.native.wasm:0x14219fc
  | $S_P_CoreLib_Internal_Console_Error__Write | @ | dotnet.native.wasm:0x6a6783
  | $S_P_CoreLib_System_RuntimeExceptionHelpers__FailFast | @ | dotnet.native.wasm:0xd0c2d8
  | $S_P_CoreLib_System_RuntimeExceptionHelpers__RuntimeFailFast | @ | dotnet.native.wasm:0x10b8b5b
  | $S_P_CoreLib_System_Runtime_EH__HandleUnhandledException | @ | dotnet.native.wasm:0x10b6dfa
  | $SkiaSharp_SkiaSharp_DelegateProxies__SKBitmapReleaseDelegateProxyImplementation$F1_Filter | @ | dotnet.native.wasm:0xed436a
  | $S_P_CoreLib_System_Runtime_EH__CallFilterFunclet | @ | dotnet.native.wasm:0x114210f
  | $S_P_CoreLib_System_Runtime_EH__DispatchException | @ | dotnet.native.wasm:0xd903c
  | $S_P_CoreLib_System_Runtime_EH__RhpThrowEx | @ | dotnet.native.wasm:0x108ba01
  | $S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowNullReferenceException | @ | dotnet.native.wasm:0x108b96d
  | $SkiaSharp_SkiaSharp_DelegateProxies__SKBitmapReleaseDelegateProxyImplementation | @ | dotnet.native.wasm:0xed4317
  | $SkMakePixelRefWithProc(int, int, unsigned long, void*, void (*)(void*, void*), void*)::PixelRef::~PixelRef() | @ | dotnet.native.wasm:0x185b3e4
  | $SkRefCntBase::internal_dispose() const | @ | dotnet.native.wasm:0x18765ed
  | $SkBitmap::~SkBitmap() | @ | dotnet.native.wasm:0x185b791
  | $sk_bitmap_destructor | @ | dotnet.native.wasm:0x1943b5e
  | $SkiaSharp_SkiaSharp_SKBitmap__DisposeNative | @ | dotnet.native.wasm:0xdbe66c
  | $SkiaSharp_SkiaSharp_SKNativeObject__Dispose | @ | dotnet.native.wasm:0xa68216
  | $SkiaSharp_SkiaSharp_SKBitmap__Dispose | @ | dotnet.native.wasm:0xdbe69d
  | $SkiaSharp_SkiaSharp_SKNativeObject__Dispose_0 | @ | dotnet.native.wasm:0xc74915
  | $Avalonia_Skia_Avalonia_Skia_ImmutableBitmap___ctor_3

</span></div>
</body>
</html>

### Code of Conduct

maxkatz6:
**Description of Change**

This PR is pretty big. If necessary, I can split HarfBuzz and Skia changes on two PRs. In theory, generated code can be be extracted into a separated PR as well, since all changes should be under conditional compilation.

What was changed step by step:
- Generator was updated to emit LibraryImport instead of DllImport, when USE_LIBRARY_IMPORT is set.
- Generator was updated to emit function pointers instead of managed delegates, when USE_LIBRARY_IMPORT is used.
- Since function pointers can't be used as generics, every `DelegateProxies.Create<T>` usage was replaced without generics.
- Since Skia expects `bool` as 1 byte, and function pointers by default marshal it as 4 bytes, I had to disable runtime marshalling. This change required to replace remaining DllImports with LibraryImport as well. As a bonus - final app size should be smaller without runtime marshalling (assuming other dependencies don't use it too).

Note:
- Due to older Mono bug, LibraryImport were defined with `void*` instead of full function pointer, but it doesn't seem to cause other issues
- Generated file diffs only have additions and no removals, tried to keep this way to keep less impact.
- For HarfBuss I used 2.8.2 native baseline version, as it seems latest when it was regenerated (to keep diff smaller).
- Tested on Windows and WASM (only .NET 9, but both Mono and LLVM). Will test on mobile and mac once there is any nightly build.

**Bugs Fixed**

- Fixes #2615
- Fixes #2916 (only reproducible with function pointers)

Not sure if this PR fixed them, or some another, but these two issues are not reproducible anymore:
- Fixes #1928
- Fixes #1931

**API Changes**

None. Unless I missed any.

**Behavioral Changes**

None. Hopefully.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


mattleibow:
YES!

mattleibow:
So far this:

```
D:\a\1\s\binding\SkiaSharp\GRGlInterface.cs(216,83): error CA1420: Managed parameter or return types require runtime marshalling to be enabled (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1420) [D:\a\1\s\binding\SkiaSharp\SkiaSharp.csproj::TargetFramework=net7.0-tizen]
D:\a\1\s\binding\SkiaSharp\GRGlInterface.cs(234,11): error CA1421: 'System.Runtime.InteropServices.Marshal.PtrToStructure<SkiaSharp.GRGlInterface.EvasGlApi>(nint)' uses runtime marshalling even when 'DisableRuntimeMarshallingAttribute' is applied. Use features like 'sizeof' and pointers directly to ensure accurate results. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1421) [D:\a\1\s\binding\SkiaSharp\SkiaSharp.csproj::TargetFramework=net7.0-tizen]
```

mattleibow:
The device failures are because all my certs just expired. The other test failures are the fact that the main branch is read and cannot pull assets. I have updated certs and re-triggered a full build.

The compile will still fail for windows I think, but the devices may run. Not 10% sure as I think there may be some bad conditionals somewhere causing the error.

mattleibow:
> Will test on mobile and mac once there is any nightly build.
 
The device tests should be running for iOS, Android and macOS for .NET 7, so we shall see. I do need to add .NET 8 but I will do that once this is merged.

mattleibow:
New build here: https://dev.azure.com/xamarin/public/_build/results?buildId=118700&view=results

maxkatz6:
> So far this:

Thanks. Pushed another commit that should fix tizen build. EvasGlApi+reflection probably could be replaced with a `nint[]` array + named dictionary over it, but that's something for another day.

maxkatz6:
Also updated remaining main interop projects for consistency: SkiaSharp.Resources, SkiaSharp.SceneGraph, SkiaSharp.Skottie. Didn't touch `SkiaSharp.Views.**` projects, as these would require even more changes.

maxkatz6:
@mattleibow tests are failing, but I am a bit confused why.
From linux (and similar iOS) build tests, it fails on:
```
  Error Message:
   System.EntryPointNotFoundException : Unable to find an entry point named 'sk_rtree_factory_new' in shared library 'libSkiaSharp'.
  Stack Trace:
     at SkiaSharp.SkiaApi.sk_rtree_factory_new()
   at SkiaSharp.SKPictureRecorder.BeginRecording(SKRect cullRect, Boolean useRTree) in /home/vsts/work/1/s/binding/SkiaSharp/SKPictureRecorder.cs:line 43
   at SkiaSharp.Tests.SKPictureRecorderTest.CanCreateDrawableFromRecorder(Boolean useRTree) in /home/vsts/work/1/s/tests/Tests/SkiaSharp/SKPictureRecorderTest.cs:line 60
```

But I tripple checked, and sk_rtree_factory_new is correctly defined in LibraryImport/DllImport, as well as referenced skia submodule has this method included. 
Am I missing something?

mattleibow:
as part of the build I check to see if the native things have changed and then download the previous native binaries. However, the last few builds on main never finished. I triggered them so should work now.

maxkatz6:
@mattleibow side question, but what frameworks does skia sharp still supports? Especially from Mono world.
It seems like net framework does support function pointers, if you provide UnmanagedCallersOnly polyfill. But I doubt old Xamarin/Mono will work this way, without MonoPInvokeCallback. It's still technically can be used via netstandard target of the lib.

filipnavara:
> Xamarin/Mono

The old mono/mono (along with Xamarin.iOS) went out of support earlier this year. I doubt there's any point in wasting more time on it. If someone uses the unsupported framework they can stick to older SkiaSharp release.

maxkatz6:
Another problem is DisableRuntimeMarshalling, which also makes it easier to marshal bools in function pointers. Since DisableRuntimeMarshalling is not supported in .NET Framework, I don't think I can use function pointers without replacing all booleans with bytes or something.

mattleibow:
> Another problem is DisableRuntimeMarshalling, which also makes it easier to marshal bools in function pointers. Since DisableRuntimeMarshalling is not supported in .NET Framework, I don't think I can use function pointers without replacing all booleans with bytes or something.

We can do this in the generator... I usually add a `[MarshalAs (UnmanagedType.I1)]` to everything that is bool. Does that stop working with LibraryImport?

```cs
internal static extern sk_maskfilter_t sk_maskfilter_new_blur_with_flags (SKBlurStyle param0, Single sigma, [MarshalAs (UnmanagedType.I1)] bool respectCTM);
```

mattleibow:
> @mattleibow side question, but what frameworks does skia sharp still supports? Especially from Mono world. It seems like net framework does support function pointers, if you provide UnmanagedCallersOnly polyfill. But I doubt old Xamarin/Mono will work this way, without MonoPInvokeCallback. It's still technically can be used via netstandard target of the lib.

You can see from the root package: https://www.nuget.org/packages/SkiaSharp/3.0.0-preview.3.1#supportedframeworks-body-tab

And this PR is going to bump the SDK to net8 things: https://github.com/mono/SkiaSharp/pull/2927

Basically, whatever you can build with the .NET 8 SDK:

* `net6.0` for console/server apps (since that is still supported by MS)
* `net8.0` for console/server apps
* `net8.0-<platform>` for desktop/mobile
* `netstandard2.0` (just for MSBuild tasks because VS is still using full framework)
* `net462` (literally just to appease VS build tasks like resizetizer)

In summary, just net8.0 and the basics of net6.0 because that is still LTS. Then also netstandard and netfx because we have to use VS.

If we have to do bad and slow things for netstandard and netfx then we can. If you want perf, then you want a real modern framework. For me full framework is _just_ because the IDE is still on that and most machines running VS are more powerful so we can go slow paths. I support no fancy things like AOT/linking/anything with netfx at all and if we break some corner case with some corner thing, then we can.

maxkatz6:
>  I usually add a [MarshalAs (UnmanagedType.I1)] to everything that is bool. 

I don't think it's possible to add an attribute on function ponter.

mattleibow:
This works (builds at least):

```cs
[LibraryImport ("test")]
internal static partial nint test_method (nint a, [MarshalAs (UnmanagedType.I1)] bool b);

public static readonly delegate* unmanaged[Cdecl] <nint, bool, nint> Proxy = &Testing;

[UnmanagedCallersOnly(CallConvs = new [] {typeof(CallConvCdecl)})]
internal static nint Testing (nint a, [MarshalAs (UnmanagedType.I1)] bool b)
{
    return 0;
}
```

Generated code:

```cs
public unsafe static readonly global::System.UIntPtr/*delegate* unmanaged[Cdecl]<global::System.IntPtr, bool, global::System.IntPtr>*/ Proxy = (global::System.UIntPtr/*delegate* unmanaged[Cdecl]<global::System.IntPtr, bool, global::System.IntPtr>*/)(void*)(ulong)(global::System.UIntPtr/*delegate*<global::System.IntPtr, bool, global::System.IntPtr>*/)(&Testing);

[LibraryImport("test")]
[GeneratedCode("Microsoft.Interop.LibraryImportGenerator", "9.0.10.30607")]
[SkipLocalsInit]
internal static global::System.IntPtr test_method(global::System.IntPtr a, [MarshalAs(3)] bool b)
{
    sbyte __b_native2 = (sbyte)(b ? 1 : 0);
    return __PInvoke(a, __b_native2);
    [DllImport("test", EntryPoint = "test_method", ExactSpelling = true)]
    [CompilerGenerated]
    static extern global::System.IntPtr __PInvoke(global::System.IntPtr __a_native, sbyte __b_native);
}

[UnmanagedCallersOnly(CallConvs = new global::System.Type[] { typeof(CallConvCdecl) })]
internal static global::System.IntPtr Testing(global::System.IntPtr a, [MarshalAs(3)] bool b)
{
    return (global::System.IntPtr)0;
}
```

maxkatz6:
@mattleibow I was wrong with UnmanagedCallersOnly being possible to polyfill on net framework. It doesn't seem to be guaranteed to work on all - https://github.com/AvaloniaUI/Avalonia/pull/16295#discussion_r1676800518. Also confirmed by runtime team.
Old delegates have to stay, I guess.

mattleibow:
I wonder if there is a source generator we could write/use to convert the `UnmanagedCallersOnly` into the old way? Then the code we type is the same, but at compile time we get the old things... Do source generators work for the old netfx tfms?

maxkatz6:
@mattleibow it should be possible to write a source generator (instead of current tool) that will find methods with a custom attribute:
```csharp
[PInvokeCallback]
internal static partial void ReleaseDelegateProxyImplementation (void* context)
{
}

// and later
var callback = DelegateProxies.ReleaseDelegateProxy; // ReleaseDelegateProxy is fully generated with a correct type
```

and generate:
```csharp
#if NET7_0_OR_GREATER

public static readonly delegate* unmanaged[Cdecl] <void*, void> ReleaseDelegateProxy = &ReleaseDelegateProxyImplementation;
[UnmanagedCallersOnly(CallConvs = new [] {typeof(CallConvCdecl)})]
internal static partial void ReleaseDelegateProxyImplementation();

#else

public static readonly DestroyProxyDelegate ReleaseDelegateProxy = ReleaseDelegateProxyImplementation;
[MonoPInvokeCallback (typeof (DestroyProxyDelegate))]
internal static partial void ReleaseDelegateProxyImplementation();

#endif
```

maxkatz6:
Or, since you already know ahead all the callbacks, can avoid an extra attribute with source generators, and just generate these proxy headers with current tool.

maxkatz6:
Either way it should work with netframework, since it's just a source code generation, nothing different in runtime.

maxkatz6:
@mattleibow just pushed a commit moving all the #if #else from DelegateProxy files to auto generated code. I reused the same generator tool for now just by extending it.

This way almost all the boilerplate is hidden behind generated code.

mattleibow:
Yeah, much smarter... Writing a source generator to generate source for generated sources... At this point am I thinking like AI.

maxkatz6:
@mattleibow hi!
Checked CI build and found two problems with tests:
1. I used `protected internal` in methods used by delegate proxies. It doesn't affect library consumers as per [protected internal (C# Reference)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal#overriding-protected-internal-members) (If the derived class is defined in a different assembly from the base class, overridden members have protected access. - which is exactly how it worked before this PR). But since Tests projects has InternalsVisibleTo configured, it also has to use `protected internal`.
2. There were two tests asserting generated managed delegates. Since these managed delegates are not generated anymore in .NET7+ builds, I disabled them. .NET Framework tests still check for them.

Hopefully CI should pass now.
Do you think anything else needs to be done for this PR?

mattleibow:
After _much_ thrashing on CI just to get a green main branch, I see we now have some failing tests that are related to this PR.

tests/SkiaSharp.Tests.Console/SkiaSharp.Tests.Console.csproj
```diff
@@ -12,6 +12,10 @@
     <Platforms>AnyCPU;x64;x86</Platforms>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">
```
mattleibow, line 15:
```suggestion
  <PropertyGroup Condition="!$(TargetFramework.StartsWith('net4'))">
```

filipnavara:
> The first one is interesting be cause it can't find `LoadLibrary`. I see it is actually `LoadLibraryA`: https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya Not sure if the new attributes require special treatment of this? I see the old way had `CharSet = CharSet.Ansi`.

The distinction of `A`/`W`-suffixed APIs has not been relevant for a long while. All supported Windows versions have `LoadLibraryW` Unicode API and it should be used, if necessary by specifying `Entrypoint = "LoadLibraryW", StringMarshalling = StringMarshalling.Utf16` in the `LibraryImport` attribute. (`LoadLibraryA` is legacy API dating back to Windows 9x era.)

Modern .NET has [NativeLibrary.Load](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.load?view=net-8.0) but I guess that may not be applicable here if the tests still run on .NET Framework.

binding/SkiaSharp/GRGlInterface.cs
```diff
@@ -132,6 +137,12 @@ private static class AngleLoader
 			private static extern IntPtr GetProcAddress (IntPtr hModule, [MarshalAs (UnmanagedType.LPStr)] string lpProcName);
 
 			private static IntPtr LoadLibrary (string lpFileName) => LoadPackagedLibrary(lpFileName, 0);
+#elif USE_LIBRARY_IMPORT
+			[LibraryImport ("Kernel32.dll", SetLastError = true)]
```
filipnavara, line 141:
```suggestion
			[LibraryImport ("Kernel32.dll", SetLastError = true, Entrypoint = "LoadLibraryW", StringMarshalling = StringMarshalling.Utf16)]
```

binding/SkiaSharp/GRGlInterface.cs
```diff
@@ -132,6 +137,12 @@ private static class AngleLoader
...
+			private static partial IntPtr LoadLibrary ([MarshalAs (UnmanagedType.LPStr)] string lpFileName);
```
filipnavara, line 142:
```suggestion
			[LibraryImport ("Kernel32.dll", SetLastError = true, Entrypoint = "LoadLibraryW", StringMarshalling = StringMarshalling.Utf16)]
			private static partial IntPtr LoadLibrary (string lpFileName);
```

mattleibow:
I need to check that these new things are not slower: https://github.com/dotnet/runtime/issues/106491

binding/SkiaSharp/GRGlInterface.cs
maxkatz6: Thank you.

maxkatz6:
@mattleibow both tests should be fixed now.

I can run our control catalog on my iOS device later this week.

binding/SkiaSharp/GRGlInterface.cs
```diff
@@ -132,6 +137,12 @@ private static class AngleLoader
 			private static extern IntPtr GetProcAddress (IntPtr hModule, [MarshalAs (UnmanagedType.LPStr)] string lpProcName);
 
 			private static IntPtr LoadLibrary (string lpFileName) => LoadPackagedLibrary(lpFileName, 0);
+#elif USE_LIBRARY_IMPORT
+			[LibraryImport ("Kernel32.dll", SetLastError = true, Entrypoint = "LoadLibraryW", StringMarshalling = StringMarshalling.Utf16)]
```
mattleibow, line 141:
```suggestion
			[LibraryImport ("Kernel32.dll", SetLastError = true, EntryPoint = "LoadLibraryW", StringMarshalling = StringMarshalling.Utf16)]
```

filipnavara: Sorry for the typo

mattleibow:
Woohoo! This is finally green and finally good to go. Merged and now time for that <beverage of choice>! 

Thanks so much for this! This has been on the list since net7.0 started doing cool things.

PR #2922, opened 2024-07-04
Do not stretch the background

mattleibow:
**Description of Change**

When a view is resized, the canvas is not yet resized and updated to match. Before this PR, the canvas would be stretched to fill the view. This PR make sure to not stretch the canvas but instead keep it fixed.

The reason for the scaling in the first place is that `WriteableBitmap` does not have a DPI parameter. As a result, we have to create a larger bitmap than the view such that when it is scaled down to match the view size, the resolution matches that of the rest of the app. Previously, the larger canvas was just squashed to fit, but this was non-uniform. This PR always scales the canvas to the correct size preserving the aspect ratio.

**Bugs Fixed**

- None.

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using Windows.ApplicationModel;
-using Windows.Graphics.Display;
 
 #if WINDOWS
 using Microsoft.UI.Dispatching;
@@ -39,6 +38,7 @@ namespace SkiaSharp.Views.UWP
 
 		private IntPtr pixels;
 		private WriteableBitmap bitmap;
+		private ImageBrush brush;
 		private bool ignorePixelScaling;
 		private bool isVisible = true;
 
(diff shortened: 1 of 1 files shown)


source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs
```diff
@@ -101,8 +101,13 @@ private static void OnVisibilityChanged(DependencyObject d, DependencyPropertyCh
 		private void OnXamlRootChanged(XamlRoot xamlRoot = null, XamlRootChangedEventArgs e = null)
 		{
 			var root = xamlRoot ?? XamlRoot;
-			Dpi = root?.RasterizationScale ?? 1.0;
-			Invalidate();
+			var newDpi = root?.RasterizationScale ?? 1.0;
+			if (newDpi != Dpi)
```
mattleibow, line 105: This check here prevents hundreds of invalidations. The `XamlRootChanged` event will fire when _any_ of the properties of the XAML root changes. This does not seem to fire just when the current view gets a new XAML root as the name implies.

PR #2620, opened 2023-09-18
Add support for multi-thread and/or SIMD WebAssembly

mattleibow:
**Description of Change**

Support multithreading or SIMD blazor apps out the box.

**Bugs Fixed**

- Fixes #2614
- Fixes https://github.com/mono/SkiaSharp/issues/3037
- Fixes https://github.com/mono/SkiaSharp/issues/3035

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Has tests (if omitted, state reason in description)
- [ ] Rebased on top of main at time of PR
- [ ] Merged related skia PRs
- [ ] Changes adhere to coding standard
- [ ] Updated documentation

PR #2927, opened 2024-07-05
Update .NET to 8.0

mattleibow:
**Description of Change**

Updating .NET to the latest stable and prepare for .NET 9.0.


benbaker76:
Please include an update to the [install-dotnet-workloads.ps1](https://github.com/mono/SkiaSharp/blob/main/scripts/install-dotnet-workloads.ps1#L27) so it will build using the latest `wasm-tools`

mattleibow:
It is there:

https://github.com/mono/SkiaSharp/blob/main/scripts%2Finstall-dotnet-workloads.ps1#L23

benbaker76:
> It is there:

I'm building for Blazor Wasm (.NET 8.0) and it forced me to install `wasm-tools-net6` or it would give me an error `To build this project, the following workloads must be installed: wasm-tools-net6. To install these workloads, run the following command: dotnet workload restore`.

I'm using [Svg.Skia](https://github.com/wieslawsoltes/Svg.Skia) along with `SkiaSharp.Views.Blazor`. Without the latter I'd get `System.DllNotFoundException: libSkiaSharp` (previously I tried `SkiaSharp.NativeAssets.WebAssembly` instead but it didn't work).

mattleibow:
What are you deploying? The `wasm-tools-net6` is only used in net6, and I think nothing is using that in this repo after this PR?

benbaker76:
> What are you deploying? The `wasm-tools-net6` is only used in net6, and I think nothing is using that in this repo after this PR?

My concern is that the script is specifically referring to `wasm-tools-net6` rather than the latest (just `wasm-tools`) but you may be right in that it will use the latest once this PR is merged. I look forward to it and thanks for your work on the project.

jairbubbles:
Hi @mattleibow, do you plan to release a preview package that works on `net9.0` + wasm?

With `SkiaSharp.Views.Blazor@3.0.0-preview.4.1` I get:
```
System.DllNotFoundException: libSkiaSharp
```

When tweaking `SkiaSharp.NativeAssets.WebAssembly.targets` locally to include the native files I get a bit further but ends up with:

```
Unhandled exception rendering component: Could not set up parent class, due to: Could not load type of field 'SkiaSharp.Views.Blazor.Internal.JSModuleInterop:moduleTask' (0) due to: Could not resolve type with token 0100003b from typeref (expected class 'Microsoft.JSInterop.IJSUnmarshalledObjectReference' in assembly 'Microsoft.JSInterop, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60') assembly:Microsoft.JSInterop, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60 type:Microsoft.JSInterop.IJSUnmarshalledObjectReference member:(null) assembly:SkiaSharp.Views.Blazor.dll type:JSModuleInterop member:(null)
```

Please note that I'm trying in .NET 9 as I want to try out:

```xml
<WasmEnableThreads>true</WasmEnableThreads>
```

Which comes with its own set of issues 😅


Thank you!

mattleibow:
Lets talk on this PR that I stopped as things were not actually implemented in .NET at the time: https://github.com/mono/SkiaSharp/pull/2620

PR #2962, opened 2024-07-30
Implement SkCanvas::SaveLayerRec

### LINKED ISSUE #2773: [FEATURE] Expose SkCanvas::SaveLayerRec API

`SkCanvas::SaveLayerRec` is missing from SkiaSharp's API surface, this API is very useful for implementing backdrop effects/filters, we need this API in order to implement [CompositionBackdropBrush](https://learn.microsoft.com/en-us/uwp/api/windows.ui.composition.compositionbackdropbrush) properly in Uno Platform.

It would be great if `SkCanvas::SaveLayerRec` got exposed in SkiaSharp

We currently use a combination of `SKCanvas.SaveLayer(SKPaint)`, `SKCanvas.DrawSurface(SKSurface, SKRect)`, and `SKCanvas.Restore()` as a workaround, but it isn't ideal and has few issues (e.g. need to downscale the canvas manually, no custom clipping, slower since it's not optimized for backdrop effects unlike `SkCanvas::SaveLayerRec`, image filter of SKPaint isn't fully applied to the saved region (due to scaling miscalculations?), etc...)

### Additional context

_No response_

### Code of Conduct

ahmed605:
**Description of Change**

Implement SkCanvas::SaveLayerRec

**Bugs Fixed**

- Fixes #2773 

**API Changes**

Added: 
 
- `int SKCanvas.SaveLayer(SKRect bounds, SKPaint paint, SKImageFilter backdrop, SKSaveLayerFlags saveLayerFlags = SKSaveLayerFlags.None)`
- `int SKCanvas.SaveLayer(SKRect bounds, SKPaint paint, SKSaveLayerFlags saveLayerFlags)`

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/skia/pull/130

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/SkiaSharp/SKCanvas.cs
+++ b/binding/SkiaSharp/SKCanvas.cs
@@ -68,6 +68,16 @@ namespace SkiaSharp
 			return SkiaApi.sk_canvas_save_layer (Handle, null, paint == null ? IntPtr.Zero : paint.Handle);
 		}
 
+		public int SaveLayer (SKRect bounds, SKPaint paint, SKImageFilter backdrop, SKSaveLayerFlags saveLayerFlags = SKSaveLayerFlags.None)
+		{
+			return SkiaApi.sk_canvas_save_layer_rec (Handle, &bounds, paint == null ? IntPtr.Zero : paint.Handle, backdrop == null ? IntPtr.Zero : backdrop.Handle, (uint)saveLayerFlags);
+		}
+
+		public int SaveLayer (SKRect bounds, SKPaint paint, SKSaveLayerFlags saveLayerFlags)
+		{
+			return SaveLayer (bounds, paint, null, saveLayerFlags);
+		}
+
 		public int SaveLayer () =>
 			SaveLayer (null);
 
@@ -1065,5 +1075,14 @@ namespace SkiaSharp
 				canvas = null;
 			}
 		}
+	}
+
+	[Flags]
+	public enum SKSaveLayerFlags
+	{
+		None = 0,
+		PreserveLCDText = 1 << 1,
+		InitWithPrevious = 1 << 2,
+		UseF16ColorType = 1 << 4,
 	}
 }
--- a/binding/SkiaSharp/SkiaApi.generated.cs
+++ b/binding/SkiaSharp/SkiaApi.generated.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Runtime.InteropServices;
 
 #region Namespaces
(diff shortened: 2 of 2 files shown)


mattleibow:
Not sure my magic with CI is working with the auto detection of the skia branch (probably because I am not doing it right for forks) so maybe just update the submodule here and we can merge in one go later once the skia PR is merged.

binding/SkiaSharp/SKCanvas.cs
```diff
@@ -1066,4 +1076,13 @@ public void Restore ()
 			}
 		}
 	}
+
+	[Flags]
+	public enum SKSaveLayerFlags
```
mattleibow, line 1081:
```suggestion
	public enum SKCanvasSaveLayerFlags
```

binding/SkiaSharp/SKCanvas.cs
```diff
@@ -68,6 +68,16 @@ public int SaveLayer (SKPaint paint)
 			return SkiaApi.sk_canvas_save_layer (Handle, null, paint == null ? IntPtr.Zero : paint.Handle);
 		}
 
+		public int SaveLayer (SKRect bounds, SKPaint paint, SKImageFilter backdrop, SKSaveLayerFlags saveLayerFlags = SKSaveLayerFlags.None)
```
mattleibow, line 71: Would it be worth creating a rec struct now instead of adding overloads? The next skia update will have a filters list and a color space, so reducing the new things to a new property may be nicer.

```cs
public struct SKCanvasSaveLayerRec
{
    public SKRect? Bounds { get; set; }
    public SKPaint? Paint  { get; set; }
    public SKImageFilter? Backdrop { get; set; }
    public SKCanvasSaveLayerFlags { get; set; }
}
```

Then in the save layer we can "unpack" this struct and pass it as the single C function. This way we don't need to create yet another overload to the save layer world just to add colorspace.

ramezgerges:
@mattleibow Hello, I'm an Uno Platform dev and I'm following up on this PR (and [the accompanying mono/skia PR](https://github.com/mono/skia/pull/130). I haven't actually tested the new APIs yet, but here are my thoughts.

* The `SaveLayer` signature that takes an `SKCanvasSaveLayerRec` (instead of individual components) seems to be identical to the C++ API, so it makes sense. 
* Shouldn't we make `SKCanvasSaveLayerRec` a struct instead? It seems to be something you create, immediately pass to `SaveLayer`, and then throw away. The example in google/skia (https://github.com/google/skia/blob/d2ac2b0d6467fdca614bd85b49bf01363fe94604/docs/examples/Canvas_SaveLayerRec.cpp#L15) confirms this.

mattleibow:
Thanks for the review @ramezgerges, this makes sense today with the small struct. But, if you look at the latest skia, does your thought still seem valid? The struct is getting fairly large: https://github.com/google/skia/blob/main/include/core/SkCanvas.h#L689 There are new properties to be added - color space, image filters as well as a collection/array of image filters.

Not sure either way here, but at what point to we think a class is better? Will we ever re-use this object for multiple calls?

I was also thinking that due to the large-ish size and potentially fairly complex nature, it should be a class: https://learn.microsoft.com/dotnet/standard/design-guidelines/choosing-between-class-and-struct

However, rules in .NET are just best ideas for most cases, not the law. So, maybe we can break it. But, a struct may still be the best. Not sure if any of this changes anything? Let me know if you still think a struct is better.

ramezgerges:
Thanks for getting back to me, @mattleibow. Just to be clear, the struct-vs-class point is merely a suggestion/observation, and I'm okay with the PR either way.

* I didn't notice the new properties that were added. However, I think the argument is still the same: assuming that 99% of the usescases follow the "Create a SaveLayerRec -> Immediately give it to `SaveLayer` -> forget about it" pattern, then allocating a new object every time seems wasteful. If we're worried about the cost of copying, then can't we just make `SaveLayer` take a `ref` (or a `ref readonly`) parameter? i.e.
`public int SaveLayer (ref readonly SKCanvasSaveLayerRec rec)`

* If we keep it as a class, can we instead add a `Reset` method to reset the SaveLayerRec to its initial state? We've had to deal with excessive SKPath/SKPaint allocations in our main loop before and the workaround I did was to use an ObjectPool + a `Reset()` call to get a "fresh" SKPath/SKPaint.

mattleibow:
@ramezgerges Sorry for the super long wait. I think your comments are the way to go, so I have updated the rec type to be a struct and also use the new `in` keywork so that we don't make a copy of the struct either.

Let me know if this will work for you and if there are any other changes that you think need to be made.

ramezgerges:
@mattleibow No worries. I tested the CI packages today and things are working great!

mattleibow:
Thanks for creating this PR and sticking with it for a year :)

I have merged and it should go out into the preview feed shortly. If there are any issues or adjustments, feel free to open a PR.

Thanks again!

jeromelaban:
Many thanks @mattleibow!

PR #2963, opened 2024-07-31
Add compatibility overloads for SKFilterQuality

Youssef1313:
Libraries may want to remain supporting both SkiaSharp 2.x and 3.x, as we currently do in Uno Platform.

We achieve this by detecting the assembly version and using UnsafeAccessor when needed for APIs that have changed in SkiaSharp 3.

However, there is one remaining issue for Uno Platform, which is SKFilterQuality. As the overloads now take SKSamplingOptions which is a completely new type in SkiaSharp 3, we can't use UnsafeAccessor to get this to work.

This PR adds new obsolete APIs that accepts SKFilterQuality and forwards it to the SKSamplingOptions overload. For Uno Platform, we will use `paint.FilterQuality` when on 2.x, and use the new shader overloads via UnsafeAccessor when on 3.x

NOTE: This is applicable for any library that needs to remain on 2.x while allowing consumers to upgrade to 3.x.

@mattleibow, if you can review please

### DIFF
--- a/binding/SkiaSharp/SKBitmap.cs
+++ b/binding/SkiaSharp/SKBitmap.cs
@@ -766,11 +766,19 @@ namespace SkiaSharp
 		public SKShader ToShader (SKShaderTileMode tmx, SKShaderTileMode tmy, SKSamplingOptions sampling) =>
 			ToShader (tmx, tmy, sampling, null);
 
+		[Obsolete ("Use ToShader(SKShaderTileMode tmx, SKShaderTileMode tmy, SKSamplingOptions sampling) instead.")]
+		public SKShader ToShader (SKShaderTileMode tmx, SKShaderTileMode tmy, SKFilterQuality quality) =>
+			ToShader (tmx, tmy, quality.ToSamplingOptions(), null);
+
 		public SKShader ToShader (SKShaderTileMode tmx, SKShaderTileMode tmy, SKMatrix localMatrix) =>
 			ToShader (tmx, tmy, SKSamplingOptions.Default, &localMatrix);
 
 		public SKShader ToShader (SKShaderTileMode tmx, SKShaderTileMode tmy, SKSamplingOptions sampling, SKMatrix localMatrix) =>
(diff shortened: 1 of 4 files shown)


Youssef1313:
Ping @mattleibow for review

mattleibow:
I am not sure how this would work. These are all new APIs that never existed in 2.x:

https://github.com/mono/SkiaSharp/blob/release/2.88.8/binding/Binding/SKBitmap.cs#L877-L884

What are you currently doing today? What version of SkiaSharp are you working with?

Youssef1313:
@mattleibow We are compiling against 2.x, but still want to support users who upgrade to 3.x via a compatibility layer we did using UnsafeAccessor.

Currently, we can't use UnsafeAccessor for `SKSamplingOptions` as it's a whole new type. All we can do is some ugly reflection that we want to avoid.

With this PR, assuming ported to 2.x, we will be able to use the new methods both on 2.x and 3.x

mattleibow:
How will this work in 2.x? Are you able to create a PR against release/2.x to show. Not sure 2.x has this arg so is it just a no-op?

Also, adding a test to make sure nothing blows up will also be great.

Thanks!

Youssef1313:
@mattleibow Okay I missed that this won't be possible to backport to 2.x, still will allow us to easily use UnsafeAccessor for 3.x specifically, and fallback to set quality on the paint when on 2.x

mattleibow:
OK, I see how this works. I think this is fine then, they are obsolete and removed from the public API. So, this should make it safe.

I see the 2.x version of skia had a more rudimentary form of the sampling options, and I contemplated adding that. But, rather add this than change the past and cause other issues.

mattleibow (approved):

Youssef1313:
@mattleibow Great! I assume the CI failures are not related to my PR? Not sure if they need to be fixed separately before this PR can be merged or not?

mattleibow:
The CI failures are a failure of CI. Still trying to figure it out. Every month it seems that CI needs some tweaks.

mattleibow (approved):

PR #2974, opened 2024-08-16
Disable Mac Catalyst just so we can go green

mattleibow:
**Description of Change**

It appears that the Mac Catalyst tests are now always crashing. We were green in the last build on July 30, but not anymore.

So sad.

I didn't touch it.


mattleibow:
After 1000 red PR builds, i just merged and now it is green. Looks like CI is just temperamental.

PR #2987, opened 2024-09-03
Added SKCanvas.DrawPoints with a count override to allow for partial …

### LINKED ISSUE #2986: [FEATURE] Passing DrawPoints a Partial Array

Sometimes it's useful to pass an array of points to `SKCanvas.DrawPoints` but not to draw all of them.  Currently there is no public method of accomplishing a partial array draw.

Add an override to `SKCanvas.DrawPoints` to allow for passing a count parameter to the `SkiaApi.sk_canvas_draw_points` method instead of just passing the entire array length as it currently does.

Currently the way to get around this is to create an array with the proper size ahead of time and allocate it on the heap but that requires allocation.  It would be even nicer to utilize Spans for this so they can be sliced, but I see multiple PRs #2617 & #2669 are stalled so I'm not sure about their integration.

### Additional context

_No response_

### Code of Conduct

DJGosnell:
…point array drawing.

**Description of Change**

Sometimes it's useful to pass an array of points to `SKCanvas.DrawPoints` but not to draw all of them.  The API currently doesn't accept Spans<T> at this point, so it is a stop-gap until that decision has been made.

No tests added currently as there is not currently a test in the `tests/Tests/SkiaSharp/SKCanvasTest.cs` file for drawing points.

**Bugs Fixed**

- Fixes #2986

**API Changes**

Add `public void SKCanvasDrawPoints (SKPointMode mode, SKPoint[] points, int count, SKPaint paint)` overload.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [] Has tests (if omitted, state reason in description)
- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/SkiaSharp/SKCanvas.cs
+++ b/binding/SkiaSharp/SKCanvas.cs
@@ -420,6 +420,17 @@ namespace SkiaSharp
 			}
 		}
 
+		public void DrawPoints (SKPointMode mode, SKPoint[] points, int count, SKPaint paint)
+		{
+			if (paint == null)
+				throw new ArgumentNullException (nameof (paint));
+			if (points == null)
+				throw new ArgumentNullException (nameof (points));
+			fixed (SKPoint* p = points) {
+				SkiaApi.sk_canvas_draw_points (Handle, mode, (IntPtr)count, p, paint.Handle);
+			}
+		}
+
 		// DrawPoint
 
 		public void DrawPoint (SKPoint p, SKPaint paint)


DJGosnell:
@dotnet-policy-service agree

mattleibow (commented):
This is good. Maybe also throw an out of range exception if the count is negative or greater than the length of the array. Maybe also have a start index? This way it can be used in a sort of batch draw? 

It may also be nice if you just added an ReadOnlySpan overload here too and then it will make the whole PR all nice. If you do that, then you can make both array overloads just use the span overload with a subset span of the whole array.

PR #3033, opened 2024-10-08
Add caching to DrawShapedText

### LINKED ISSUE #3032: [FEATURE] Add caching in DrawShapedText

The `DrawShapedText` method is meant to be an easy replacement for `DrawText` which adds text shaping via HarfBuzz but otherwise uses the same parameters as `DrawText`. It creates and disposes a `SKShaper` at every call which is very slow.

The `SKShaper` should not be created and disposed at every call. It should be cached and reused instead.

Furthermore also the result of the text shaping should be cached. This would improve the performance even more.

There is an overload of `DrawShapedText` which takes a `SKShaper` and does not create/dispose it at every call, but people need to have more insider knowledge of HarfBuzzSharp to use it. Also the caching of the results is something which everybody has to do to use HarfBuzz efficiently. It would be better if it was built in.

### Additional context

I wrote a [blog post](https://www.mrumpler.at/performance-tuning-drawshapedtext/) about this problem. It also includes benchmarks which show the possible improvement.

### Code of Conduct

MichaelRumpler:
**Description of Change**

This enables a user to set a cache duration. When set to a value > 0, it caches the `SKShaper` and the `SKShaper.Result` thus repeated calls with the same font and/or text are faster.

**Bugs Fixed**

- Fixes #3032

**API Changes**

Added: 
 
- `public static void SetShaperCacheDuration(this SKCanvas canvas, uint milliseconds)`

**Behavioral Changes**

If the cache duration was set to a value bigger than zero, `DrawShapedText` will cache the `SKShaper` and the `SKShaper.Result` for that amount of milliseconds. Therefore repeated calls with the same `SKFont` and/or `string` will be faster.

If the cache duration is set to 0, then the cache will be cleared, all contained `SKShaper` objects disposed and caching will be disabled.

The default cache duration is 0. So if `SetShaperCacheDuration` is never called, `DrawShapedText` works as before.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/CanvasExtensions.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/CanvasExtensions.cs
@@ -167,19 +167,25 @@ namespace SkiaSharp.HarfBuzz
 		{
 			var outdated = DateTime.Now - TimeSpan.FromMilliseconds(cacheDuration);
 
-			foreach (var kv in shaperCache.AsEnumerable())
+			lock (shaperCache)
 			{
-				if (kv.Value.cachedAt < outdated)
+				foreach (var kv in shaperCache.AsEnumerable())
 				{
-					if (shaperCache.Remove(kv.Key))
-						kv.Value.shaper.Dispose();
+					if (kv.Value.cachedAt < outdated)
+					{
+						if (shaperCache.Remove(kv.Key))
+							kv.Value.shaper.Dispose();
+					}
 				}
 			}
 
-			foreach (var kv in shapeResultCache.AsEnumerable())
+			lock (shapeResultCache)
 			{
-				if (kv.Value.cachedAt < outdated)
-					shapeResultCache.Remove(kv.Key);
+				foreach (var kv in shapeResultCache.AsEnumerable())
+				{
+					if (kv.Value.cachedAt < outdated)
+						shapeResultCache.Remove(kv.Key);
+				}
 			}
 
 			if ((shaperCache.Count == 0 && shapeResultCache.Count == 0) || cacheDuration == 0)
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/CanvasExtensions.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/CanvasExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
-using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 
 namespace SkiaSharp.HarfBuzz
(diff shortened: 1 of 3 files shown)


taublast:
Very nice one!! 

Possible to avoid `ConcurrentDictionary`? IMHO it affects very badly animations scenarios, triggering GC due to its allocation mechanics, thus creating messy lag spikes during animations. 
Once one starts a gaming loop that might come into play. 

Maybe add a `lock` to a `Dictionary`.. Just thinking.

mattleibow:
Is it possible to use `System.Runtime.Caching.MemoryCache`?

https://learn.microsoft.com/dotnet/api/system.runtime.caching.memorycache

Not sure if this is a good thing or not? Do people use this API? Is this sufficient?

MichaelRumpler:
I didn't want to introduce another dependency. But I can try it and see what the benchmarks say.

A `Dictionary` with `lock` was a bit faster than the `ConcurrentDictionary`, but not very much. What I actually planned, but didn't find the time to is to look at the source of System.Runtime.Caching and see if I can do something similar.

MichaelRumpler:
I tested `System.Runtime.Caching.MemoryCache` but it has several drawbacks:

- it uses a string as key
- it stores an `object` which must be boxed

Both make it slower than the `Dictionary<int, Tuple>`.

Here are the benchmarks with the cache variants:

| Method                           | Mean       | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|--------------------------------- |-----------:|----------:|----------:|-------:|-------:|----------:|
| DrawShapedAscii                  | 134.396 us | 0.4493 us | 0.4203 us | 0.2441 |      - |    3144 B |
| DrawAsciiWithShaper              |  19.046 us | 0.1644 us | 0.1283 us | 0.3052 |      - |    2744 B |
| DrawAsciiWithConcurrentCache     |   9.845 us | 0.0486 us | 0.0455 us | 0.0916 | 0.0763 |     848 B |
| DrawAsciiWithDictionaryCache     |   9.575 us | 0.0355 us | 0.0315 us | 0.0763 | 0.0610 |     752 B |
| DrawAsciiWithRuntimeCache        |  10.934 us | 0.0564 us | 0.0528 us | 0.1831 | 0.1526 |    1560 B |
| DrawShapedLigatures              | 191.784 us | 1.3686 us | 1.2133 us | 0.2441 |      - |    3144 B |
| DrawLigaturesWithShaper          |  72.422 us | 0.5261 us | 0.4921 us | 0.2441 |      - |    2744 B |
| DrawLigaturesWithConcurrentCache |   8.561 us | 0.0280 us | 0.0262 us | 0.0916 | 0.0763 |     848 B |
| DrawLigaturesWithDictionaryCache |   8.372 us | 0.0513 us | 0.0480 us | 0.0763 | 0.0610 |     752 B |
| DrawLigaturesWithRuntimeCache    |  10.120 us | 0.0264 us | 0.0206 us | 0.7019 | 0.6714 |    6072 B |
| DrawShapedEmojis                 |  62.447 us | 0.2296 us | 0.2035 us | 0.1221 |      - |    1888 B |
| DrawEmojisWithShaper             |  17.385 us | 0.0405 us | 0.0338 us | 0.1526 |      - |    1488 B |
| DrawEmojisWithConcurrentCache    |  13.649 us | 0.2212 us | 0.2069 us | 0.0916 | 0.0763 |     848 B |
| DrawEmojisWithDictionaryCache    |  13.553 us | 0.1254 us | 0.1173 us | 0.0763 | 0.0610 |     752 B |
| DrawEmojisWithRuntimeCache       |  15.427 us | 0.2149 us | 0.2010 us | 0.7019 | 0.6714 |    6120 B |
| DrawShapedRandom                 | 125.474 us | 1.1728 us | 1.0397 us |      - |      - |    1943 B |
| DrawRandomWithShaper             |  10.477 us | 0.0267 us | 0.0209 us | 0.1831 |      - |    1543 B |
| DrawRandomWithConcurrentCache    |   7.811 us | 0.0747 us | 0.0662 us | 0.1068 | 0.0992 |     911 B |
| DrawRandomWithDictionaryCache    |   7.538 us | 0.0872 us | 0.0773 us | 0.0916 | 0.0763 |     815 B |
| DrawRandomWithRuntimeCache       |   9.752 us | 0.1542 us | 0.1367 us | 0.1831 | 0.1678 |    1582 B |

The project with the benchmarks for the various variants is on https://github.com/MichaelRumpler/SkiaTextRendering.

As you can see the simple `Dictionary` with a `lock` was always fastest with the least amount of memory used. So I changed it to this now.

source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/CanvasExtensions.cs
```diff
@@ -99,7 +104,89 @@ public static void DrawShapedText(this SKCanvas canvas, SKShaper shaper, string
...
+			var key = HashCode.Combine(typeface.FamilyName, typeface.IsBold, typeface.IsItalic);
+
+			SKShaper shaper;
+			lock (shaperCache)
+			{
+				shaper = shaperCache.TryGetValue(key, out var value)
+					? value.shaper
+					: new SKShaper(typeface);
+
+				shaperCache[key] = (shaper, DateTime.Now);      // update timestamp
```
mattleibow, line 140: Can this not be done in a single step using `GetOrAdd`?

https://learn.microsoft.com/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-8.0#system-collections-concurrent-concurrentdictionary-2-getoradd(-0-system-func((-0-1)))

MichaelRumpler: I don't know how `GetOrAdd` could help, but I tried `AddOrUpdate` [here](https://github.com/MichaelRumpler/SkiaTextRendering/blob/f937f9c459c79cfdbc50a9290a010cdeb31936d8/SkiaTextRendering/Extensions/CanvasExtensionsConcurrentDict.cs#L98):

	var newTuple = shaperCache.AddOrUpdate(key,
		_ => (new SKShaper(typeface), DateTime.Now),
		(_, old) => (old.shaper, DateTime.Now));

But the [simple](https://github.com/MichaelRumpler/SkiaTextRendering/blob/f937f9c459c79cfdbc50a9290a010cdeb31936d8/SkiaTextRendering/Extensions/CanvasExtensions.cs#L99C1-L107C1) `Dictionary` with a `lock` is still faster.

	lock (shaperCache)
	{
		shaper = shaperCache.TryGetValue(key, out var value)
			? value.shaper
			: new SKShaper(typeface);

		shaperCache[key] = (shaper, DateTime.Now);      // update timestamp
	}

MichaelRumpler:
My last comment here was 18 months ago and it was never replied to. Life has moved on and I do not use SkiaSharp anymore. I also don't know how you got the results from #3915 
> (1) collection-modified-during-enumeration crash in ClearCache, (2) use-after-dispose race condition with timer thread, (3) coordinate handling semantic change needs verification.

So I won't work my way back into this issue. Use it or close it. I don't mind either way.

taublast:
Awesome direction. Please consider making this optional via some switch, since it would affect other caching built on top of SkiaSharp by third-party consumers.

Gillibald:
Why exactly is a built-in cache needed? DrawShapedText is just a convenience method. You should be using SkTextBlobBuilder and caching the built text blob. You could also introduce a helper that creates a SkTextBlob from a given string via SkShaper.

PR #3079, opened 2024-11-25
Fix flicker on canvas resize.

jrhodnik:


jrhodnik:
@dotnet-policy-service agree

Issue #3137, opened 2025-01-09
[FEATURE] Add `SkImageFilters::RuntimeShader`


kkwpsv:
### Is your feature request related to a problem?

I want to use `RuntimeShader` to overlay fragment shader effects.

### Describe the solution you would like

Exposed `SkImageFilters::RuntimeShader` in C# like other `SKImageFilter.CreateXXX`.

### Describe alternatives you have considered

None.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kkwpsv:
`SKRuntimeShaderBuilder` is implemented in C#.

taublast:
Speaking of alternatives, You could actually use SKPaint.Shader to achieve the same result, while the API to assign a shader to SKPaint.ImageFilter is missing.

mattleibow:
Thanks for the PR @kkwpsv. I added a review. I think the code is generally OK, but sort of I disagree with the way Google implemented all this API. I am thinking we should follow the way they did the other 3 (Shader, Blender, ColorFilter) just to have it all consistent. Not sure why they went so far out of the way.

mattleibow:
Superseded by #3776 which tracks the full SkSL image filter API. PR #3778 is in progress.

PR #3141, opened 2025-01-15
Add RuntimeShader Support.

kkwpsv:
**Description of Change**

Add RuntimeShader support.
See https://github.com/mono/SkiaSharp/issues/3137.

**Bugs Fixed**

None.

**API Changes**

Added: 
 
- `SKImageFilter SKImageFilter.CreateRuntimeShader(SKRuntimeShaderBuilder builder, string childShaderName, SKImageFilter? input)`
- `SKImageFilter SKImageFilter.CreateRuntimeShader (SKRuntimeShaderBuilder builder, float maxSampleRadius, string childShaderName, SKImageFilter? input)`
- `SKImageFilter SKImageFilter.CreateRuntimeShader (SKRuntimeShaderBuilder builder, float maxSampleRadius, string[] childShaderNames, SKImageFilter?[] inputs)`
- `SKRuntimeEffectUniforms (SKRuntimeEffectUniforms effectUniforms)`
- `SKRuntimeEffectChildren (SKRuntimeEffectChildren effectChildren)`
- `SKRuntimeEffectBuilder (SKRuntimeEffectBuilder builder)`
- `SKRuntimeShaderBuilder (SKRuntimeShaderBuilder builder)`

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/skia/pull/145

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/SkiaSharp.Resources/ResourcesApi.generated.cs
+++ b/binding/SkiaSharp.Resources/ResourcesApi.generated.cs
@@ -63,6 +63,7 @@ using sk_region_t = System.IntPtr;
 using sk_rrect_t = System.IntPtr;
 using sk_rtree_factory_t = System.IntPtr;
 using sk_runtimeeffect_t = System.IntPtr;
+using sk_runtimeshaderbuilder_t = System.IntPtr;
 using sk_shader_t = System.IntPtr;
 using sk_stream_asset_t = System.IntPtr;
 using sk_stream_filestream_t = System.IntPtr;
--- a/binding/SkiaSharp.SceneGraph/SceneGraphApi.generated.cs
+++ b/binding/SkiaSharp.SceneGraph/SceneGraphApi.generated.cs
@@ -63,6 +63,7 @@ using sk_region_t = System.IntPtr;
 using sk_rrect_t = System.IntPtr;
 using sk_rtree_factory_t = System.IntPtr;
 using sk_runtimeeffect_t = System.IntPtr;
+using sk_runtimeshaderbuilder_t = System.IntPtr;
 using sk_shader_t = System.IntPtr;
 using sk_stream_asset_t = System.IntPtr;
 using sk_stream_filestream_t = System.IntPtr;
--- a/binding/SkiaSharp.Skottie/SkottieApi.generated.cs
+++ b/binding/SkiaSharp.Skottie/SkottieApi.generated.cs
@@ -63,6 +63,7 @@ using sk_region_t = System.IntPtr;
 using sk_rrect_t = System.IntPtr;
 using sk_rtree_factory_t = System.IntPtr;
 using sk_runtimeeffect_t = System.IntPtr;
+using sk_runtimeshaderbuilder_t = System.IntPtr;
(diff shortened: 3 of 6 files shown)


kkwpsv:
@dotnet-policy-service agree

kkwpsv:
https://github.com/kkwpsv/AvaloniaSkiaRuntimeShaderDemo

kkwpsv:
Here is a demo.
https://github.com/kkwpsv/AvaloniaSkiaRuntimeShaderDemo

mattleibow (changes requested):
Thanks for this PR. This looks like a good feature to get in. Sorry about the long wait for a review @kkwpsv. 😞 

I know the C++ API has this different way to make the "shader", but it always felt a bit odd to have this in sort of separate locations when the Shader, Blender and ColorFilter were all in SKRuntimeEffect.

I wonder if instead of putting the code in SKImageFilter, we rather add a SKRuntimeImageFilterBuilder and in fact mirror most of what we do for SKShader but instead of using the word shader like in C++ we use the word ImageFilter. I would even go so far as to add a `CreateImageFilter` method that just calls into `CreateShader`. And then also make a `BuildImageFilter` that then calls that - which would mean it really is just a shader. The `Build()` method on this would then call the image filter code in the C API and return a new `SKImageFilter`

They dropped the different builders in C++, so we might avoid exposing this at all in the C API: https://github.com/google/skia/blob/main/include/effects/SkImageFilters.h#L458. I would think we could rather do what we did for all the others and just pass the values all in as args:

```cpp
sk_shader_t* sk_runtimeeffect_make_shader(sk_runtimeeffect_t* effect, sk_data_t* uniforms, sk_flattenable_t** children, size_t childCount, const sk_matrix_t* localMatrix)
```

They have this PR where they removed the builders and just have the base one: https://github.com/google/skia/commit/0b776ce0b8d647a1a7b243a3d1657d4ac9d602a1. I don't think we should do that as the separate builders are OK and also allow for a more fluent style code. But, that is also up for discussion.

Let me know what you think about these ideas!

kkwpsv:
> I wonder if instead of putting the code in SKImageFilter, we rather add a SKRuntimeImageFilterBuilder and in fact mirror most of what we do for SKShader but instead of using the word shader like in C++ we use the word ImageFilter. I would even go so far as to add a `CreateImageFilter` method that just calls into `CreateShader`. And then also make a `BuildImageFilter` that then calls that - which would mean it really is just a shader. The `Build()` method on this would then call the image filter code in the C API and return a new `SKImageFilter`

I think there is no problem with your idea.
What I am doing now is just to make the API close to C++.

mattleibow:
@kkwpsv are you able to make the changes?

taublast:
Need any help? This all sounds very exciting!

mattleibow:
Closing — superseded by #3778 which implements the full SkSL image filter API (ToImageFilter + SKRuntimeImageFilterBuilder) with the updated m147 Skia APIs. Thank you for the contribution!

PR #3248, opened 2025-04-24
[azure-pipelines-native] Add CI triggers for primary branches

pjcollins:
Restores the CI triggers to allow SkiaSharp-Native to run against main
and release branches.

### DIFF
--- a/scripts/azure-pipelines-native.yml
+++ b/scripts/azure-pipelines-native.yml
@@ -1,4 +1,7 @@
-trigger: none
+trigger:
+- main
+- dev*
+- release*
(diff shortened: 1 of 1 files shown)


pjcollins:
I updated the trigger settings in the new pipeline to test that things are configured correctly and https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=11468584&view=results kicked off automatically so I think the new pipeline should be good to go. I'll revert those changes now and remove the dev branch from the list of branches to build.

mattleibow (approved):

PR #3258, opened 2025-04-30
Fix crash in GRContext metal finalizer

### LINKED ISSUE #3178: [BUG] iOS: Crash when GC collects SKMetalView resources
### Description

SkiaSharp v3.116.1
.Net 9
maui                       9.0.0/9.0.100         SDK 9.0.100  
maui-ios                   9.0.0/9.0.100         SDK 9.0.100  
IDE: JetBrains Rider

In our app we have several views, one of which contains an SKMetalView. Every time the GC runs to collect resources, the app crashes with the following stack:

```
at <unknown> <0xffffffff>
	  at SkiaSharp.SkiaApi:gr_direct_context_abandon_context <0x00007>
	  at SkiaSharp.GRContext:AbandonContext <0x0010f>
	  at SkiaSharp.GRContext:DisposeNative <0x00083>
	  at SkiaSharp.SKNativeObject:Dispose <0x001c3>
	  at SkiaSharp.SKObject:Dispose <0x0007b>
	  at SkiaSharp.GRContext:Dispose <0x0007b>
	  at SkiaSharp.SKNativeObject:Finalize <0x000a7>
	  at System.Object:runtime_invoke_virtual_void__this__ <0x00087>
```

### Code

To reproduce the issue [SkiaSharpSample:](https://github.com/AlisiaNew/SkiaSharpSample)
1. Click Open Skia Screen, then go back
2. Click Open Text Screen, then go back
3. Repeat steps 1 and 2 and the app will crash

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Version of SkiaSharp

3.116.0 (Current)

2.88.9 (Previous)

### IDE / Editor

iOS

iOS 17.5
iOS 18.2

### Devices

_No response_

### Relevant Screenshots

_No response_

### Relevant Log Output

```shell
at <unknown> <0xffffffff>
	  at SkiaSharp.SkiaApi:gr_direct_context_abandon_context <0x00007>
	  at SkiaSharp.GRContext:AbandonContext <0x0010f>
	  at SkiaSharp.GRContext:DisposeNative <0x00083>
	  at SkiaSharp.SKNativeObject:Dispose <0x001c3>
	  at SkiaSharp.SKObject:Dispose <0x0007b>
	  at SkiaSharp.GRContext:Dispose <0x0007b>
	  at SkiaSharp.SKNativeObject:Finalize <0x000a7>
	  at System.Object:runtime_invoke_virtual_void__this__ <0x00087>
```

### Code of Conduct

jeremy-visionaid:
**Description of Change**

Use NSObject.DangerousRetain to prevent a double free of the command buffer

**Bugs Fixed**

- Fixes #3178

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


jeremy-visionaid:
@mattleibow Perhaps this should be handled in gr_direct_context_make_metal instead?

binding/SkiaSharp/GRContext.cs
```diff
@@ -91,6 +91,11 @@ public static GRContext CreateMetal (GRMtlBackendContext backendContext, GRConte
 			var device = backendContext.DeviceHandle;
 			var queue = backendContext.QueueHandle;
 
+#if __IOS__ || __MACOS__ || __TVOS__
+			using Foundation.NSObject queueObject = Foundation.NSObject.FromObject(backendContext.Queue);
+			queueObject.DangerousRetain ();
+#endif
+
```
mattleibow: For now, since we really oly control the life of the queue in the `SKMetalView`, can we move this to in the `Initialize` there?

tests/SkiaSharp.Tests.Devices/Tests/iOS/iOSExtensionsTests.cs
```diff
@@ -99,5 +101,21 @@ public void CGRectToSKRect(int x, int y, int w, int h)
 
 			Assert.Equal(expected, actual);
 		}
+
+		[SkippableFact]
+		public void GRContextDisposeDoesNotCrash()
```
mattleibow, line 106: For this test, we probably can do something linke this:

```cs
using var metalView = new SKMetalView();
((IMTKViewDelegate)metalView).Draw(metalView);
```

I know this is not as nice as what you have here, but for now let's keep this contained until we figure out the rules.

mattleibow (changes requested):
I am not 100% sure if we are supposed to retain, but according to instruments this does not leak. So I guess we are supposed to?

I see various places that have retain on there, so this is looking safe for now.

binding/SkiaSharp/GRContext.cs
jeremy-visionaid: I don't think that would be the right approach... We could fix the sketchy Skia constructor instead perhaps?

```C++
sk_sp<GrDirectContext> GrDirectContext::MakeMetal(void* device, void* queue,
                                                  const GrContextOptions& options) {
    GrMtlBackendContext backendContext = {};
    backendContext.fDevice.retain(device);
    backendContext.fQueue.retain(queue);

    return GrDirectContext::MakeMetal(backendContext, options);
}
```

That seems to me to be as close as we can get to a root cause fix (pending feedback from the Skia guys to confirm).

jeremy-visionaid: I've posted the above to your [Skia post](https://groups.google.com/g/skia-discuss/c/fYp8lLue4zU), but mine are pending approval still from yesterday and haven't show up.

mattleibow: I also added a comment. Hopefully they see one of them.

jeremy-visionaid: I was thinking of submitting a CL to Skia, but I see now that the deprecated function is already gone in m124. So, we'd have to change the C API implementation at some point anyway, which also fixes the problem:

```C++
gr_direct_context_t* gr_direct_context_make_metal(void* device, void* queue) {
#ifdef SK_METAL
    GrMtlBackendContext backendContext = {};
    backendContext.fDevice.retain(device);
    backendContext.fQueue.retain(queue);
    return ToGrDirectContext(GrDirectContext::MakeMetal(backendContext).release());
#else
    return nullptr;
#endif
}
```

Sorry for the noise there, I should have done my due diligence and spotted that sooner...

tests/SkiaSharp.Tests.Devices/Tests/iOS/iOSExtensionsTests.cs
jeremy-visionaid: I would prefer to focus on the root cause fix if that's OK. I'm not using SKMetalView, so I'm looking to fix the problem for more direct callers of the Metal backend.

jeremy-visionaid:
I've moved the corresponding fix for the double free to https://github.com/mono/skia/pull/159

mattleibow (approved):

PR #3291, opened 2025-06-13
SKAutoCanvasRestore avoid crash at Dispose when canvas was already disposed

taublast:
**Description of Change**

Have catched this few times myself: might be my bad coding style, but nevertheless `SKAutoCanvasRestore` should not crash app if the canvas inside was already disposed for any reason whatever. So checking to avoid that inside `SKAutoCanvasRestore` `Dispose` method.

**Bugs Fixed**

`SKAutoCanvasRestore` could crash if GC-ed/disposed after the containing canvas native Handler was already released while `canvas` field still not NULL and `Restore()` would be called, then crash app accessing Handle 0.

- Fixes

No known issues identified, still could be related to some unknown random crashes.

No Api changes, a small code change, please see 1 file change below.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/SkiaSharp/SKCanvas.cs
+++ b/binding/SkiaSharp/SKCanvas.cs
@@ -1065,7 +1065,10 @@ namespace SkiaSharp
 
 		public void Dispose ()
 		{
-			Restore ();
+			//canvas can be GC-ed before us
+			if (canvas != null && canvas.Handle != IntPtr.Zero) {
+				Restore ();
+			}				
 		}
 
 		/// <summary>


binding/SkiaSharp/SKCanvas.cs
```diff
@@ -1065,7 +1065,10 @@ public SKAutoCanvasRestore (SKCanvas canvas, bool doSave)
 
 		public void Dispose ()
 		{
-			Restore ();
+			//canvas can be GC-ed before us
+			if (canvas != null && canvas.Handle != IntPtr.Zero) {
```
mattleibow: Maybe move this code into Restore? The handle check. Otherwise calling restpore will still crash.

taublast: Yes why not, at the same time this check inside Dispose can be important for the future, if one day someone adds new code below Restore, and this one would crash. So it's like protectional for the future.

mattleibow (approved):

Issue #3284, opened 2025-06-02
[BUG] SKMetalView issues on older hardware (iOS 15)


pancirno:
### Description

Commit c2a1fde introduced changes related to simulator performance by including calls to DepthStencilStorageMode setter. DepthStencilStorageMode was added in iOS 16.0, which means older systems we'd like to support will crash on initializing this control.

Additionaly, with the said changes to the simulator performance, the class now calls to an obsoleted GRBackendRenderTarget constructor, which is weird.

### Code

Program crashes immediately on instancing the SKMetalView class.

### Expected Behavior

Program starts and renders things correctly in SKMetalView.

### Actual Behavior

Program crashes on startup with initialization of SKMetalView

### Version of SkiaSharp

3.119.0 (Current)

### Last Known Good Version of SkiaSharp

2.88.8 (Deprecated)

### IDE / Editor

Visual Studio (Windows)

### Platform / Operating System

iOS

### Platform / Operating System Version

iOS 15

### Devices

iPhone SE

### Relevant Log Output

```shell
`ObjCRuntime.ObjCException: 'Objective-C exception thrown.  Name: NSInvalidArgumentException Reason: -[SKMetalView setDepthStencilStorageMode:]: unrecognized selector sent to instance 0x10a255a80
Native stack trace:
	0   CoreFoundation                      0x00000001821a7c74 EEC1287D-059B-38C8-9BC1-58A0C8B1E6C2 + 601204
	1   libobjc.A.dylib                     0x00000001999d7ee4 objc_exception_throw + 56
	2   CoreFoundation                      0x00000001822786e8 EEC1287D-059B-38C8-9BC1-58A0C8B1E6C2 + 1455848
	3   UIKitCore                           0x00000001852ff20c 9D301877-2593-385C-8F72-F075AA0B48FA + 15397388
	4   CoreFoundation                      0x000000018214154c EEC1287D-059B-38C8-9BC1-58A0C8B1E6C2 + 181580
	5   CoreFoundation                      0x000000018214075c _CF_forwarding_prep_0 + 92
	6   XXXXXXXXXXXXXXXXXXX                 0x0000000104c56a7c xamarin_dyn_objc_msgSendSuper + 164
	7   XXXXXXXXXXXXXXXXXXX                 0x0000000104c0bee0 wrapper_managed_to_native_ObjCRuntime_Messaging_void_objc_msgSendSuper_UIntPtr_intptr_intptr_uintptr + 160
[...]

PR #3293, opened 2025-06-17
Hotfix metal for ios 15 and lower

taublast:
**Description of Change**

It has been reported that on lower iOS current metal itialiazation settings suggested by Apple are resulting in a crash, this hotfix solves this for iOS 15 and lower.

**Bugs Fixed**

- Fixes 

[[BUG] SKMetalView issues on older hardware (iOS 15)](https://github.com/mono/SkiaSharp/issues/3284)

**API Changes**

None.

**Behavioral Changes**

Should act as before .commit https://github.com/mono/SkiaSharp/commit/c2a1fde1bd3986d907f92445047b8fbc130c122b for devices with iOS 15 and lower.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/.github/workflows/label-with-ai.yml
+++ b/.github/workflows/label-with-ai.yml
@@ -3,34 +3,59 @@ name: Apply labels using AI
 on:
   issues:
     types: [opened, reopened]
+  workflow_dispatch:
+    inputs:
+      issue_number:
+        description: 'Issue number to triage'
+        required: true
+        type: number
 
 jobs:
   apply-label-and-add-comment:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       issues: write
+      models: read
+    env:
+      TRIAGE_AI_TOKEN: ${{ secrets.TRIAGE_AI_TOKEN }}
     steps:
-      - uses: mattleibow/dev-tools@action/label-with-ai
-        name: Apply AREA labels
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Determine AREA labels
+        uses: mattleibow/triage-assistant@v0
         with:
-          label-pattern: '^area/.*'
-          apply-label: 'false'
-          add-comment: 'false'
-      - uses: mattleibow/dev-tools@action/label-with-ai
-        name: Apply PLATFORM labels
+          issue: ${{ inputs.issue_number || github.event.issue.number }}
+          label-prefix: 'area/'
+          template: 'single-label'
+
+      - name: Determine PLATFORM labels
+        uses: mattleibow/triage-assistant@v0
         with:
-          label-pattern: '^os/.*'
-          apply-label: 'false'
-          add-comment: 'false'
-      - uses: mattleibow/dev-tools@action/label-with-ai
-        name: Apply BACKEND labels
+          issue: ${{ inputs.issue_number || github.event.issue.number }}
+          label-prefix: 'os/'
+          template: 'multi-label'
+
+      - name: Determine BACKEND labels
+        uses: mattleibow/triage-assistant@v0
         with:
-          label-pattern: '^backend/.*'
-          apply-label: 'false'
-          add-comment: 'false'
-      - uses: mattleibow/dev-tools@action/label-with-ai
-        name: Apply TENET labels
+          issue: ${{ inputs.issue_number || github.event.issue.number }}
+          label-prefix: 'backend/'
+          template: 'multi-label'
+
+      - name: Determine TENET labels
+        uses: mattleibow/triage-assistant@v0
         with:
-          label-pattern: '^tenet/.*'
-          apply-label: 'false'
-          add-comment: 'false'
+          issue: ${{ inputs.issue_number || github.event.issue.number }}
+          label-prefix: 'tenet/'
+          template: 'single-label'
+
+      - name: Apply all labels and add comment
+        uses: mattleibow/triage-assistant@v0
+        with:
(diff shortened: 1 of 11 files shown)


taublast:
We had SampleCount 1 before fixing metal for simulator, so in context of [info from Jeremy0(https://github.com/mono/SkiaSharp/pull/3156#issuecomment-2846719662) it could be safer to return to the state of SampleCount for phisical devices we had previously.  
I did not notice any difference on real device (not simulator) with SampleCount 1 or 2, so 1 is fine with me personally to revert to the way it was.

taublast:
I'm using this code several times a week on real device/real mac/real iphone, if this info could help..

jeremy-visionaid:
@taublast IIRC, the sample you provided for the performance measurements used a redundant intermediate blit (and also used a sample count of 1 on the intermediate surface). Have you fix those yet? It would be good to rule them out. Might be helpful to get the actual render timings as opposed to the FPS too.

In my use case I did not notice any significant performance difference in the simulator with the defaults (i.e. 1), and I still think that the docs do not recommend this.  If I'm reading it correctly, the docs and the sample just avoid using a sample count of 2 - they don't say not to use 1.

I'm happy to go and triple check my measurements, and I have an older MacBook I can test on too, but prob be a bit later in the day as I've got other more urgent bits I need to get to first.

taublast:
> @taublast IIRC, the sample you provided for the performance measurements used a redundant intermediate blit (and also used a sample count of 1 on the intermediate surface). Have you fixed those yet? It would be good to rule them out. Might be helpful to get the actual render timings as opposed to the FPS too.

@jeremy-visionaid 
The sample i provided in the first place was the current skiasharp handler, and my custom handlers appeared after the fix was already working, nothing to do with it. To answer direct about the handler, yes it is now better "timed" with more fluid rendering than what we had.

> In my use case I did not notice any significant performance difference in the simulator with the defaults (i.e. 1), and I still think that the docs do not recommend this. If I'm reading it correctly, the docs and the sample just avoid using a sample count of 2 - they don't say not to use 1.

Exact quoted examples from apple were followed both for simulator and real device. Simulator settings worked perfectly, on real devices we had an issue with iOS lower than 16 (thanks apple).  
This PR sets SampleCount to 1 for real devices as it was already the case previously, default is 1 and our code is not changing this.  
The "1" variable is there for an explicit set, to be visible and customizable.  
And I think we already had this interesting conversation between me "simulator went from unusable to awesome" and you "did not notice any significant performance difference", will not continue on this.

> I'm happy to go and triple check my measurements, and I have an older MacBook I can test on too, but prob be a bit later in the day as I've got other more urgent bits I need to get to first.

Please suite yourself, meanwhile I would not choose to revert the good to avoid the bad, instead of fixing the bad to achieve a better state.

jeremy-visionaid:
Hey, sorry, I didn't mean for my post to seem so terse - just got a lot of stuff on. I was only pointing out that there were other factors in the sample you provided me that might have been a contributing factor to the performance issues you experienced - I wasn't aware of the timeline of your own handlers to know whether that was important or not. I can only say that original PR didn't seem to affect the performance for my particular usage on my particular hardware - and I completely accept there might be something with either of those factors that make this PR the better one.

With the sample count set to 1 for real devices, I don't see that it's a particularly big deal either way about which PR gets accepted (just technical correctness and simulator will render slightly different results). I'd rather the attention go to higher priority stuff like #3258 since that's blocking for stuff like Avalonia than foresake the great for the good.

jeremy-visionaid:
OK, I've got some time now to look at this. I recall now that my initial investigation was based off https://github.com/taublast/Maui.Game.SpaceShooter
and
https://github.com/taublast/DrawnUi
in order to eliminate my use case being a factor.

I still had some of the modifications locally, so I've simplified them and pused those to a fork:
https://github.com/jeremy-visionaid/Maui.Game.SpaceShooter/tree/samplecount-test
https://github.com/jeremy-visionaid/DrawnUi/tree/samplecount-test
The timings are obviously very quick and dirty - I have more sophisticated timings in my own app, but I hope they would be sufficient to show if there were a significant performance issue related to the sample count in the simulator.

At the time, there was a bug in dotnet and it wasn't possible to compile a release build for the simulator without setting UseInterpreter:
https://github.com/dotnet/macios/issues/22205
This has now been resolved, so I've disabled UseInterpreter again.

Here are the timings with dotnet sdk 9.0.301:

M2 Pro - iPad A16 18.5 Simulator - Debug
SampleCount = 1: ~5.5 ms
SampleCount = 4: ~5.5 ms

M2 Pro - iPad A16 18.5 Simulator - Release
SampleCount = 1: ~5.5 ms
SampleCount = 4: ~5.5 ms

I also added a 5 ms sleep as a control, which increased the render timings to over 10 ms as expected. The shooter app maintained 60 fps for all the tests.

So, the test method is pretty blunt, please let me know if you disagree with it, but it is at least something that everyone here could look at and spin up as a test pretty quickly to inform a decision here.

jeremy-visionaid:
Urgh. I forgot to hit reload when I switched to the release build 🤦 It's too late here for me to repeat the timings now since release builds are so very slow.

In the meantime I tested my x86 machine...

MacBook Pro 2019 - iPad A16 18.5 Simulator - Debug
SampleCount = 1: ~15 ms
SampleCount = 4: ~16 ms

So it _seems_ sample count of 1 might be measurably faster there, which is probably what you'd expect, but one off timings are obviously pretty sketchy - it would really need an average to be more certain.

PR #3363, opened 2025-09-05
Move all docs into code comments

mattleibow:
**Description of Change**

We don't want or need the separate repo anymore.


binding/SkiaSharp/SKGraphics.cs
```diff
@@ -48,17 +50,20 @@ public static long GetResourceCacheTotalBytesUsed () =>
 		public static long GetResourceCacheTotalByteLimit () =>
 			(long)SkiaApi.sk_graphics_get_resource_cache_total_byte_limit ();
 
+		/// <param name="bytes"></param>
```
jeromelaban, line 53: Empty? Also, why not the others on the same file?

binding/SkiaSharp/SKImageInfo.cs
```diff
@@ -141,40 +261,61 @@ public readonly SKImageInfo WithSize (int width, int height)
...
+		/// <param name="newAlphaType">The alpha/transparency type.</param>
+		/// <returns>Returns the new <see cref="SKImageInfo" />.</returns>
 		public readonly SKImageInfo WithAlphaType (SKAlphaType newAlphaType)
 		{
 			var copy = this;
 			copy.AlphaType = newAlphaType;
 			return copy;
 		}
 
+		/// <param name="obj"></param>
```
jeromelaban, line 300: Same here? Empty on purpose?

binding/SkiaSharp/SKMatrix.cs
```diff
@@ -119,6 +148,8 @@ public readonly void GetValues (float[] values)
 		public static SKMatrix CreateIdentity () =>
 			new SKMatrix { scaleX = 1, scaleY = 1, persp2 = 1 };
 
+		/// <param name="x"></param>
```
jeromelaban, line 151: Same here, empty on purpose?

mattleibow:
Closing this for now as the automated tooling to process nugets do not support nugets with different apis in different tfms. For now, we still have to use a separate docs repo. I will continue to investigate, but right now these docs are out of date.

PR #3376, opened 2025-10-03
Disposing SKStreamAsset when asset MemoryBase is not IntPtr.Zero

michaldobrodenka:
Extension `public static Blob ToHarfBuzzBlob(this SKStreamAsset asset)` may leak when asset MemoryBase is not IntPtr.Zero.

After freeing memory, Dispose to SKStreamAsset (SKNativeObject) is added.

**Bugs Fixed**

- Discussion #3373

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
@@ -27,7 +27,7 @@ namespace SkiaSharp.HarfBuzz
 			{
 				var ptr = Marshal.AllocCoTaskMem(size);
 				asset.Read(ptr, size);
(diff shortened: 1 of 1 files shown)


michaldobrodenka:
@dotnet-policy-service agree


pi 3. 10. 2025 o 13:37 dotnet-policy-service[bot] ***@***.***>
napísal(a):

michaldobrodenka:
Fixes #3378

molesmoke:
It is kind of odd that `SKStreamAsset` is disposed on one path but not the other (and the caller doesn't know which). My two cents is that the problem is more that `SKShaper` doesn't dispose the intermediate `SKStreamAsset` from the `Typeface`. The leaking stream could already have been destroyed by the time the constructor exits (assuming that the 'Font' no longer requires access to the `Blob` - since it's written that way already).

PR #3418, opened 2025-11-11
Dev/perf testing

mattleibow:
This is a PR to investigate the crazy perf difference found here: https://github.com/wieslawsoltes/FastSkiaSharp

Discussion thread: https://x.com/wieslawsoltes/status/1984761177520685154

The native code is 120fps, but skiasharp is sub 10fps.

By first playing around found that I got to 40fps by using the new opengl flags in the SKGLView.

This all appears to be inside the skia code, which appears to be using similar compiler args.

My fork which downgrades skia to the same version in skiasharp is a tiny bit slower, 111-119 fps: https://github.com/mattleibow/FastSkiaSharp/tree/main_119

So the newer skia _is_ faster, but only a bit. There is still a significant difference between teh code in the c++ and the C#

image


ajitid:
Were we able to find out what is causing this perf difference? If yes, is it fixable?

Issue #3322, opened 2025-07-31
[BUG] Unhandled exception in Dispose() when show/hide multiple SKCanvasView in Blazor


jjzhang12:
### Description

It's not easy to reproduce that in a separated sample application, but it's consistantly reproducible in our Blaozor application which use multiple SkiaCanvasView to draw graphic. See actual behavior for the exception.

Also the Dispose() in SKCanvasView() does not follow proper .net Dispose pattern, it lacks a virtual Dispose(bool) method so the exception can be suppressed in the derived class.

### Code

N/A

### Expected Behavior

No exception and error message in Console from Skiasharp.

### Actual Behavior

When show and hide multiple SKCanvasView, an exception observed from Dispose() of SKCanvasView:

crit: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: TypeError: Failed to execute 'unobserve' on 'ResizeObserver': parameter 1 is not of type 'Element'.
Microsoft.JSInterop.JSException: TypeError: Failed to execute 'unobserve' on 'ResizeObserver': parameter 1 is not of type 'Element'.
 ---> TypeError: Failed to execute 'unobserve' on 'ResizeObserver': parameter 1 is not of type 'Element'.
   --- End of inner exception stack trace ---
   at Microsoft.JSInterop.WebAssembly.WebAssemblyJSRuntime.InvokeJS(String identifier, String argsJson, JSCallResultType resultType, Int64 targetInstanceId)
   at Microsoft.JSInterop.JSInProcessRuntime.Invoke[IJSVoidResult](String identifier, Int64 targetInstanceId, Object[] args)
   at Microsoft.JSInterop.Implementation.JSInProcessObjectReference.Invoke[IJSVoidResult](String identifier, Object[] args)
   at Microsoft.JSInterop.JSInProcessObjectReferenceExtensions.InvokeVoid(IJSInProcessObjectReference jsObjectReference, String identifier, Object[] args)
   at SkiaSharp.Views.Blazor.Internal.JSModuleInterop.Invoke(String identifier, Object[] args)
   at SkiaSharp.Views.Blazor.Internal.SizeWatcherInterop.Stop()
   at SkiaSharp.Views.Blazor.Internal.SizeWatcherInterop.OnDisposingModule()
   at SkiaSharp.Views.Blazor.Internal.JSModuleInterop.Dispose()
[...]

PR #3427, opened 2025-12-03
Avoid exception and console error message when SKCanvasView is created and then immediately disposed in Blazor

jjzhang12:
**Description of Change**

In JavaScript methods, return early when the target element is no longer present in the DOM, and remove the related error log for this scenario.

In Blazor, elements can be removed from the DOM before a component’s Dispose() method is invoked. This means that after any async call, the corresponding DOM element may already have been removed due to navigation. There’s no reliable way to distinguish whether a missing element is due to normal component disposal or an actual logic error. This change updates the code to simply stop execution when the element is absent and suppress error logging for this case.

Ref: https://github.com/dotnet/aspnetcore/issues/45777

**Bugs Fixed**

- Fixes #

https://github.com/mono/skiasharp/issues/3322

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


CullenStClair:
@mattleibow - Could this fix please be integrated ASAP? This bug causes notable usability issues since the errors it causes cannot be easily suppressed as a workaround to prevent the Blazor error UI from being displayed. Many thanks!

PR #3466, opened 2026-01-27
Don't dispose SKStreamAsset in ToHarfBuzzBlob

jeremy-visionaid:
**Description of Change**

HarfBuzz Blob should not have dispose ownership of SKStreamAsset as it might still be in use by the caller.

SKShaper should also deterministically dispose the stream.

**Bugs Fixed**

Discussion https://github.com/mono/SkiaSharp/discussions/3373

**API Changes**

None.

**Behavioral Changes**

ToHarfBuzzBlob no longer takes dispose ownership of the given SKStreamAsset

**Required skia PR**

None.

**PR Checklist**

- [X] Has tests (if omitted, state reason in description)
- [X] Rebased on top of main at time of PR
- [X] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
@@ -21,7 +21,7 @@ namespace SkiaSharp.HarfBuzz
 			var memoryBase = asset.GetMemoryBase();
 			if (memoryBase != IntPtr.Zero)
 			{
-				blob = new Blob(memoryBase, size, MemoryMode.ReadOnly, () => asset.Dispose());
+				blob = new Blob(memoryBase, size, MemoryMode.ReadOnly);
 			}
 			else
 			{
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/BlobExtensions.cs
@@ -26,8 +26,16 @@ namespace SkiaSharp.HarfBuzz
 			else
 			{
 				var ptr = Marshal.AllocCoTaskMem(size);
-				asset.Read(ptr, size);
-				blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
+				try
+				{
+					asset.Read(ptr, size);
+					blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
+				}
+				catch
+				{
+					Marshal.FreeCoTaskMem(ptr);
+					throw;
+				}
 			}
 
 			blob.MakeImmutable();
--- a/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/SKShaper.cs
+++ b/source/SkiaSharp.HarfBuzz/SkiaSharp.HarfBuzz/SKShaper.cs
@@ -17,7 +17,8 @@ namespace SkiaSharp.HarfBuzz
 			Typeface = typeface ?? throw new ArgumentNullException(nameof(typeface));
 
 			int index;
-			using (var blob = Typeface.OpenStream(out index).ToHarfBuzzBlob())
+			using (var stream = Typeface.OpenStream(out index))
+			using (var blob = stream.ToHarfBuzzBlob())
 			using (var face = new Face(blob, index))
 			{
 				face.Index = index;
--- a/tests/Tests/SkiaSharp/BlobExtensionsTest.cs
+++ b/tests/Tests/SkiaSharp/BlobExtensionsTest.cs
@@ -1,0 +1,38 @@
+using System.IO;
+using HarfBuzzSharp;
+using SkiaSharp.Tests;
+using Xunit;
+
+namespace SkiaSharp.HarfBuzz.Tests
+{
+	public class BlobExtensionsTest : SKTest
+	{
+		[SkippableFact]
+		public void ToHarfBuzzBlobDoesNotDisposeStream()
+		{
+			var fontFileName = Path.Combine(PathToFonts, "content-font.ttf");
+			using var tf = SKTypeface.FromFile(fontFileName);
+			using var stream = tf.OpenStream(out var index);
+			using (var blob = stream.ToHarfBuzzBlob())
+			{
+			}
+
+			Assert.False(stream.IsDisposed);
+		}
+
+		[SkippableFact]
+		public void ToHarfBuzzBlobBlobCanBeUsedOutsideStreamLifetime()
+		{
+			static Blob GetBlob(SKTypeface typeface)
+			{
+				using var stream = typeface.OpenStream(out var index);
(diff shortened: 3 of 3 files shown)


mattleibow (changes requested):
## PR Review: Don't dispose SKStreamAsset in ToHarfBuzzBlob (#3466)

Hey @jeremy-visionaid! Thanks for fixing this ownership issue. The change to not dispose the stream is correct — `ToHarfBuzzBlob()` shouldn't dispose a stream it doesn't own. 👍

### ⚠️ Note: Lifetime Requirement

The blob may reference the stream's memory directly (zero-copy for performance). **The caller must keep the stream alive for the lifetime of the blob.** This was always the case, but the previous code obscured it by having the blob dispose the stream.

@mattleibow — The docs for `ToHarfBuzzBlob` should mention that sometimes the stream data is not copied, and the stream must stay alive for the lifetime of the blob.

### ✅ Good changes

1. **SKShaper.cs** — Explicitly disposing the stream with `using` is proper resource management
2. **`ToHarfBuzzBlobDoesNotDisposeStream` test** — Good regression test

### 🟡 Add code comment about lifetime requirements

Please add a comment in the code explaining the lifetime contract:

```csharp
var memoryBase = asset.GetMemoryBase();
if (memoryBase != IntPtr.Zero)
{
    // The blob references the stream's memory directly (zero-copy).
    // The caller must keep the stream alive for the lifetime of the blob.
    blob = new Blob(memoryBase, size, MemoryMode.ReadOnly);
}
```

### 🔴 Exception safety issue in copy path

The `else` branch has a memory leak if `asset.Read()` or `new Blob()` throws:

```csharp
else
{
    var ptr = Marshal.AllocCoTaskMem(size);
    asset.Read(ptr, size);  // If this throws, ptr leaks
    blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
}
```

**Suggested fix:**

```csharp
else
{
    var ptr = Marshal.AllocCoTaskMem(size);
    try
    {
        asset.Read(ptr, size);
        blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
    }
    catch
    {
        Marshal.FreeCoTaskMem(ptr);
        throw;
    }
}
```

### 🔴 Test issue: `ToHarfBuzzBlobBlobCanBeUsedOutsideStreamLifetime`

This test demonstrates unsafe behavior. `SKTypeface.OpenStream()` returns a memory-backed stream with valid `memoryBase`, so the blob points directly to the stream's buffer. When the stream is disposed inside `GetBlob()`, the blob has a dangling pointer — the test passes by luck.

**Suggestion:** Remove this test as it violates the lifetime contract.

### 💡 Future enhancement: `bool copyData` parameter

Following existing SkiaSharp patterns (e.g., `SKMemoryStream(data, length, bool copyData)`), consider adding an overload:

```csharp
public static Blob ToHarfBuzzBlob(this SKStreamAsset asset, bool copyData = false)
```

When `copyData = true`, always copy to CoTaskMem regardless of `memoryBase`. This gives users a safe option when they don't want to manage stream lifetime. (Not required for this PR)

mattleibow (changes requested):
## PR Review: Don't dispose SKStreamAsset in ToHarfBuzzBlob (#3466)

Hey @jeremy-visionaid! Thanks for fixing this ownership issue. The change to not dispose the stream is correct — `ToHarfBuzzBlob()` shouldn't dispose a stream it doesn't own. 👍

### ⚠️ Note: Lifetime Requirement

The blob may reference the stream's memory directly (zero-copy for performance). **The caller must keep the stream alive for the lifetime of the blob.** This was always the case, but the previous code obscured it by having the blob dispose the stream.

@mattleibow — The docs for `ToHarfBuzzBlob` should mention that sometimes the stream data is not copied, and the stream must stay alive for the lifetime of the blob.

### ✅ Good changes

1. **SKShaper.cs** — Explicitly disposing the stream with `using` is proper resource management
2. **`ToHarfBuzzBlobDoesNotDisposeStream` test** — Good regression test

### 🟡 Add code comment about lifetime requirements

Please add a comment in the code explaining the lifetime contract:

```csharp
var memoryBase = asset.GetMemoryBase();
if (memoryBase != IntPtr.Zero)
{
    // The blob references the stream's memory directly (zero-copy).
    // The caller must keep the stream alive for the lifetime of the blob.
    blob = new Blob(memoryBase, size, MemoryMode.ReadOnly);
}
```

### 🔴 Exception safety issue in copy path

The `else` branch has a memory leak if `asset.Read()` or `new Blob()` throws:

```csharp
else
{
    var ptr = Marshal.AllocCoTaskMem(size);
    asset.Read(ptr, size);  // If this throws, ptr leaks
    blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
}
```

**Suggested fix:**

```csharp
else
{
    var ptr = Marshal.AllocCoTaskMem(size);
    try
    {
        asset.Read(ptr, size);
        blob = new Blob(ptr, size, MemoryMode.ReadOnly, () => Marshal.FreeCoTaskMem(ptr));
    }
    catch
    {
        Marshal.FreeCoTaskMem(ptr);
        throw;
    }
}
```

### 🔴 Test issue: `ToHarfBuzzBlobBlobCanBeUsedOutsideStreamLifetime`

This test is unsafe if the stream has a `memoryBase` (which `SKTypeface.OpenStream()` likely does). The blob would have a dangling pointer after the stream is disposed.

**Suggested fix:** Add an assert to ensure the test is actually exercising the copy path:

```csharp
[SkippableFact]
public void ToHarfBuzzBlobBlobCanBeUsedOutsideStreamLifetime()
{
    static Blob GetBlob(SKTypeface typeface)
    {
        using var stream = typeface.OpenStream(out var index);
        
        // Ensure this test is exercising the copy path, not the zero-copy path
        Assert.Equal(IntPtr.Zero, stream.GetMemoryBase());
        
        return stream.ToHarfBuzzBlob();
    }

    var fontFile = Path.Combine(PathToFonts, "content-font.ttf");
    using var tf = SKTypeface.FromFile(fontFile);
    using var blob = GetBlob(tf);
    Assert.Equal(1, blob.FaceCount);
}
```

If the assert fails, the test should be removed as it would be testing undefined behavior.

### 💡 Future enhancement: `bool copyData` parameter

Following existing SkiaSharp patterns (e.g., `SKMemoryStream(data, length, bool copyData)`), consider adding an overload:

```csharp
public static Blob ToHarfBuzzBlob(this SKStreamAsset asset, bool copyData = false)
```

When `copyData = true`, always copy to CoTaskMem regardless of `memoryBase`. This gives users a safe option when they don't want to manage stream lifetime. The test could then become:

```csharp
[SkippableFact]
public void ToHarfBuzzBlobWithCopyCanBeUsedOutsideStreamLifetime()
{
    static Blob GetBlob(SKTypeface typeface)
    {
        using var stream = typeface.OpenStream(out var index);
        return stream.ToHarfBuzzBlob(copyData: true);  // Force copy, safe to outlive stream
    }

    var fontFile = Path.Combine(PathToFonts, "content-font.ttf");
    using var tf = SKTypeface.FromFile(fontFile);
    using var blob = GetBlob(tf);
    Assert.Equal(1, blob.FaceCount);
}
```

(Not required for this PR)

PR #3560, opened 2026-03-12
Bump SkiaSharp to milestone 132

ramezgerges:
**Description of Change**

This PR updates SkiaSharp to milestone 132 on top of the companion Skia bump in https://github.com/mono/skia/pull/171, while keeping the managed-side changes focused on the surface area that actually moved with the upstream update.

In the spirit of the original community PR, the main changes here are:

- update the generated interop and native definitions for the m132 C API
- add the new `SKColorType` values exposed by upstream: `Bgra10101010Xr` and `RgbF16F16F16x`
- align enum/struct mappings and Vulkan/JPEG native layouts with the upstream ABI changes
- update the native build/versioning scripts for the new milestone
- refresh the affected tests so the newly exposed color types and related behavior stay covered

**Review Links**

- Companion Skia PR: https://github.com/mono/skia/pull/171
- AI review gist: https://gist.github.com/mattleibow/b1c57a78d86cc5dec3fff1182de48f01
- Rendered HTML review: https://gistcdn.githack.com/mattleibow/b1c57a78d86cc5dec3fff1182de48f01/raw/295bb903e36917d3faf80f74ba4d3ab06adcb6ec/171.html

**Bugs Fixed**

- Updates SkiaSharp to Skia milestone 132 and keeps the bindings, tests, and build definitions aligned with the new native ABI.

**API Changes**

### Type Changed: `SkiaSharp.SKColorType`

Added values:

```csharp
Bgra10101010Xr = 25,
RgbF16F16F16x = 26,
```

These follow the upstream Skia additions for HDR / wide-gamut image workflows. As part of that change, the native `SKColorType` numbering shifts accordingly and the generated bindings/tests were refreshed to match.

### Type Changed: `SkiaSharp.SKJpegEncoderOptions`

The native layout now includes the upstream EXIF orientation fields introduced in m132.

### Type Changed: `SkiaSharp.GRVkBackendContextNative` and `SkiaSharp.GrVkYcbcrConversionInfo`

These native structs were updated to match the upstream Vulkan shape changes in Skia.

**Behavioral Changes**

- Default font resolution now follows the newer upstream post-`SkFontMgr::RefDefault()` behavior through the Skia shim updates in the companion PR.
- Image, codec, and filter behavior now inherits the m132 Skia changes, including newer color/image pipeline support and orientation-aware decode paths.
- The earlier temporary changelog draft has been removed so the reviewer-facing summary lives here and in the linked review report instead of duplicating stale notes.

**Required skia PR**

Requires https://github.com/mono/skia/pull/171

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


mattleibow:
This PR is looking good so far. I had to make some tweaks to the DEPS - mostly commenting out the deps we are not using - as CI was running out of space. Somehow after this, there is 12 GB free still. So the deep clones and all the unused took up all the space.

I saw you were reworking the `SKManagedStream` and this is good. Before things might not have always worked, but now it should be all good. However, I think back in the day, I somehow had the ideea that `fork()` and `duplicate()` need to return the same type. I asked Copilot for an opinion and it seemed to think that there is no technical reason why we should need to do that. 

One big advantage is that we basically can hand off the data to native code and then not need a managed instance as well as not need to go through managed interop.

Copilot suggested something like this:

```csharp
private Stream? stream;

// Lazily-created snapshot for duplicate/fork. Stored in native memory
// via SKData (ref-counted), so multiple duplicates share one copy with
// zero managed allocations per duplicate.
private SKData? snapshotData;

protected override void DisposeManaged ()
{
	if (snapshotData is not null) {
		snapshotData.Dispose ();
		snapshotData = null;
	}

	if (disposeStream && stream is not null) {
		stream.Dispose ();
		stream = null;
	}

	base.DisposeManaged ();
}

private SKData? GetOrCreateSnapshot ()
{
	if (snapshotData is not null)
		return snapshotData;

	var pos = stream!.Position;
	stream.Position = 0;
	snapshotData = SKData.Create (stream, stream.Length);
	stream.Position = pos;

	return snapshotData;
}

protected internal override IntPtr OnDuplicate ()
{
	if (!stream!.CanSeek)
		return IntPtr.Zero;

	var data = GetOrCreateSnapshot ();
	if (data is null)
		return IntPtr.Zero;

	// Return a native SkMemoryStream backed by the cached SKData.
	// The native constructor ref-counts the SKData, so the data
	// stays alive as long as any stream references it. No managed
	// wrapper, no GCHandle, no per-duplicate byte[] allocation.
	return SkiaApi.sk_memorystream_new_with_skdata (data.Handle);
}

protected internal override IntPtr OnFork ()
{
	if (!stream!.CanSeek)
		return IntPtr.Zero;

	var data = GetOrCreateSnapshot ();
	if (data is null)
		return IntPtr.Zero;

	var handle = SkiaApi.sk_memorystream_new_with_skdata (data.Handle);
	SkiaApi.sk_stream_seek (handle, (IntPtr)stream.Position);
	return handle;
}
```

But if we really want to not return deifferent types, we could do this:

```cs
protected internal override IntPtr OnDuplicate ()
{
   if (!stream!.CanSeek)
       return IntPtr.Zero;

   var data = GetOrCreateSnapshot ();
   if (data is null)
       return IntPtr.Zero;

   // Wrap the native SKData as an UnmanagedMemoryStream, then as SKManagedStream.
   // The SKDataStream reads directly from native memory — no byte[] copy.
   // disposeHost: false — snapshotData owns the lifetime.
   var dataStream = data.AsStream (false);
   var newStream = new SKManagedStream (dataStream, true);
   return newStream.Handle;
}
```

Since the `OnDuplicate` and `OnFork` are really only called by native code, this should be fine and also should allow us to basically skip all the work.

What are your thoughts on this? Have you seen a need to keep the same type or foresee in issues? Not saying this is better or anything, just wanted to see what you think. 

Another thing, I am wondering is this stream change should just go into a separate PR? this way we don't change code and bump in one go. Not bad to have it all, I think I am just being nitpicky.

binding/SkiaSharp/Definitions.cs
```diff
@@ -60,6 +60,8 @@ public enum SKColorType
 		Srgba8888 = 22,
 		R8Unorm = 23,
 		Rgba10x6 = 24,
+		Bgra10101010Xr = 25,
```
mattleibow: Would this be more consistent with the others, seeing `Bgr101010xXR`

```suggestion
		Bgra10101010XR = 25,
```

binding/SkiaSharp/EnumMappings.cs
```diff
@@ -68,6 +68,8 @@ internal static SKColorTypeNative ToNative (this SKColorType colorType) =>
 				SKColorType.Bgr101010xXR => SKColorTypeNative.Bgr101010xXr,
 				SKColorType.Srgba8888 => SKColorTypeNative.Srgba8888,
 				SKColorType.R8Unorm => SKColorTypeNative.R8Unorm,
+				SKColorType.Bgra10101010Xr => SKColorTypeNative.Bgra10101010Xr,
+				SKColorType.RgbF16F16F16x => SKColorTypeNative.RgbF16F16F16x,
```
mattleibow: I think this has a case mismatch:

```suggestion
				SKColorType.RgbF16F16F16x => SKColorTypeNative.RgbF16f16f16x,
```

mattleibow (commented):
I think this PR is ready. I will give another review next week when my brain is fresh. But besides the enums (and maybe the stream) here I think is good to go.

mattleibow:
I am trying the new stream things in a separate PR which we can use to confirm the logic works on older skia - then this PR can be a pure milestone bump: https://github.com/mono/SkiaSharp/pull/3589

PR #3730, opened 2026-04-24
Fix #3693: move default-typeface resolution to the managed layer

### LINKED ISSUE #3693: [Android] SIGSEGV in sk_typeface_ref_default on Android API 36 (Android 16)
## Summary

SkiaSharp crashes with a `SIGSEGV` (signal 11) during app startup on Android API 36 (Android 16). The crash occurs in `sk_typeface_ref_default` when `SKObject` static initialization calls through to the native default typeface/font-manager path. This blocks all Android test execution on API 36.

## Crash Signature

```
Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8
    #01 pc 00000000003654ec  libSkiaSharp.so (sk_typeface_ref_default+44)
```

- **Signal:** 11 (SIGSEGV)
- **Code:** SEGV_MAPERR (invalid memory access)
- **Fault address:** `0x8` (null pointer  offset 8 from a null struct pointer)dereference 
- **Thread:** `.NET TP Worker` (background thread pool worker during MAUI/SkiaSharp initialization)

## Affected Platforms

| Platform | Arch | API | Result |
|----------|------|-----|--------|
| Android emulator (CI) | x86_64 |  SIGSEGV |36 | 
| Android emulator (local) | arm64-v8a |  SIGSEGV |36 | 
| Android emulator (CI, previous) | x86_64 | 33 Works | | 
| All non Works |- | Android  | platforms | 

### Native call chain

```
C#: SKObject..cctor()
 SKTypeface.EnsureStaticInstanceAreInitialized()
 SKTypeface..cctor()
 SkiaApi.sk_typeface_ref_default()

Native: sk_typeface_ref_default()                    // sk_typeface.cpp:53
 sk_get_default_typeface()                        // sk_default_fontmgr.h:23
 sk_get_default_fontmgr()->legacyMakeTypeface("", SkFontStyle::Normal())
 SkFontMgr_New_Android(nullptr)               // sk_default_fontmgr.cpp:36
```

### What goes wrong

1. `sk_get_default_fontmgr()` creates an Android font manager via `SkFontMgr_New_Android( this succeeds.nullptr)` 
2. `legacyMakeTypeface("", SkFontStyle::Normal())` is called on the Android font  on API 36, this returns a **null** `sk_sp<SkTypeface>`.manager 
3. Back in `sk_typeface_ref_default()`, the code does:
   ```cpp
   static sk_sp<SkTypeface> defaultTf = sk_get_default_typeface();
   defaultTf-> CRASH: defaultTf is nullref();  // 
   return ToTypeface(defaultTf.get());
   ```
4. The unconditional `->ref()` dereferences the null pointer, causing the SIGSEGV at fault address `0x8` (the ref-count field offset).

### Why API 36?

Android 16 (API 36) likely changed font system behavior such that `legacyMakeTypeface("", Normal)` no longer returns a fallback typeface when given an empty family name. This worked on API 33 and earlier.

### Local (ARM64 emulator on macOS)

```bash
# Requires: API 36 emulator image, SkiaSharp repo with matching native/managed versions
dotnet cake --target=tests-android --configuration=Release   --device=android-emulator-64 --deviceVersion=36 --skipExternals="all"
```

**XHarness exit code:** `80 (APP_CRASH)`

### CI

- **Build 156765** (x86_64): SIGSEGV in `sk_typeface_ref_default+97`
- **Build 156767** (x86_64, with lazy-init workaround): APP_CRASH exit code 80
- Pipeline: SkiaSharp (Public), job: Android (Linux), PR #3676

## Potential Fixes

1. **Null guard in `sk_typeface_ref_ check if `defaultTf` is null before calling `->ref()`, return nullptr instead of crashing:default`** 
   ```cpp
   sk_typeface_t* sk_typeface_ref_default(void) {
       static sk_sp<SkTypeface> defaultTf = sk_get_default_typeface();
       defaultTf->ref();
       return ToTypeface(defaultTf.get());
   }
   ```
   Plus corresponding null handling in the C# SKTypeface static constructor.

2. **Investigate `legacyMakeTypeface` on API  determine why `SkFontMgr_New_Android(nullptr)->legacyMakeTypeface("", Normal)` returns null. The Android font manager may need a non-empty family name or a different API path on API 36.36** 

3. ** add the null guard as a defensive fix, and separately investigate the upstream Android font manager behavior.Both** 

## Related

- PR # Android CI improvements (disk space, emulator hardening)3676 
- Known issue # Android DEVICE_NOT_FOUND CI failures (resolved by disk space fix in #3676)3610

ramezgerges:
**Description of Change**
- Fixes a `SIGSEGV` at SkiaSharp static initialization on Android API 36 (Android 16). Any app that touched an `SKObject`-derived type crashed with `Fatal signal 11 (SIGSEGV)` at `sk_typeface_ref_default+44`.
- Migrates the managed `SKTypeface` factories to go through `SKFontManager.Default` instead of C API functions that all went through the same internal `sk_get_default_fontmgr` helper.
- Drops the corresponding P/Invokes. **ABI break** — requires a SkiaSharp major version bump.

**Bugs Fixed**

- Fixes https://github.com/mono/SkiaSharp/issues/3693

**API Changes**

### `SKTypeface.cs`

- `FromFamilyName(familyName, style)` → `SKFontManager.Default.MatchFamily(familyName, style) ?? SKTypeface.Default`.
  - Behavioral improvement: when `MatchFamily` would return null (Android API 36 with empty family name, typo'd family names on some platforms), callers now receive `SKTypeface.Default` instead of null. No in-repo caller ever checked for a null return, so the practical impact is "no more silent NREs downstream of `FromFamilyName`."
- `FromFile(path, index)` → `SKFontManager.Default.CreateTypeface(path, index)`.
- `FromStream(Stream, index)` / `FromStream(SKStreamAsset, index)` → `SKFontManager.Default.CreateTypeface(stream, index)`.
- `FromData(data, index)` → `SKFontManager.Default.CreateTypeface(data, index)`.
- Static constructor now resolves the default typeface entirely in managed code:
  ```csharp
  static SKTypeface ()
  {
      defaultTypeface = new SKTypefaceStatic (CreateDefaultHandle ());
  }

  public static SKTypeface CreateDefault () => GetObject (CreateDefaultHandle ());

  private static IntPtr CreateDefaultHandle ()
  {
      var style = SKFontStyle.Normal;
      var handle = SkiaApi.sk_fontmgr_match_family_style (
          SKFontManager.Default.Handle, IntPtr.Zero, style.Handle);
      return handle == IntPtr.Zero ? SkiaApi.sk_typeface_create_empty () : handle;
  }
  ```
- `SKTypefaceStatic` nested class retained — it's the marker type the `GarbageCleanupFixture` in the test suite allowlists for static singletons.

### `SKFontManager.cs`

- Static constructor now calls `sk_fontmgr_create_default` (fresh allocation + managed-side pin) instead of the removed `sk_fontmgr_ref_default`. `SKFontManager.Default` is the sole cache of the default font manager.

### `SKFont.cs`

- Both constructors always pass a non-null typeface handle to the native `sk_font_new_with_values`:
  - Parameterless `new SKFont()` defers to `SKFont(null, …)`, which resolves to `SKTypeface.Default.Handle`.
  - `new SKFont(typeface, …)` coalesces a null `typeface` to `SKTypeface.Default`.
- Enables removing `sk_font_new` from the C ABI entirely.

### P/Invokes removed (via regen of `SkiaApi.generated.cs`)

- `sk_typeface_create_default`
- `sk_typeface_ref_default`
- `sk_typeface_create_from_name`
- `sk_typeface_create_from_file`
- `sk_typeface_create_from_stream`
- `sk_typeface_create_from_data`
- `sk_fontmgr_ref_default`
- `sk_font_new`

### `binding/libSkiaSharp.json`

- Removed the `sk_typeface_create_from_name` marshal hint (the function is gone).

### `externals/skia`

- Submodule pointer bumped to the companion skia PR. See **mono/skia#TBD** for the full native-side changeset.

## Breaking changes

1. **ABI break in `libSkiaSharp.so`/`.dll`/`.dylib`** — 7 previously-exported `SK_C_API` symbols removed. External C/C++ consumers need to migrate to `sk_fontmgr_*` entry points. Documented in the native PR.
2. **`SKPaint()` default constructor** — the backing `SkCompatPaint`'s default `SkFont` now wraps `SkTypeface::MakeEmpty()` instead of the platform default family. Text rendering through `SKPaint` (i.e. `SKPaint.Typeface`, `SKPaint.TextSize`, etc.) is already `[Obsolete]` with the message "Use SKFont instead"; modern callers pass an explicit `SKFont` that is always constructed with a real typeface. Apps that still rely on the obsolete surface and depend on implicit-default-font behavior will render no glyphs until they migrate to `SKFont`. This is deliberate for the major bump.
3. **`SKTypeface.FromFamilyName` with null/empty family** now always returns a usable typeface instead of potentially null. This is a behavior improvement, not a regression.

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/skia/pull/195

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


Gillibald:
Why is it a behavior improvement when SKTypeface.FromFamilyName duddently always succeeds? Handle null values instead of SKTypeface.Default. The same is valid for MatchFamily.

mattleibow:
# PR #3730 Review — Behavioral Analysis

Thanks for working on this @ramezgerges — the Android API 36 null-typeface issue definitely needs fixing. I've done a deep analysis of the behavioral changes in this PR against `release/3.119.x` and the upstream Skia m132 C++ code.

## TL;DR

The core goal (making `SKTypeface.Default` resilient to Android API 36 returning null) is good. Most of the changes are correct. One item to discuss:

1. ✅ **`FromFamilyName` returning `?? Default`** — this is actually fine and makes behavior consistent across platforms (see analysis below)
2. ⚠️ **`SKFont(null)` forces `SKTypeface.Default`** — unnecessary because Skia m132 already handles null internally

## `FromFamilyName` with `?? Default` — this is correct ✅

```csharp
// Current PR (line 70-71 of SKTypeface.cs)
public static SKTypeface FromFamilyName(string familyName, SKFontStyle style) =>
    SKFontManager.Default.MatchFamily(familyName, style) ?? Default;
```

I initially flagged this as a concern, but after deeper analysis of the m119 C++ `legacyMakeTypeface` behavior across all platforms:

| Platform | m119 `FromFamilyName("missing")` | Returned null? |
|---|---|---|
| macOS | `legacyMakeTypeface` → falls back to "Lucida Sans" | **Never null** |
| Windows | `legacyMakeTypeface` → DirectWrite fallback | **Never null** |
| Linux | `legacyMakeTypeface` → `matchFamilyStyle(nullptr)` fallback | **Never null** |
| Custom/WASM | `legacyMakeTypeface` → `fDefaultFamily->matchStyle()` | **Never null** |
| Android | `onMatchFamilyStyle` only, no fallback | **Null on miss** |

Android was the **only** platform where `FromFamilyName("missing")` returned null. The `?? Default` in the PR actually makes Android consistent with every other platform. This is the right call — `FromFamilyName` is the convenience API. Users who want strict match-or-null should use `SKFontManager.Default.MatchFamily()` directly.

## Issue: `SKFont(null)` — let C++ handle it ⚠️

```csharp
// Current PR
public SKFont(SKTypeface typeface, float size = DefaultSize, ...)
    : this(SkiaApi.sk_font_new_with_values((typeface ?? SKTypeface.Default).Handle, ...))
```

In Skia m132, `SkFont` **already handles null typefaces** at the C++ level:

```cpp
SkFont::SkFont(sk_sp<SkTypeface> face, SkScalar size, SkScalar scaleX, SkScalar skewX)
    : fTypeface(std::move(face)), ... {
    if (!fTypeface) {
        fTypeface = SkTypeface::MakeEmpty();  // ← C++ auto-handles null
    }
}
```

And `setTypeface(nullptr)` does the same:

```cpp
void SkFont::setTypeface(sk_sp<SkTypeface> tf) {
    fTypeface = std::move(tf);
    if (!fTypeface) {
        fTypeface = SkTypeface::MakeEmpty();
    }
}
```

Furthermore, m132 **removed** `getTypefaceOrDefault()` and `refTypefaceOrDefault()` entirely — `getTypeface()` now asserts non-null because the constructor guarantees it. The C# `font.Typeface` getter calls through `sk_font_get_typeface` → `refTypeface()` which is guaranteed non-null in m132.

This means:
- Passing `IntPtr.Zero` to `sk_font_new_with_values` → C++ sets `MakeEmpty()` → `font.Typeface` getter returns a valid (empty) typeface
- **No managed-side intervention needed**

The managed `?? SKTypeface.Default` forces the **platform default** (Helvetica on macOS, etc.) instead of letting C++ use the **empty typeface**. These are semantically different — the empty typeface is an explicit "no font" sentinel that draws nothing, while the platform default is a real font with glyphs.

### Recommendation

Revert to the release behavior:

```csharp
public SKFont(SKTypeface typeface, float size = DefaultSize, ...)
    : this(SkiaApi.sk_font_new_with_values(typeface?.Handle ?? IntPtr.Zero, size, scaleX, skewX), true)
```

And similarly, `SKFont()` can just pass null:

```csharp
public SKFont()
    : this(null, DefaultSize, DefaultScaleX, DefaultSkewX)
{
}
```

Since `sk_font_new_with_values(null, ...)` → C++ `SkFont(nullptr, ...)` → auto-sets `MakeEmpty()`, this is safe and matches the C++ contract.

## What's good in this PR

- **`SKTypeface.Default` implementation** (`matchFamilyStyle(null, Normal)` with `create_empty()` fallback) — this is the right approach for the Android API 36 fix
- **`SKTypeface.CreateDefault()`** — same, good
- **`FromFamilyName` with `?? Default`** — makes Android consistent with all other platforms
- **`FromFile/FromStream/FromData` delegation to `SKFontManager`** — cleaner, removes duplicate code, null behavior preserved
- **`SKFontManager.Default` using `create_default()` instead of `ref_default()`** — fine
- **Removing the old C APIs** that no longer exist in m132 — necessary cleanup

## Summary

| Item | Current PR | Assessment |
|------|-----------|-----------|
| `FromFamilyName ?? Default` | `MatchFamily(...) ?? Default` | ✅ Keep — makes Android consistent with other platforms |
| `SKFont(null)` | `(null ?? Default).Handle` | ⚠️ Revert to `typeface?.Handle ?? IntPtr.Zero` — let C++ handle it |
| `SKFont()` | `this(null, ...)` → hits above | ⚠️ Same — flows through correctly once fixed |
| Everything else | ✅ | Keep as-is |

mattleibow:
@Gillibald I think I agree with tyou on your comment. Skia has changed a fair bit so this is getting more tricky as there is now no such thing as a "default typeface" and "null typefaces" are almost gone.

mattleibow:
## Follow-up: `SKFont(null)` and the empty typeface

One more thought on the `SKFont` constructor. In Skia m132, `SkFont(nullptr)` no longer stores null — it stores `SkTypeface::MakeEmpty()` (a singleton empty typeface with 0 glyphs that draws nothing). This is an upstream change we can't avoid.

The current PR works around this by forcing `SKTypeface.Default` at the C# layer:

```csharp
public SKFont(SKTypeface typeface, ...)
    : this(SkiaApi.sk_font_new_with_values((typeface ?? SKTypeface.Default).Handle, ...))
```

But this changes the semantic — `new SKFont().Typeface` used to return `null` (meaning "no typeface explicitly set"), now it returns the platform default. On m132, if we just pass null through, C++ sets `MakeEmpty()` and the getter returns a valid empty typeface instead. Either way it's a behavior change from m119, but returning the empty typeface is more honest — it's what C++ actually holds.

### Suggested approach: let C++ handle null, expose `Empty` and `IsEmpty` so users can check

**SKTypeface** — cache the empty typeface alongside Default, expose `Empty` and `IsEmpty`:

```csharp
private static readonly SKTypeface defaultTypeface;
private static readonly SKTypeface emptyTypeface;

static SKTypeface ()
{
    defaultTypeface = new SKTypefaceStatic (CreateDefaultHandle ());
    emptyTypeface = new SKTypefaceStatic (SkiaApi.sk_typeface_create_empty ());
}

public static SKTypeface Default => defaultTypeface;

public static SKTypeface Empty => emptyTypeface;

public bool IsEmpty => Handle == emptyTypeface.Handle;
```

`MakeEmpty()` returns a C++ singleton (`SkNoDestructor<SkEmptyTypeface>`), so the handle is stable and this comparison is reliable.

**SKFont constructor** — just pass through, let C++ do its thing:

```csharp
public SKFont ()
    : this (null, DefaultSize, DefaultScaleX, DefaultSkewX)
{
}

public SKFont (SKTypeface typeface, float size = DefaultSize, ...)
    : this (SkiaApi.sk_font_new_with_values (typeface?.Handle ?? IntPtr.Zero, size, scaleX, skewX), true)
{
    if (Handle == IntPtr.Zero)
        throw new InvalidOperationException ("Unable to create a new SKFont instance.");
}
```

**SKFont.Typeface getter** — no special mapping needed, just return what C++ gives us:

```csharp
public SKTypeface Typeface {
    get => SKTypeface.GetObject (SkiaApi.sk_font_get_typeface (Handle));
    set => SkiaApi.sk_font_set_typeface (Handle, value == null ? IntPtr.Zero : value.Handle);
}
```

Users who previously checked `font.Typeface == null` can now check `font.Typeface.IsEmpty` instead.

### Behavior comparison

| Scenario | m119 (release) | m132 + this approach |
|---|---|---|
| `new SKFont().Typeface` | `null` | `SKTypeface.Empty` (check `.IsEmpty`) |
| `new SKFont(myTypeface).Typeface` | `myTypeface` | `myTypeface` ✅ |
| `font.Typeface = null; font.Typeface` | `null` | `SKTypeface.Empty` (check `.IsEmpty`) |
| `font.Typeface = myTypeface; font.Typeface` | `myTypeface` | `myTypeface` ✅ |

`SKPaint.Typeface` delegates to `SKFont.Typeface`, so it's automatically covered.

mattleibow:
## `SkCompatPaint` initializes with empty typeface — `new SKPaint()` can no longer draw text

Found a regression: `new SKPaint().MeasureText("Hello World!")` returns **0** on the PR. The existing test `SKPaintTest.MeasureTextMeasuresTheText` fails with `Assert.True(width > 0)`.

### Root cause

In `SkCompatPaint.cpp`, the default constructor changed between milestones:

**m119:**
```cpp
SkCompatPaint::SkCompatPaint()
    : fFont(SkFont())   // null typeface → lazy resolves to platform default via getTypefaceOrDefault()
```

**m132:**
```cpp
SkCompatPaint::SkCompatPaint()
    : fFont(SkFont(SkTypeface::MakeEmpty()))   // empty typeface — 0 glyphs, draws nothing
```

On m119, `SkFont()` stored a null `fTypeface`, but all drawing/measuring code went through `getTypefaceOrDefault()` which lazily resolved to the platform default. On m132, that lazy resolution was removed — `getTypefaceOrDefault()` no longer exists, and `SkFont` eagerly stores `MakeEmpty()` for null. So the empty typeface is the final answer — no fallback.

This means:
- `new SKPaint()` → font has empty typeface → all text measurement returns 0 → all text drawing produces nothing
- Lots of existing SkiaSharp code does `var paint = new SKPaint(); canvas.DrawText("hello", ...)` — this silently breaks

Reverting to `fFont(SkFont())` doesn't help — m132's `SkFont()` constructor also sets `MakeEmpty()`. The lazy default is gone at the C++ level.

### Possible fixes

**Option A: Use the font manager in `SkCompatPaint.cpp`** (submodule change)

```cpp
SkCompatPaint::SkCompatPaint()
    : fFont(SkFont(SkFontMgr::RefDefault()->legacyMakeTypeface(nullptr, SkFontStyle::Normal())))
    , ...
```

This gets the platform default typeface, matching m119 behavior. Downside: depends on font manager being available at construction time.

**Option B: Handle at the C# layer**

Set `SKTypeface.Default` on the paint's font after construction:

```csharp
public SKPaint()
    : this(SkiaApi.sk_compatpaint_new(), true)
{
    if (Handle == IntPtr.Zero)
        throw new InvalidOperationException("Unable to create a new SKPaint instance.");
    GetFont().Typeface = SKTypeface.Default;
}
```

Simpler, no submodule change needed, and uses the same `SKTypeface.Default` that this PR already establishes.

**Option C: Do both** — fix `SkCompatPaint` for C/C++ consumers, and set Default in C# for belt-and-suspenders.

Either way, `new SKPaint()` needs to have a real typeface with glyphs if we want text to work out of the box.

mattleibow:
## Test font and tests for `IsEmpty` validation

I created a hand-crafted TrueType font with valid tables and metadata but `numGlyphs=0` to validate that `IsEmpty` (`GlyphCount == 0`) correctly identifies both `SkEmptyTypeface` and real-but-empty fonts. All tests pass on the latest PR.

### ZeroGlyphs.ttf (base64, 724 bytes)

A minimal TrueType font with head, maxp, hhea, hmtx, cmap, name, post, glyf, loca tables — valid structure, family name "ZeroGlyphs", but `numGlyphs=0`.

[ZeroGlyphs.ttf.zip](https://github.com/user-attachments/files/27059266/ZeroGlyphs.ttf.zip)

Drop the font into tests/Content/Fonts/ZeroGlyphs.ttf

### Test cases

```csharp
[SkippableFact]
public void ZeroGlyphFontIsEmpty()
{
    using var tf = SKTypeface.FromFile(Path.Combine(PathToFonts, "ZeroGlyphs.ttf"));
    if (tf == null)
        return; // Skia rejected it — also valid

    Assert.Equal(0, tf.GlyphCount);
    Assert.True(tf.IsEmpty);
}

[SkippableFact]
public void ZeroGlyphFontHasFamilyName()
{
    using var tf = SKTypeface.FromFile(Path.Combine(PathToFonts, "ZeroGlyphs.ttf"));
    if (tf == null)
        return;

    // Unlike SkEmptyTypeface, this font has real metadata
    Assert.Equal("ZeroGlyphs", tf.FamilyName);
}

[SkippableFact]
public void ZeroGlyphFontMeasuresZero()
{
    using var tf = SKTypeface.FromFile(Path.Combine(PathToFonts, "ZeroGlyphs.ttf"));
    if (tf == null)
        return;

    using var font = new SKFont(tf);
    var width = font.MeasureText("Hello");
    Assert.Equal(0, width);
}

[SkippableFact]
public void EmptyTypefaceIsEmpty()
{
    Assert.True(SKTypeface.Empty.IsEmpty);
    Assert.Equal(0, SKTypeface.Empty.GlyphCount);
}

[SkippableFact]
public void DefaultTypefaceIsNotEmpty()
{
    Assert.False(SKTypeface.Default.IsEmpty);
    Assert.True(SKTypeface.Default.GlyphCount > 0);
}
```

### Results

All 5 pass on the latest PR commit. This validates that `IsEmpty => GlyphCount == 0` correctly identifies:
- `SKTypeface.Empty` (the `SkEmptyTypeface` singleton) ✅
- A real font file with valid tables but zero glyphs ✅
- `SKTypeface.Default` is not empty ✅

Per the OpenType spec, a valid font must have at least 1 glyph (`.notdef`), so `GlyphCount == 0` reliably means "this typeface cannot render anything."

mattleibow:
## Android API 36 test results — `SKTypeface.Default` is empty on API 36

Ran the full test suite on an **Android API 36 emulator** (Pixel_7, arm64-v8a) via xharness. Results:

| | Count |
|---|---|
| **Passed** | 3360 |
| **Failed** | 42 |
| **Skipped** | 187 |

### Root cause: `SKTypeface.Default` resolves to `Empty` on API 36

On Android API 36, `matchFamilyStyle(null, Normal)` returns null (the original #3693 bug). The PR's `CreateDefaultHandle()` falls back to `sk_typeface_create_empty()` — so `SKTypeface.Default` IS the empty typeface on this platform. This cascades:

- `SKTypeface.Default.IsEmpty == true` → no glyphs
- `SKTypeface.Default.FamilyName == ""` → empty name
- `new SKPaint()` initializes with `defaultFont` backed by `SKTypeface.Default` → also empty → `MeasureText` returns 0

### Failure breakdown

| Category | Count | Cause |
|---|---|---|
| **SKFontTest** `new SKFont()` empty typeface | 18 | m132 `MakeEmpty()` — same as macOS (#3735) |
| **SKPaintTest** text measurement | 16 | `SKPaint()` uses `Default` which is empty on API 36 |
| **SKPaintTest** new tests (DefaultPaintCanMeasureText, etc.) | 3 | Same — `Default` is empty |
| **SKPaintTest** ContainsTextIsCorrect | 1 | Same — `Default` has no glyphs |
| **SKTypefaceTest** DefaultFamilyNameIsNotNull | 1 | `Default.FamilyName` is empty |
| **SKTypefaceTest** DefaultIsNotEmpty | 1 | `Default.IsEmpty` is true |
| **SKFontTest** FontWithDefaultCanMeasure | 1 | `new SKFont(Default)` — Default is empty |
| **SKFontTest** ContainsTextIsCorrect | 1 | Default has no glyphs |

### The fix needs a better fallback on Android

The current fallback chain is:
```
matchFamilyStyle(null, Normal) → null → sk_typeface_create_empty()
```

But `create_empty()` has zero glyphs — it's not a usable default font. On Android, `legacyMakeTypeface(null, Normal)` returns a real default from `fDefaultStyleSet` when the family name is null. The fallback should use something that actually has glyphs, e.g.:

```cpp
// In CreateDefaultHandle or equivalent:
auto handle = matchFamilyStyle(null, Normal);
if (!handle) handle = legacyMakeTypeface(null, Normal);  // has fDefaultStyleSet fallback
if (!handle) handle = create_empty();  // last resort
```

Or at the C# level, try `MatchFamily(null)` then fall back to iterating font families to find any available font.

### macOS comparison

macOS console tests: **5363 passed, 18 failed** — the 18 are all the same `new SKFont()` empty typeface issue (#3735). No Android-specific failures on macOS.

PR #3796, opened 2026-04-29
Fix finalizer crash in SKPath when SKPathBuilder is collected first

ramezgerges:
**Description of Change**

Repro from a downstream app on 3.119: the finalizer thread crashed in sk_pathbuilder_detach_path with this stack:

    SkiaApi.sk_pathbuilder_detach_path(IntPtr)
    SKPath.FlushBuilder()
    SKPath.get_Handle()
    SKNativeObject.Dispose(Boolean)
    SKObject.Dispose(Boolean)
    SKPath.Dispose(Boolean)
    SKNativeObject.Finalize()

SKPath holds a private SKPathBuilder for batched mutations and overrides Handle to flush pending mutations on every read. SKPathBuilder is itself an SKObject with its own finalizer and the CLR makes no guarantee about finalization order — when the builder finalizes first, its native pathbuilder is freed and its Handle is set to IntPtr.Zero. The path's finalizer then runs SKNativeObject.Dispose(false), which reads Handle on line 269 to check `!= IntPtr.Zero`. That dispatches into our overridden getter, FlushBuilder calls
sk_pathbuilder_detach_path(_builder.Handle), and the C side dereferences nullptr (or a dangling pointer if the same address was reallocated).

Skip FlushBuilder once disposal has begun. The pending builder mutations are about to be discarded along with the path either way, and DisposeNative still cleans up _builder defensively (the CAS in SKNativeObject.Dispose makes a second Dispose a no-op if the builder already finalized).

Added a regression test that deterministically reproduces the crash without the fix (Test host process crashed) and passes with it.

**Bugs Fixed**

- Fixes #

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/binding/SkiaSharp/SKPath.cs
+++ b/binding/SkiaSharp/SKPath.cs
@@ -25,9 +25,17 @@ namespace SkiaSharp
 		// reads path.Handle directly and P/Invokes with it. If mutations have been batched
 		// into _builder but not yet flushed, base.Handle points at a stale native SkPath.
 		// Flushing in the getter keeps every reader — internal or external — honest.
+		//
+		// Skip the flush once disposal has begun. SKPathBuilder is itself an SKObject with
+		// its own finalizer, so on the finalizer thread it may already have been collected
+		// (its Handle is IntPtr.Zero, native pointer freed). Touching it here would call
+		// sk_pathbuilder_detach_path on a null/dangling handle. The pending mutations are
+		// going to be discarded with the path anyway, and DisposeNative cleans up _builder
+		// defensively.
 		public override IntPtr Handle {
 			get {
-				FlushBuilder ();
+				if (!IsDisposed)
+					FlushBuilder ();
 				return base.Handle;
 			}
 			protected set => base.Handle = value;
--- a/tests/Tests/SkiaSharp/SKPathTest.cs
+++ b/tests/Tests/SkiaSharp/SKPathTest.cs
@@ -725,5 +725,31 @@ namespace SkiaSharp.Tests
 		}
 
 #pragma warning restore CS0618
+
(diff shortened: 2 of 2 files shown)


mtmattei:
here is verbatim crash:

  Environment:
  - Windows 11, .NET SDK 10.0.103 (band 100, latestFeature roll-forward)
  - Uno.Sdk 6.6.0-dev.208
  - SkiaSharp 4.147.0-preview.1.1 (managed + Views.Uno.WinUI + HarfBuzz)
  - TFM: net10.0-desktop
  - Uno SkiaRenderer feature enabled, UnoDisableLottieSkiaVersionCheck=true
  - App: minimal Uno Single Project shell with NavigationView + 3 pages,
    each rendering via Uno.WinUI.Graphics2DSK.SKCanvasElement subclasses

  Crash (identical in Debug and Release, ~20-30s after window opens, no
  user interaction required):

  Fatal error.
  0xC0000005
     at SkiaSharp.SkiaApi.sk_pathbuilder_detach_path(IntPtr)
     at SkiaSharp.SKPath.get_Handle()
     at SkiaSharp.SKObject.Dispose(Boolean)
     at SkiaSharp.SKPath.Dispose(Boolean)
     at SkiaSharp.SKNativeObject.Finalize()
     at System.GC.RunFinalizers()

  The app code never directly constructs SKPath — these are paths created
  inside Uno's own rendering pipeline. Suggests an SKPath/SKPathBuilder
  lifecycle bug at the Uno-preview/SkiaSharp-4.147-preview boundary
  (double-free, or native handle freed before managed finalizer runs).

Issue #3732, opened 2026-04-24
[obsolete] Promote SKPaint text/font obsolete members to error and plan SkCompatPaint removal


mattleibow:
# Mark SKPaint text/font obsolete members as `[Obsolete(error: true)]` and reduce SkCompatPaint

## Summary

`SKPaint` has **~115 obsolete members** that delegate to `SKFont` for text-related operations. These are currently `[Obsolete("Use SKFont instead.")]` (warning only). We should promote these to `[Obsolete("...", error: true)]` to force migration, and then evaluate removing the underlying `SkCompatPaint` C++ shim.

## Background

`SKPaint` historically combined paint properties (color, stroke, blend) with text properties (typeface, size, measure text). Upstream Skia separated text into `SkFont` years ago. SkiaSharp added `SkCompatPaint` (a custom C++ class in `src/xamarin/`) to bridge the gap — it embeds an `SkFont` inside an `SkPaint`.

All text-related `SKPaint` members are already marked `[Obsolete]` with migration guidance to `SKFont`. It's time to make them errors.

### Font/text properties (13 members)
- `Typeface` → `SKFont.Typeface`
- `TextSize` → `SKFont.Size`
- `TextScaleX` → `SKFont.ScaleX`
- `TextSkewX` → `SKFont.SkewX`
- `IsLinearText` → `SKFont.LinearMetrics`
- `SubpixelText` → `SKFont.Subpixel`
- `LcdRenderText` → `SKFont.Edging`
- `IsEmbeddedBitmapText` → `SKFont.EmbeddedBitmaps`
- `IsAutohinted` → `SKFont.ForceAutoHinting`
- `HintingLevel` → `SKFont.Hinting`
- `FakeBoldText` → `SKFont.Embolden`
- `FontSpacing` → `SKFont.Spacing`
- `FontMetrics` / `GetFontMetrics()` → `SKFont.Metrics` / `SKFont.GetFontMetrics()`

### Text measurement (~24 overloads)
- `MeasureText(...)` → `SKFont.MeasureText(...)`
- `BreakText(...)` → `SKFont.BreakText(...)`

### Glyph queries (~18 overloads)
- `CountGlyphs(...)`, `GetGlyphs(...)`, `ContainsGlyphs(...)` → `SKFont` equivalents
- `GetGlyphWidths(...)`, `GetGlyphPositions(...)`, `GetGlyphOffsets(...)` → `SKFont` equivalents

### Text path (~18 overloads)
[...]

PR #4068, opened 2026-05-25
Promote SKPaint text/font obsolete members to error

ramezgerges:
**Description of Change**

Phase 1 of #3732. Promotes the warning-only [Obsolete] attributes on SKPaint's ~115 text/font members, the 5 paint-only DrawText/DrawTextOnPath overloads on SKCanvas, and the 11 SKFilterQuality-consuming overloads across SKBitmap/SKImage/SKShader/SKPixmap to [Obsolete(..., error: true)], so they are stripped from the reference assembly by the existing RemoveObsoleteSymbols build task.

Internal callers in SKCanvas (DrawImage/DrawAtlas/DrawText) and SkiaSharp.HarfBuzz continue to respect legacy paint.FilterQuality and paint.TextAlign state through four new non-obsolete bypass helpers on SKPaint (GetLegacyFont, GetLegacyTextAlign, GetLegacyTextEncoding, GetLegacyFilterQualitySampling), exposed via the existing InternalsVisibleTo declarations. No behavior change visible to consumers of either binding.

Also incorporates the two warning-level DrawText obsoletions from PR #3165, with affected Gallery samples migrated to pass SKTextAlign.Left explicitly (none of them set paint.TextAlign).

Obsolete SKPaint text-API tests removed from SKPaintTest.cs; equivalent SKFont-API coverage already exists in SKFontTest.cs. SkCompatPaint stays in the native shim per the issue's Phase 1 plan.

**Bugs Fixed**

- Fixes #

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation


binding/SkiaSharp/SKPaint.cs
```diff
@@ -37,7 +37,6 @@ public static SKSamplingOptions ToSamplingOptions (this SKFilterQuality quality)
 
 	public unsafe class SKPaint : SKObject, ISKSkipObjectRegistration
 	{
-		[Obsolete]
```
mattleibow, line 40: Is this intentional?

ramezgerges: Not really but nonetheless correct? `[Obsolete]` on the private field will only raise warnings inside the file, is this wanted?

binding/SkiaSharp/SKPixmap.cs
```diff
@@ -156,10 +156,10 @@ public unsafe Span<T> GetPixelSpan<T> (int x, int y)
 				if (bpp != size)
 					throw new ArgumentException ($"Size of T ({size}) is not the same as the size of each pixel ({bpp}).", nameof (T));
 
-				spanLength = info.Width * info.Height;
+				spanLength = checked(info.Width * info.Height);
 
 				if (x != 0 || y != 0)
-					spanOffset = y * info.Height + x;
+					spanOffset = checked(y * info.Height + x);
```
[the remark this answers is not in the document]
mattleibow: Good catch — this is a real bug but it's **pre-existing on `main`** (not introduced by this PR). This PR only wrapped the existing expression in `checked()`. Filed as #4111 to fix separately.

binding/SkiaSharp/SKPaint.cs
```diff
@@ -859,14 +858,39 @@ public float[] GetHorizontalTextIntercepts (IntPtr text, IntPtr length, float[]
...
+		internal SKFont GetLegacyFont () =>
+			font ??= OwnedBy (SKFont.GetObject (SkiaApi.sk_compatpaint_get_font (Handle), false), this);
+
+		internal SKSamplingOptions GetLegacyFilterQualitySampling () =>
+			SkiaApi.sk_compatpaint_get_filter_quality (Handle) switch {
+				1 => new SKSamplingOptions (SKFilterMode.Linear, SKMipmapMode.None),
+				2 => new SKSamplingOptions (SKFilterMode.Linear, SKMipmapMode.Linear),
+				3 => new SKSamplingOptions (SKCubicResampler.Mitchell),
+				_ => new SKSamplingOptions (SKFilterMode.Nearest, SKMipmapMode.None),
+			};
```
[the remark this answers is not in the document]
mattleibow: Already addressed — the third commit (`2ba4f418 Throw on unknown FilterQuality`) added a `throw new ArgumentOutOfRangeException` for the `_` default case. See line 894 in the current version of `SKPaint.cs`.

binding/SkiaSharp/SKPaint.cs
mattleibow: Removing `[Obsolete]` from the private field is fine here. The attribute was only suppressing CS0618 warnings within the class itself when the field was accessed from other `[Obsolete]` members. Now that internal access goes through `GetLegacyFont()` (which is not obsolete), the attribute serves no purpose. Keeping it would be harmless but misleading — it suggests something deprecated when it's just an implementation detail.

PR #4112, opened 2026-06-03
Bugfix #2967 - Change counter into boolean

### LINKED ISSUE #2967: [BUG] Multiple Unloaded events in WinUI causes empty SKXamlCanvas
### Description

I'm displaying different SKXamlCanvas controls in separate WinUI TabView tabs. When I remove multiple tabs, the currently selected tab drops its rendered content. I found that this is really a WinUI issue, https://github.com/microsoft/microsoft-ui-xaml/issues/9880, where multiple Unloaded events are triggered on the wrong FrameworkElement. In the SKXamlCanvas code I can see that the backgound is cleared on Unloaded events, so that's why my content disappears. I also found my way to https://github.com/mono/SkiaSharp/issues/1118 where you seem to have fixed a somewhat related issue.

It seems like WinUI has an erratic Load/Unload behavior in other circumstances as well, and that you maybe cannot rely on them coming in correct order.
https://github.com/microsoft/microsoft-ui-xaml/issues/1900
https://github.com/microsoft/microsoft-ui-xaml/issues/8342

When debugging this, I noticed that when the SKXamlCanvas recieves incorrect Unloaded events it actually has IsLoaded == true. Maybe the current fix with the loadUnloadCounter could be replaced with a private isLoaded flag which can be set in OnLoaded, and then prevent multiple Loaded events to be processed? The flag should then only get reset in OnUnloaded if IsLoaded actually is false.

### Code

See https://github.com/microsoft/microsoft-ui-xaml/issues/9880

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Version of SkiaSharp

2.88.3 (Current)

### IDE / Editor

Visual Studio (Windows)

Windows

_No response_

### Devices

_No response_

### Relevant Screenshots

_No response_

### Relevant Log Output

_No response_

### Code of Conduct

Mangepange:
**Description of Change**

Changed the integer counter for keeping track of Load and Unload events in SKXamlCanvas into a boolean.
This prevents issues when the Unload event comes before Load, and also when multiple Unload events are triggered (both of these situations arise due to WinUI issues).

(No added tests since I couldn't find any previous regarding the SKXamlCanvas)

**Bugs Fixed**

- Fixes #2967 

**API Changes**

None.

**Behavioral Changes**

None.

**Required skia PR**

None.

**PR Checklist**

- [ ] Merged related skia PRs
- [ ] Updated documentation

### DIFF
--- a/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs
+++ b/source/SkiaSharp.Views/SkiaSharp.Views.WinUI/SKXamlCanvas.cs
@@ -43,7 +43,7 @@ namespace SkiaSharp.Views.UWP
 		private bool isVisible = true;
 
 		// workaround for https://github.com/mono/SkiaSharp/issues/1118
-		private int loadUnloadCounter = 0;
+		private bool isLoaded;
 
 		public SKXamlCanvas()
 		{
@@ -124,9 +124,10 @@ namespace SkiaSharp.Views.UWP
 
 		private void OnLoaded(object sender, RoutedEventArgs e)
 		{
-			loadUnloadCounter++;
-			if (loadUnloadCounter != 1)
+			if (isLoaded)
 				return;
+
+			isLoaded = true;
 
 #if WINDOWS
 			XamlRoot.Changed += OnXamlRootChanged;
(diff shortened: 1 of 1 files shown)


Mangepange:
@dotnet-policy-service agree

Issue #3817, opened 2026-04-30
[BUG] SKFontManager.Default throws an exception


flabbet:
### Description

Calling `SKFontManager.Default` in the latest `4.147.0-preview.1.1` throws an exception

```
Unhandled exception. System.TypeInitializationException: The type initializer for 'SkiaSharp.SKFontManager' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'SkiaSharp.SKObject' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'SkiaSharp.SKTypeface' threw an exception.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at SkiaSharp.SKTypeface..cctor()
   --- End of inner exception stack trace ---
   at SkiaSharp.SKTypeface.EnsureStaticInstanceAreInitialized()
   at SkiaSharp.SKObject..cctor()
   --- End of inner exception stack trace ---
   at SkiaSharp.SKObject..ctor(IntPtr handle, Boolean owns)
   at SkiaSharp.SKFontManager..ctor(IntPtr handle, Boolean owns)
   at SkiaSharp.SKFontManager.SKFontManagerStatic..ctor(IntPtr x)
   at SkiaSharp.SKFontManager..cctor()
   --- End of inner exception stack trace ---
   at SkiaSharp.SKFontManager.get_Default()

```

Tested with both Win32 native assets package and macos native assets package. Happens on .NET 8 and .NET 10

### Code

```cs
using SkiaSharp;
var def = SKFontManager.Default;
```

### Version of SkiaSharp

Other (Please indicate in the description)

### Last Known Good Version of SkiaSharp

3.116.0 (Current)

### IDE / Editor

Other (Please indicate in the description)

### Platform / Operating System

Windows

### Platform / Operating System Version

- macOS
- Windows 11

### Relevant Log Output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

flabbet:
Found a workaround for preview 4.147.0-preview.3.1.

before calling `SKFontManager.Default`, call `SKTypeface.CreateDefault();`

PR #4116, opened 2026-06-04
Alternate fix for #3817: re-entrant static-init NRE in SKTypeface (+ #4080 test suite)

mattleibow:
## What this is

An **alternate attempt** to PR #4080 for the re-entrant static-init `NullReferenceException` (#3817). Instead of #4080's broader rework, this PR takes **PR #4107's narrower `SKTypeface` fix** and brings in **#4080's full test suite**, reconciling the two so every test passes.

Both PRs target the same bug; this is the smaller production change of the two, validated against the larger test set.

## Production changes

- **`SKTypeface.cs`** — PR #4107's fix. The static cctor builds the empty/default typefaces from raw native handles in a `try/finally`, severing the managed-singleton reads (`SKFontManager.Default` / `SKFontStyle.Normal`) that triggered the re-entrant cold-start crash.
- **Static singleton wrappers** (`SKColorSpaceStatic`, `SKDataStatic`, `SKBlenderStatic`, `SKFontStyleStatic`, `SKFontManagerStatic`, `SKTypefaceStatic`, `SKColorFilterStatic`, `SKFontStatic`) — each constructor now calls `PreventPublicDisposal()` so the singleton carries `IgnorePublicDispose = true`. These wrappers already protected disposal *structurally* (no-op `Dispose(bool)` + `owns=false`); setting the flag unifies that with the `IgnorePublicDispose` contract #4080's tests assert. Purely additive and ABI-safe.

## Tests (from #4080)

- `SkiaSharp.Tests.SingletonInit.Console` — dedicated #3817 cold-start regression guard (single-process assembly).
- `GarbageCleanupFixture.cs` — restored to main's static-type name-list (the #4080 `IgnorePublicDispose`-keyed rewrite is not needed here).
- Full set of #4080 lifecycle / concurrency / reentrancy / singleton-init test files.

## Result

macOS arm64, net10.0, full single-process assembly run:

**`Failed: 0, Passed: 5706, Skipped: 13`** — fully green, including the #3817 cold-start guard and the previously-flaky `SKDocumentTest.StreamIsNotCollectedPrematurely`.

## Conclusion

#4107's minimal `SKTypeface` fix resolves the #3817 re-entrancy crash, and the `PreventPublicDisposal()` reconciliation lets it satisfy #4080's complete dispose-protection / concurrency / lifecycle test suite. This is the lower-risk production change relative to #4080.

Refs #3817. Alternative to #4080; brings in #4107's fix.


mattleibow:

PR #4117, opened 2026-06-04
Latch process-global singletons immortal (heap-fix iteration for #3817 / #4080)

### LINKED ISSUE #3817: [BUG] SKFontManager.Default throws an exception

mattleibow:
## Why

This is the **green-it iteration** branch for the singleton lifecycle rework (#4080, fixes #3817). It contains all of #4080's rework + the ported concurrency/lifecycle tests **rebased on current `main`**, plus one additional production fix on top. The goal is to get **net48 (`tests-netfx`) CI reliably green** here without churning #4080, then merge the final fix.

## The bug this fix closes

#4080 fixed the #3817 static-constructor cycle by making process-global singletons (sRGB color spaces, `SKData.Empty`, the default font manager/typeface, blend-mode blenders, the gamma color filters, the `SKFontStyle` presets) normal `owns:true` wrappers and moving the `isDisposed` CAS out of `Dispose(bool)` into **three** entry points: public `Dispose()`, `DisposeInternal()`, and the finalizer.

`IgnorePublicDispose` only guards the **public** `Dispose()` path. So `DisposeInternal()` (owned-child teardown, ownership handoff, stale-wrapper replacement) and the finalizer can still CAS `isDisposed` and run `Dispose(true)` → `DisposeNative()` → `SafeUnRef()`, **freeing a process-global native singleton** out from under every other consumer. On net48 this surfaces as the intermittent native `AccessViolationException` blocking #4080. The pre-rework code was truly immortal because its `Dispose(bool)` was an unconditional no-op.

## The fix

A one-way **immortal latch** (`isImmortalSingleton`, `Volatile`) that short-circuits all three teardown entry points **before** the `isDisposed` CAS:

- Set inside `HandleDictionary.GetOrAddObject` **under the registry lock** (alongside `PreventPublicDisposal`) in both the found-existing and freshly-constructed branches.
- `RegisterHandle` refuses to replace/dispose an already-immortal handle.
- Normal `owns:true` semantics are unchanged for every non-singleton wrapper.

This restores the pre-rework "truly immortal" guarantee for singletons while keeping #4080's `LazyInitializer` cctor-cycle fix.

## Tests

- Matched **immortal vs mortal** `DisposeInternal` control pair (deterministic, native-free counting fake): immortal → native-free path never runs; mortal → runs exactly once.
- `AllSingletonsAreImmortal` + per-unref-path (`SKColorSpace` virtual, `SKData` non-virtual, default typeface) `DisposeInternal` no-op proofs over the **real** singletons.
- The immortal fake is rooted in a static list, mirroring how every real singleton is held alive by a static field, so it leaves no dead `WeakReference` for `GarbageCleanupFixture`.

Local: Debug full suite green (5716 passed / 12 skipped / 0 failed). net48 can only be verified in CI — that's what this PR is for.


mattleibow:

PR #4124, opened 2026-06-08
Experiment: #4119 minus the immortal latch (isolate whether immortal is load-bearing)

mattleibow:
## Purpose — a controlled single-variable experiment

We've spent too long comparing whole PRs that each move **multiple levers at once** (eager-vs-lazy init, immortal latch, callback isolation, dispose-protection), so no comparison isolates anything. This PR exists to settle **one** question with CI as the oracle:

> **Is the "immortal singleton" latch actually load-bearing for the intermittent net48/x86 `AccessViolationException`, or is it redundant?**

### Method
Branched off **#4119 HEAD (`910491d`)** — eager init + dispose-protection + immortal latch, with callback isolation already reverted. The **only** delta in this PR is **removing the immortal machinery**. Everything else is the same base.

That base is **green 3x on AzDO net48/x86** (builds 158085/86/87). So:

| Result on net48/x86 | Conclusion |
|---|---|
| **Red** (AVE returns) | The immortal latch **is** load-bearing — keep it. |
| **Green** across multiple runs | The latch is **redundant** on top of eager + dispose-protection — drop it. |

### What's removed
- `SKObject`: `isImmortalSingleton`, `IsImmortalSingleton`, `MakeImmortalSingleton()`, `GetOrAddImmortalSingletonObject`, and the three teardown guards (finalizer / `Dispose` / `DisposeInternal`).
- `HandleDictionary`: `immortal` param/overload, invariant check, promote/new-immortal branches, `RegisterHandle` guard.
- Per-type accessors repointed to `GetOrAddDisposeProtectedObject`; immortal-only tests removed; doc rewritten to dispose-protected-only with the redundancy rationale.

### What's retained
**Dispose-protection** (getter-route dedup) — undisputed and present in **both** #4080 and #4119.

### Note
This is a **diagnostic**, not a merge candidate as-is. Please read the net48/x86 legs rather than squash-merging. Local: build clean (0 errors), full managed suite green on macOS arm64.


mattleibow:

PR #2630, opened 2023-09-22
Bind Skottie's Animation Builder

mattleibow:
**Description of Change**

This PR adds a few of the essential APIs to enable loading embedded files in the Lottie animation. Currently, there is just the support for file-based and data URI-based assets.

**Bugs Fixed**

- Fixes #2173

**API Changes**

```cs
class Animation {
  public static AnimationBuilder CreateBuilder (AnimationBuilderFlags flags = AnimationBuilderFlags.None);
}

public sealed unsafe class AnimationBuilder : SKObject, ISKSkipObjectRegistration {
  public AnimationBuilder SetFontManager (SKFontManager fontManager);
  public AnimationBuilder SetResourceProvider (ResourceProvider resourceProvider);
  public AnimationBuilderStats Stats { get; }
  public Animation? Build (Stream stream);
  public Animation? Build (SKStream stream);
  public Animation? Build (SKData data);
  public Animation? Build (string path);
}

public struct AnimationBuilderStats {
  public readonly TimeSpan TotalLoadTime { get; }
  public readonly TimeSpan JsonParseTime { get; }
  public readonly TimeSpan SceneParseTime { get; }
  public readonly int JsonSize  { get; }
  public readonly int AnimatorCount  { get; }
}

public abstract unsafe class ResourceProvider : SKObject, ISKReferenceCounted, ISKSkipObjectRegistration {
  public SKData? Load (string resourceName);
  public SKData? Load (string resourcePath, string resourceName);
}

public class CachingResourceProvider : ResourceProvider {
  public CachingResourceProvider (ResourceProvider resourceProvider);
}

public class DataUriResourceProvider : ResourceProvider {
  public DataUriResourceProvider (bool preDecode = false);
  public DataUriResourceProvider (ResourceProvider? fallbackProvider, bool preDecode = false);
}

public class FileResourceProvider : ResourceProvider {
  public FileResourceProvider (string baseDirectory, bool preDecode = false);
}
```

**Behavioral Changes**

None.

**Required skia PR**

https://github.com/mono/skia/pull/114

**PR Checklist**

- [x] Has tests (if omitted, state reason in description)
- [x] Rebased on top of main at time of PR
- [ ] Merged related skia PRs
- [x] Changes adhere to coding standard
- [ ] Updated documentation

PR #4126, opened 2026-06-08
Reproduce and fix the net48/x86 Skottie render AccessViolationException

mattleibow:
## Goal

Identify, reproduce, and **properly** fix the intermittent `AccessViolationException` that fails the **net48/x86** CI leg for the #3817 singleton work (seen on #4080). No serialization, no "make-it-green" hacks.

## What the crash actually is (evidence)

In build 158052 the net48/x86 shard aborted **3 separate times**, and every time the unhandled `AccessViolationException` stack pinned to:

```
at SkiaSharp.SkottieApi.skottie_animation_render(IntPtr, IntPtr, SKRect*) SkottieApi.generated.cs:536
at SkiaSharp.Tests.AnimationBuilderTest.CanRenderWithBase64(String) AnimationBuilderTest.cs
```

Other test classes (`EnumMappingsTest`, `SKRectTest`, `SKPaintTest`) were merely running around it as parallel-collection bystanders and survived. A consistent crash **site** (3/3) points at the code at that site, not random cross-thread heap corruption (which would scatter).

### Established by investigation

- `CanRenderWithBase64` and the `lottie-base64_*.json` files are **pre-existing** (added in #2630, 2023) and live on `main`. This is **not** introduced by the singleton rework.
- The crash requires **render + lazy embedded-image decode** (`DataUriResourceProvider` default `preDecode: false`). The load-only base64 tests using the same provider never crash.
- The lotties contain **no text layers and no fonts**, so this is **not** font/singleton related.
- It is **not** premature provider GC (a 200× build→GC→render loop is clean locally).
- The P/Invoke signature is correct, so it is **not** a managed marshalling bug.

## Reproduction harness (this PR)

`tests/Tests/Skottie/SkottieAveReproTest.cs` adds three experiments. All three are **clean on macOS arm64/net10**, consistent with an x86-only manifestation:

- **Exp1** — single-threaded build/GC/render loop (isolates a single-path bug)
- **Exp2** — parallel skottie render + GC pressure
- **Exp3** — skottie render concurrent with heterogeneous Skia work (codec decode, text paint)

The net48/x86 leg of this PR is the controlled x86 environment used to determine the bug class:
- If **Exp1** crashes on x86 → genuine single-path decode/lifetime bug → fix the decode path.
- If only **Exp2/Exp3** crash → cross-thread corruption → hunt the specific corruptor.

The proper fix will be committed to this PR once the x86 signal lands.

### DIFF
--- a/binding/SkiaSharp/HandleDictionary.cs
+++ b/binding/SkiaSharp/HandleDictionary.cs
@@ -57,15 +57,18 @@ namespace SkiaSharp
 		/// <returns>The instance, or null if the handle was null.</returns>
 		internal static TSkiaObject GetOrAddObject<TSkiaObject> (IntPtr handle, bool owns, bool unrefExisting, Func<IntPtr, bool, TSkiaObject> objectFactory)
 			where TSkiaObject : SKObject =>
-			GetOrAddObject (handle, owns, unrefExisting, immortal: false, objectFactory);
+			GetOrAddObject (handle, owns, unrefExisting, disposeProtected: false, objectFactory);
 
 		/// <summary>
-		/// Retrieve or create an instance for the native handle. When <paramref name="immortal"/> is true
-		/// and an existing wrapper is found, IgnorePublicDispose is set on it inside the critical section
-		/// — narrowing the promote-existing race window relative to setting it after the call returns.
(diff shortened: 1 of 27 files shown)


mattleibow:

PR #4271, opened 2026-06-27
Bind SKPaint.GetFastBounds

mattleibow:
## Summary

Binds the public **`SkPaint::computeFastBounds`** / **`SkPaint::canComputeFastBounds`** pair as a single Try-style method:

- `bool SKPaint.GetFastBounds(SKRect bounds, out SKRect fastBounds)`

`GetFastBounds` returns the conservative device-space rectangle a paint can draw into for a given source rect, and **composes every effect on the paint** — path effect, stroke inflation, mask filter and image filter — so the result accounts for blurs, thick/mitered strokes, dashes, drop shadows, etc. Paired with the already-bound `SKCanvas.QuickReject`, this completes the quick-reject culling pattern that Skia documents in its own headers:

```cpp
if (!path.isInverseFillType() && paint.canComputeFastBounds()) {
    SkRect storage;
    if (canvas->quickReject(paint.computeFastBounds(path.getBounds(), &storage)))
        return; // do not draw
}
```

In C# that whole guard becomes one call:

```csharp
if (paint.GetFastBounds(path.Bounds, out var fast) && canvas.QuickReject(fast))
    return; // do not draw
```

Companion C API PR: **mono/skia#271**.

## API shape: one method, not two

Skia's own contract is that `canComputeFastBounds()` is a **guard that must be checked before** calling `computeFastBounds()` — the latter is unreliable for paints whose image filter or path effect cannot report bounds (the header literally says *"Only call this if `canComputeFastBounds()` returned true"*). That maps exactly onto the .NET Try-pattern, so the two are collapsed into a single `GetFastBounds(..., out ...) → bool`:

- The guard is folded into the return value, so the Skia contract **cannot be misused** (a `false` short-circuits before `QuickReject`).
- It matches the existing SkiaSharp out-bool bounds family — `SKPath.GetBounds(out)`, `SKPath.GetTightBounds(out)`, `SKCanvas.GetLocalClipBounds(out)`, `SKCanvas.GetDeviceClipBounds(out)` — none of which expose a separate `Can…` property (there is no `CanGetTightBounds`).
- `Compute` in SkiaSharp is reserved for value-returning mirrors of Skia `computeXxx()` (`SKPath.ComputeTightBounds`, `SKPixmap.ComputeIsOpaque`); since this member returns a `bool`, `GetFastBounds` is the consistent name.

`GetFastBounds` returns `false` (and `SKRect.Empty`) when the paint cannot compute fast bounds. The two underlying C exports remain 1:1 with Skia for any future direct use.

## Why the public `SKPaint` API and not `SKMaskFilter`?

This started as a binding of `SkMaskFilter::computeFastBounds`, but that method is **private**. The investigation below explains what happened and why the public `SkPaint` entry point is the correct thing to bind.

### History (traced in the Skia git log)

| When | Commit | What |
|------|--------|------|
| 2012-01-30 | `9efd9a048a` | `computeFastBounds` **added** to the public `SkMaskFilter` — explicitly "to perform quick-rejects when drawing objects with shadows (esp. text)." |
| 2018-01-23 | `80747ef591` (Mike Reed) | "move the guts of `SkMaskFilter.h` into `SkMaskFilterBase.h`" — **made it private**. Empty bug field; pure encapsulation. Part of Skia's sweeping move to make every effect type an opaque public handle (factory + serialization only), pushing all implementation methods (`filterMask`, `computeFastBounds`, `getFormat`, …) into internal `*Base` classes — same as `SkShaderBase`, `SkColorFilterBase`, `SkPathEffectBase`. |

So `SkMaskFilter::computeFastBounds` has been a deliberate **internal implementation detail for ~8 years**. Binding it would force our C shim to `#include "src/core/SkMaskFilterBase.h"` and call `as_MFB()` — exactly the kind of private coupling that broke in 2018 and can break again on any Skia bump.

### What about the `(to be made private)` note on `SkPaint::computeFastBounds`?

`SkPaint::computeFastBounds` carries a `(to be made private)` doc annotation, added **2018-01-03** (`2823f9f06c1`, Cary Clark) as a bulk note across several `SkPaint` methods. **Eight years later it is still public, with no churn toward privatizing it** (no commits in that direction since 2023). It remains the documented, canonical quick-reject API, and SkiaSharp already wraps a number of APIs with similar long-standing soft-deprecation notes.

Notably, `SkImageFilter::computeFastBounds` stayed **public** the whole time — Skia kept the image-filter bounds public but routed mask-filter bounds through `SkPaint`, reinforcing that `SkPaint` is the intended entry point.

### Community signal (groups.google.com/g/skia-discuss)

Real threads on the Skia discussion group show this is a recurring user need, and that `computeFastBounds` is the recommended answer:

- **"How to get the path bound including border?"** ([EFbFlfJjNxE](https://groups.google.com/g/skia-discuss/c/EFbFlfJjNxE), 2024-01-28) — a user asks for an API to get the visual bound including shadow/blur/non-zero border, noting it is unreliable to compute manually because it depends on Skia's implementation. The reply recommends **`computeFastBounds()`**.
- **"Custom ImageFilter and affected area bounds"** ([HHT3w0Kmm6Q](https://groups.google.com/g/skia-discuss/c/HHT3w0Kmm6Q)) — a user whose content is being culled references `SkBlurImageFilter` overriding `computeFastBounds` and wants to expand the affected bounds.

### Conclusion

Bind the **public** `SkPaint::computeFastBounds` / `canComputeFastBounds`. It is genuinely public, idiomatic, stable, composes the mask-filter bounds internally (so it subsumes the original blur use case), and — as a single `GetFastBounds(out) → bool` — completes the quick-reject culling trio with `SKCanvas.QuickReject`. The `// TODO: computeFastBounds` on `SKMaskFilter` is intentionally left in place since that method remains private upstream.

## Changes

- **C API** (mono/skia#271): `sk_paint_compute_fast_bounds`, `sk_paint_can_compute_fast_bounds`.
- **Bindings**: regenerated `SkiaApi.generated.cs`.
- **Wrapper**: `SKPaint.GetFastBounds(SKRect, out SKRect) → bool`.
- **Tests** (`SKPaintTest`): blur outsets by 3×sigma; stroke (round join/cap) outsets by strokeWidth/2; stroke+blur composition; fill with no effects returns source unchanged; source rect not mutated; returns `true` for simple paints; managed/native parity via direct `SkiaApi` call.
- **Sample**: `PaintFastBoundsSample` (Gallery) — visualizes the composed fast bounds with sigma/stroke/blur-style/toggle controls.
- **Benchmark**: `PaintFastBoundsBenchmark` — quick-reject culling vs. unconditional draw, plus isolated call cost.

## Testing

- Native rebuilt from source (macOS arm64); both `sk_paint_*_fast_bounds` symbols exported.
- `SKPaintTest` (fast-bounds + paint): **13 passed, 0 failed** (1 unrelated pre-existing skip).
- Full managed suite previously: **5743 passed, 0 failed, 13 pre-existing skips**.
- Benchmark and Gallery shared projects build clean.

### DIFF
--- a/benchmarks/SkiaSharp.Benchmarks/Benchmarks/PaintFastBoundsBenchmark.cs
+++ b/benchmarks/SkiaSharp.Benchmarks/Benchmarks/PaintFastBoundsBenchmark.cs
@@ -1,0 +1,111 @@
+using System;
+using BenchmarkDotNet.Attributes;
+
+namespace SkiaSharp.Benchmarks;
+
+// SKPaint.GetFastBounds exists to power quick-reject culling: before issuing an expensive
+// draw, a renderer can ask the paint for the conservative device-space rect the draw would touch
+// - composing the path effect, stroke, mask filter and image filter - and skip the draw entirely
+// when that rect misses the clip/viewport.
+//
+// This benchmark models a frame that submits many blurred rects scattered across a virtual space
+// four times larger than the viewport, so roughly three quarters of the items land off-screen and
+// are cullable. It compares drawing every item unconditionally against using GetFastBounds to
+// cull off-screen items first, plus an isolated measurement of the raw GetFastBounds call cost
+// (the P/Invoke crossing plus the native geometry math).
+//
+// Runs on the project's current TFM with the default job, matching SurfaceCanvasBenchmark.
+[MemoryDiagnoser]
+public class PaintFastBoundsBenchmark
+{
+	private const int Width = 256;
+	private const int Height = 256;
+
+	// Frames rendered per invocation. 60 ~= one second at 60 FPS.
+	[Params(1, 60)]
+	public int Frames { get; set; }
+
+	// Draw items submitted per frame.
+	[Params(500)]
+	public int ItemsPerFrame { get; set; }
+
+	private SKSurface surface;
+	private SKCanvas canvas;
+	private SKPaint paint;
+	private SKMaskFilter blur;
+	private SKRect clip;
+	private SKRect[] items;
+
+	[GlobalSetup]
+	public void GlobalSetup()
+	{
+		surface = SKSurface.Create(new SKImageInfo(Width, Height, SKColorType.Rgba8888, SKAlphaType.Premul));
+		canvas = surface.Canvas;
+		blur = SKMaskFilter.CreateBlur(SKBlurStyle.Normal, 6f);
+		paint = new SKPaint { IsAntialias = true, Color = SKColors.SteelBlue, MaskFilter = blur };
+
+		// The on-screen viewport that draws must intersect to be visible.
+		clip = SKRect.Create(0, 0, Width, Height);
+
+		// Scatter items across a 4x larger virtual space so ~75% land off-screen and are
+		// cullable. A fixed seed keeps the scene identical across runs.
+		items = new SKRect[ItemsPerFrame];
+		var rnd = new Random(42);
+		for (var i = 0; i < items.Length; i++)
+		{
+			var x = rnd.Next(-Width, Width * 3);
+			var y = rnd.Next(-Height, Height * 3);
+			items[i] = SKRect.Create(x, y, 24, 24);
+		}
+	}
+
+	[GlobalCleanup]
+	public void GlobalCleanup()
+	{
+		paint?.Dispose();
+		blur?.Dispose();
+		surface?.Dispose();
+	}
+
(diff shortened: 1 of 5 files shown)


taublast:
Awesome! Also maybe worth considering: SkImageFilter::computeFastBounds is public upstream, could bind `SkImageFilter::computeFastBounds` directly on SKImageFilter too, for measuring a filter that isn't (yet) attached to a paint — handy when sizing a cache before constructing the final paint.

PR #4306, opened 2026-07-01
[skia-sync] Drive upstream sync from versions.json with a run-number rotation

mattleibow:
## What

Reworks the `auto-skia-sync` upstream-sync automation to be driven by
`scripts/infra/docs/versions.json` instead of a fixed `current`/`next`/`latest`/`main`
mode map, simplifies the trigger surface, and fixes a related `update-versions.ps1`
bug surfaced while validating the main-tip path.

### Before
- Three daily crons mapped to `current` (main's milestone), `next` (+1), `latest`
  (highest upstream `chrome/m*`), plus dispatch-only `main`.
- The **supported stable lines were never auto-synced** — e.g. `release/4.148.x` and
  `release/4.150.x` only got bug-fix syncs if a human dispatched them by number.

### After
- **One `0 */6 * * *` cron.** Scheduled runs **rotate**: each run syncs one supported
  line from `versions.json`'s `support` block (`stable` ascending, then `preview`
  ascending), plus a trailing **`next`** = highest supported milestone + 1 (a real
  milestone bump into `main` if upstream cut `chrome/m<next>`, else the upstream `main`
  tip on `skia-sync/main`).
- **Single dispatch input `target`:** empty = rotate, `main` = upstream tip, or a
  milestone number (e.g. `151`) = explicit sync. The old `mode`/`milestone` inputs are gone.
- Round-robin index = `GITHUB_RUN_NUMBER % len`, so coverage self-adjusts to however many
  lines `versions.json` lists (verified for 2/4/5 lines).

### Correctness (from a GPT-5.5 review)
The detector runs twice per run — the pre_activation `--gate` step and the agent-job
align step (separate jobs that can't share outputs). Three fixes ensure they never diverge:
- Rotation keyed on `GITHUB_RUN_NUMBER` (identical across jobs) — no wall-clock read.
- Main's config read at the immutable `GITHUB_SHA`, not the mutable branch ref.
- `chrome_branch_exists` fails hard on a `git ls-remote` error instead of silently
  falling back to `main`.
- A supported line older than `main` with no `release/*.x` branch is skipped, not
  merged into `main`.

## 2. Fix `update-versions.ps1` writing a literal ref (tip mode)

Surfaced by the main-tip validation run: `update-versions.ps1` resolved the merged
upstream commit with a hardcoded `git rev-parse "upstream/chrome/m$Target"`. On a
main-tip sync the merged ref is `upstream/main`, so `upstream/chrome/m<Target>` doesn't
exist — and plain `git rev-parse` on a missing ref **echoes the literal argument to
stdout and exits non-zero**. The guard was only an empty-string check, so the literal
`upstream/chrome/m151` got written into `cgmanifest.json`'s `upstream_merge_commit` as
if it were a SHA (the agent had to correct it by hand).

- Add an optional `-UpstreamRef` parameter (defaults to `chrome/m<Target>`).
- Resolve via `git rev-parse --verify --quiet "upstream/<ref>^{commit}"` + `$LASTEXITCODE`,
  so a missing ref fails cleanly to `UNKNOWN` instead of writing the literal.
- Document it in the update-skia SKILL.md Phase 6 and pass `-UpstreamRef "${upstream_ref}"`
  from the workflow.

## Files
- `.github/scripts/skia-sync-detect.sh` — rotation logic, `--target` interface, hardening.
- `.github/workflows/auto-skia-sync.md` — single cron, single `target` input, Phase 6 note.
- `.github/workflows/auto-skia-sync.lock.yml` — recompiled via `gh aw compile` (frontmatter only;
  gh-aw materializes the prompt body at runtime, so body-only edits don't change the lock).
- `.agents/skills/update-skia/scripts/update-versions.ps1` — `-UpstreamRef` + robust ref resolution.
- `.agents/skills/update-skia/SKILL.md` — Phase 6 `-UpstreamRef` guidance.

## Testing
- `bash -n` clean; `rotate_select` simulated end-to-end (real `versions.json` + empty/missing
  `support` edge cases); PowerShell AST parse clean + resolution unit-tested (valid ref → SHA,
  missing ref → `UNKNOWN`); recompile is 0-error.
- Dispatched a rotate (`""`) run and a `151` run against this branch: both passes agreed on the
  target, produced the correct `skia-sync/main` / `skia-sync/m151` branch pairs, and finished
  green (native + C# build clean, tests **5584 passed / 0 failed**). Those throwaway validation
  PRs/branches have been closed and deleted.

### DIFF
--- a/.agents/skills/update-skia/SKILL.md
+++ b/.agents/skills/update-skia/SKILL.md
@@ -407,6 +407,11 @@ cd ../..  # back to parent repo (Phase 5 ends inside externals/skia)
 pwsh .agents/skills/update-skia/scripts/update-versions.ps1 -Current {CURRENT} -Target {TARGET}
 ```
 
+> **Main-tip sync?** When Phase 5 merged `upstream/main` (not a `chrome/m{TARGET}` branch),
+> add `-UpstreamRef main` so `cgmanifest.json`'s `upstream_merge_commit` is resolved from the
+> ref you actually merged. Without it the script defaults to `upstream/chrome/m{TARGET}`, which
+> does not exist on a tip merge, and `upstream_merge_commit` is left `UNKNOWN` for you to set by hand.
+
 The script handles all of these (so you don't have to do them manually):
 - `scripts/VERSIONS.txt`: milestone, increment→0, soname, assembly, file, ALL ~30 nuget lines
 - `cgmanifest.json`: commitHash, version, chrome_milestone, upstream_merge_commit
--- a/.agents/skills/update-skia/scripts/update-versions.ps1
+++ b/.agents/skills/update-skia/scripts/update-versions.ps1
@@ -18,8 +18,18 @@
 .PARAMETER Target
     The target Skia milestone number (e.g., 120)
 
+.PARAMETER UpstreamRef
+    The upstream ref that was actually merged, as named under the `upstream` remote
+    (without the `upstream/` prefix). Normally `chrome/m<Target>`; pass `main` for a
+    bleeding-edge main-tip sync (where no `chrome/m<Target>` branch is merged). Defaults
+    to `chrome/m<Target>`.
+
 .EXAMPLE
     pwsh .agents/skills/update-skia/scripts/update-versions.ps1 -Current 119 -Target 120
+
+.EXAMPLE
+    # main-tip sync (same milestone, merged from upstream/main):
+    pwsh .agents/skills/update-skia/scripts/update-versions.ps1 -Current 151 -Target 151 -UpstreamRef main
 #>
 
 param(
@@ -27,7 +37,9 @@ param(
     [int]$Current,
 
     [Parameter(Mandatory=$true)]
-    [int]$Target
+    [int]$Target,
+
+    [string]$UpstreamRef
 )
 
 $ErrorActionPreference = 'Stop'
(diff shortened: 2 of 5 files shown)


mattleibow:
<details>
<summary>📋 Suggested merge commit message (click to expand, then copy) — updated for both commits</summary>

```text
[skia-sync] Drive upstream sync from versions.json with a run-number rotation (#4306)

Context: scripts/infra/docs/versions.json (support.stable / support.preview)

The daily Skia upstream sync tracked milestones off a fixed mode map — three
crons resolving to `current` (main's milestone), `next` (+1) and `latest`
(highest upstream chrome/m*), plus dispatch-only `main`. That map never covered
the lines we actually ship: the supported stable maintenance lines (e.g.
release/4.148.x and release/4.150.x) only received bug-fix syncs when a human
manually dispatched them by number, so they silently drifted from upstream.

Make the sync self-describing instead. A single `0 */6 * * *` cron now ROTATES
over the hand-maintained `support` block in versions.json: each run syncs one
line — stables ascending, then previews ascending — plus a trailing `next`
(highest supported milestone + 1), which becomes a real milestone bump into main
when upstream has cut chrome/m<next>, or the upstream main tip (skia-sync/main)
when it has not. Coverage now follows versions.json automatically as lines are
added or retired.

The trigger surface collapses to one dispatch input `target`: empty = rotate,
`main` = upstream tip, or a milestone number (e.g. 151) = explicit sync. The old
`mode`/`milestone` inputs and the current/next/latest modes are removed.

~~ Keeping the two detector passes in agreement ~~

The detector (.github/scripts/skia-sync-detect.sh) runs twice per workflow run —
the pre_activation `--gate` pass and the agent-job "align submodule" pass — in
separate jobs that cannot share outputs. A GPT-5.5 review flagged three ways
those passes could diverge and align a different branch/submodule than the gate
approved. All three are closed:

  * the round-robin index is GITHUB_RUN_NUMBER % len, which is identical across
    every job of a run — no wall-clock read, so no chance of the two passes
    landing on different targets near a time boundary;
  * main's milestone and the support block are read at the immutable GITHUB_SHA
    rather than the mutable branch ref, so a mid-run push to main cannot change
    what the second pass sees;
  * chrome_branch_exists fails hard on a git ls-remote error instead of treating
    a transient failure as "branch absent" and silently falling back to main.

A supported line older than main's milestone with no matching release/*.x branch
is now skipped (nowhere valid to sync) rather than merged into main.

~~ Fix update-versions.ps1 writing a literal ref in tip mode ~~

Surfaced while validating the main-tip path: Phase 6's update-versions.ps1
resolved the merged upstream commit with a hardcoded
`git rev-parse "upstream/chrome/m$Target"`. On a main-tip sync the merged ref is
`upstream/main`, so `upstream/chrome/m<Target>` does not exist — and plain
`git rev-parse` on a missing ref echoes the literal argument to stdout and exits
non-zero. The guard was only an empty-string check, so the literal string
`upstream/chrome/m151` was written into cgmanifest.json's upstream_merge_commit
as if it were a SHA. Add an optional `-UpstreamRef` parameter (default
chrome/m<Target>), resolve with `git rev-parse --verify --quiet
"upstream/<ref>^{commit}"` + $LASTEXITCODE so a missing ref fails cleanly to
UNKNOWN, and pass `-UpstreamRef "${upstream_ref}"` from the workflow.

.github/workflows/auto-skia-sync.lock.yml is regenerated by `gh aw compile`; do
not edit it by hand. (gh-aw materializes the prompt body at runtime, so the
markdown-body edits here don't change the lock — only frontmatter does.)

Validated by dispatching the workflow against this branch: a rotate run
(deterministically resolved to the main-tip line) and an explicit target=151 run.
Both passes agreed on the target, produced the correct skia-sync/main and
skia-sync/m151 branch pairs, and finished green (native + C# build clean, tests
5584 passed / 0 failed). Those throwaway validation PRs/branches have been closed
and deleted.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
```

</details>

PR #4378, opened 2026-07-08
[release-notes] Scaffold mandatory sections + harden inline attribution

mattleibow:
Hardens the two prose rules that regressed on the first fully-automated CI run (bot PR #4373, polished by claude-opus-4.7), using the same structural approach that fixed the banner.

### 1. Scaffold the mandatory sections (fixes the dropped `## Breaking Changes` heading)

On that run, every "None in this release." page silently dropped its entire `## Breaking Changes` section (3.116.1, 3.118.0, 3.119.0/1/2/4, 3.119.5) — the AI omitted the section exactly when there was nothing to say. That case is fully deterministic: the script already knows whether a line has a breaking/notes companion.

`format_pr_list` now scaffolds `## Highlights` and `## Breaking Changes` into the skeleton the AI fills. With no companion it emits the whole `None in this release.` / `None in this preview line.` body verbatim (kept, never dropped); with a companion it emits a `TODO(polish)` marker to summarize. A leftover `TODO(polish)` is a greppable review failure, like a stray `<THEME>`.

### 2. Harden inline attribution (fixes the bare handles)

~37 bullets read `contributed by @maxkatz6 ❤️` (bare, unlinked, heart misplaced) instead of `❤️ [@maxkatz6](https://github.com/maxkatz6)`. Attribution is woven into prose so it can't be fully scaffolded, but the skeleton's AI comment now states the exact allowed form and bans the wrong ones inline, and rule 5 / TEMPLATE / self-review / review-checklist grow an explicit `contributed by @` ban with a grep.

### Rollout

Bumps the raw-data format version to 3, so a single run re-polishes every page with the new scaffold + instructions, then settles back to idempotent. Spec §4.4 updated. No generated content in this PR — it is source-only, so the next scheduled CI run produces the corrected pages content-only and opens the `[docs]` bot PR automatically.


mattleibow:
Superseded by the v2 three-script pipeline redesign (#4389). Mandatory-section scaffolding and inline-attribution hardening now live in the release-notes skill on dev/release-notes-v2.

PR #4428, opened 2026-07-14
Restore SK_ENABLE_LEGACY_SHADERCONTEXT to fix CPU raster perf regression (#4421)

### LINKED ISSUE #4421: [All, critical on Windows] Severe Software (CPU) Rasterization Performance Regression / Flickering starting in Milestone 146`
### Description

Starting in Milestone 146 (and affecting later releases including m147, m148, m150, m151), projects using software (CPU) rasterization—such as WinUI `SKCanvasView` backends or software-rasterized tile/UI rendering—suffer from severe performance degradation, high CPU usage, and visual flickering. The bug makes CPU rendering using SKCanvasView on Windows almost unusable, as pages with background images can sometimes take 1-2s to render with displaying black in the meanwhile.

We made a thorough analysis of the issue, and the regression was traced back to upstream Google Skia commit **`356c7267fa`** (*"Remove SK_DISABLE_LEGACY_SHADERCONTEXT"*), which removed the default definition of the `SK_ENABLE_LEGACY_SHADERCONTEXT` macro from `include/core/SkTypes.h`.

We also tested this with a custom built SkiaSharp just before the commit and using the commit to verify that this was in fact the cause. We then applied the below fix to a custom version of SkiaSharp 4.150 to ensure that it did indeed fix the issue.

### **Suggested Fix for SkiaSharp**
We should pass `-DSK_ENABLE_LEGACY_SHADERCONTEXT` in `extra_cflags` within the platform-specific build configurations in `native/*/build.cake` (specifically for Windows, and optionally other platforms) to ensure the fast legacy blitters remain active on modern Skia milestones:

```cake
extra_cflags=[ 
    '-DSKIA_C_DLL', 
    '-DSK_AVOID_SLOW_RASTER_PIPELINE_BLURS', 
    '-DSK_ENABLE_LEGACY_SHADERCONTEXT',  // <--- Add this define
    // ... other flags
]
```

@mattleibow Would you be able to fix in SkiaSharp 4.x as soon as possible? We currently need to use either 3.119 or a custom SkiaSharp 4.150 build for GnollHack on Windows. The issue may not be super-noticeable at first with small bitmaps, but it can completely throttle pages with SKCanvasViews on sometimes, making potentially the use of SKCanvasViews impossible in production on Windows.

#### **Technical Root Cause:**
1. **Fallback to Raster Pipeline**: When `SK_ENABLE_LEGACY_SHADERCONTEXT` is undefined, `SkShaderBase::makeContext` unconditionally returns `nullptr`. In `SkBlitter.cpp`, this forces the software blitter choice to abort the fast legacy path (`SkARGB32_Shader_Blitter`) and fall back to the modern **Raster Pipeline Blitter** (`CreateSkRPBlitter`).
2. **Huge CPU Overhead**: The legacy blitter uses highly optimized hand-written assembly/C++ loops (e.g. SSSE3 vector instructions or direct `memcpy` for repeating textures). The Raster Pipeline, on the other hand, is a generic virtual machine interpreter. It constructs a chain of 15–20 generic stage function pointers *per pixel group*.
3. **No `[[clang::musttail]]` on MSVC**: Since MSVC does not support guaranteed tail-calls, the raster pipeline cannot perform tail-call optimization on Windows. Each stage transition requires a full function call, exacerbating the performance penalty.
4. **Float Conversion**: The raster pipeline operates on 4-float vectors (RGBA), forcing expensive integer-to-float and float-to-integer conversions when drawing standard `BGRA_8888` bitmaps on CPU targets.

For applications rendering tiled layouts (e.g., backgrounds, borders) or scaled UI assets using CPU-based canvases, this fallback increases drawing overhead by **5x to 20x**, leading to severe lagging and flickering.

### **Code / Reproduction**
The regression is highly noticeable when rendering a tiled background or repeatedly drawing subset bitmaps in a loop on a CPU-based `SKCanvasView` on Windows (Android and iOS are negligibly impacted):

```csharp
// In a paint handler of a CPU-backed SKCanvasView
private void OnPaintSurface(object sender, SKPaintSurfaceEventArgs e)
{
    var canvas = e.Surface.Canvas;
    var paint = new SKPaint();

    // Draw repeating tiles over a large area
    for (int y = 0; y < height; y += tileHeight)
    {
        for (int x = 0; x < width; x += tileWidth)
        {
            canvas.DrawBitmap(tileBitmap, new SKRect(0, 0, tileWidth, tileHeight), new SKRect(x, y, x + tileWidth, y + tileHeight), paint);
        }
    }
}
```

#### **How to verify the fix:**
Building the native Skia library with `-DSK_ENABLE_LEGACY_SHADERCONTEXT` added to the compiler flags (`extra_cflags`) restores the fast-path legacy blitters, resolving the high CPU usage and flickering completely on milestone m150.

### Expected Behavior

Drawing tiled or scaled bitmaps on a CPU canvas should use the legacy blitter fast-path (e.g., tight loops of `memcpy` for repeating textures, or SIMD-accelerated pixel sampling), maintaining high frame rates and low CPU usage.

### Actual Behavior

Every software bitmap draw call falls back to the high-overhead Raster Pipeline, causing massive CPU bottlenecks, frame drops, and visible flickering. It practically makes CPU rendering via SKCanvasView almost unusable on Windows.

### Version of SkiaSharp

4.147-preview1.1 and later. Affects versions compiled with Skia milestone m146 and later (e.g. 4.148.0, 4.150.0).

3.119.4. Versions compiled with milestone m145 or earlier (where `SK_ENABLE_LEGACY_SHADERCONTEXT` was still defined by default).

### IDE / Editor

Visual Studio (Windows)

All, Windows

Windows (and potentially other platforms using CPU-rasterized fallbacks).

### Devices

Windows computers in particular.

### Relevant Screenshots

_No response_

### Relevant Log Output

```shell

```

### Code of Conduct

mattleibow:
Fixes #4421

## Summary

Restores the fast software (CPU) blitter path that regressed starting with Skia milestone **m146**, by passing `-DSK_ENABLE_LEGACY_SHADERCONTEXT` in `extra_cflags` for the CPU-raster native builds.

## Root cause — validated against m151 source

I verified the reporter's (@janne-hmp) analysis directly in the current submodule (`externals/skia`, m151):

1. **Upstream removed the default define.** Commit [`356c7267fa` — *"Remove SK_DISABLE_LEGACY_SHADERCONTEXT"*](https://skia.googlesource.com/skia/+/356c7267fa) deleted the block in `include/core/SkTypes.h` that used to define `SK_ENABLE_LEGACY_SHADERCONTEXT` by default. It is now **off by default**.
2. **`makeContext` now returns null.** With the macro undefined, `SkShaderBase::makeContext()` unconditionally returns `nullptr` (`src/shaders/SkShaderBase.cpp:107`).
3. **Blitter falls back to the raster pipeline.** In `src/core/SkBlitter.cpp`, a null shader context forces `CreateSkRPBlitter()` (generic raster-pipeline blitter, line 744) instead of the hand-optimized `SkARGB32_Shader_Blitter` (line 749), for `kN32` + `SrcOver` draws that carry a shader — e.g. `DrawBitmap` with a src/dst rect (scaled/tiled). On CPU surfaces this is much slower → high CPU, frame drops, and flicker, most visibly on Windows `SKCanvasView`.

**The define is the sanctioned opt-in.** Immediately before the removal, upstream landed [`1a15ed6b17` — *"Allow users to set SK_ENABLE_LEGACY_SHADERCONTEXT"*](https://skia.googlesource.com/skia/+/1a15ed6b17), explicitly making the macro user-settable so consumers could turn the path back on. So setting it via `extra_cflags` is the intended mechanism, not a hack.

## Additional notes

- **The "no `[[clang::musttail]]` on MSVC" point applies only to our internal `[MSVC]` variant.** SkiaSharp's *primary* shipped Windows binaries are built with **clang** (`clang_win`, LLVM installed by CI); the MSVC builds only run in the internal `devdiv` pipeline. The regression is nonetheless real on the clang builds (the reporter verified a custom build) — the raster pipeline is simply slower than the legacy blitter for N32 image-shader draws, independent of tail-call optimization. This `extra_cflags` fix benefits both toolchains.
- **This is a restore-of-default, and inherently temporary.** The legacy shader context is deprecated upstream; the raster pipeline is its intended replacement. There is no faster drop-in alternative today other than re-enabling the legacy path. When Skia eventually deletes it, the perf gap will need to be raised upstream (raster-pipeline optimization) instead.

## Changes

Added `-DSK_ENABLE_LEGACY_SHADERCONTEXT` next to the existing `SK_AVOID_SLOW_RASTER_PIPELINE_BLURS` perf flag in every CPU-raster `libSkiaSharp` build, restoring the exact pre-m146 default across all platforms:

| File | Covers (via delegation) |
|------|--------------------------|
| `native/windows/build.cake` | nanoserver |
| `native/macos/build.cake` | |
| `native/ios/build.cake` | maccatalyst |
| `native/tvos/build.cake` | |
| `native/linux/build.cake` | linuxnodeps |
| `native/android/build.cake` | |
| `native/tizen/build.cake` | |
| `native/wasm/build.cake` | (libSkiaSharp block only) |

Not touched: `native/winui` (WinRT projection helper, no raster), `native/winui-angle` (ANGLE GL driver).

## Benchmarks — cross-platform CI proof

`LargeImageScaleBenchmark` (#4465) exercises the regressing path with a small number of **large, full-surface upscaled** draws (the reporter's full-screen scenario), so per-pixel cost dominates rather than per-draw setup. CI builds the native lib **with** the define from this branch (via the `benchmark-source` legs, after the #4463 cache-key fix) and compares to the **without**-define nightly (mean):

| Platform | Draw (2048²) | OFF (raster pipeline) | ON (this PR) | Speedup |
|---|---|---:|---:|---:|
| **Windows** | opaque upscale | 15.03 ms | 2.89 ms | **5.2×** |
| **Windows** | cross-fade | 31.75 ms | 10.15 ms | **3.1×** |
| **macOS** | opaque upscale | 6.26 ms | 2.17 ms | **2.9×** |
| **macOS** | cross-fade | 13.46 ms | 5.29 ms | **2.5×** |
| **Linux** | opaque upscale | 50.41 ms | 2.85 ms | **17.7×** |
| **Linux** | cross-fade | 106.15 ms | 8.06 ms | **13.2×** |

The ON (legacy) times are stable across runs; the OFF (raster-pipeline) times inflate under CI-runner load (Windows steady ~5×, Linux 13–18×), consistent with the sustained-load throttling reported in #4421.

## Testing

Verified two ways: @janne-hmp installed the PR artifacts and confirmed the regression is resolved in their app, and the CI benchmark above shows a 3–18× speedup on the isolated path. Native build-configuration change; no managed code path to unit-test.


janne-hmp:
@mattleibow I just tested the PR artifacts 4.151.0-pr.4428.1, and they indeed work very well now. No bug observed anymore, so the fix works as intended.

PR #4456, opened 2026-07-16
perf: add nightly benchmark backfill mechanism

mattleibow:
## Why

The cross-platform perf tracker (#4438) persists benchmark history to the `aw-data` branch and renders an interactive dashboard, but `aw-data` currently holds only **one day** per leg — so the dashboard's `nightly` trend line is a single point. This adds a mechanism to **backfill ~2–3 weeks of historical nightly data** so the trend chart shows a real multi-week line, landed on a **scratch branch** for review before any promotion to `aw-data`.

## Approach

CI **measures** (runs BenchmarkDotNet against each old nightly on Linux/Windows/macOS and uploads one history artifact per leg); a maintainer then **downloads the artifacts, assembles locally, and pushes a scratch data branch**. The workflow therefore needs no write permission and never touches `aw-data`.

Each backfilled point measures an **old nightly** but is dated by that nightly's **real publish date** (from the EAP feed's registration endpoint), so the points land on the correct timeline.

## Changes (all additive; reuse `_common` + `track.py`)

- **`track.py`** — new optional `--date YYYY-MM-DD`. Defaults to today, so the daily path is unchanged; validated when provided.
- **`_common.py`** — new `published_dates()` reads the EAP `RegistrationsBaseUrl/3.6.0` registration index (one inline fetch) → `{version: publishedISO}`.
- **`backfill_dates.py`** — selects the newest N nightlies, **one per calendar day**, and prints `[{version, date}]` for the workflow matrix.
- **`combine_histories.py`** — **unions** per-`(OS, role)` days across the many single-day artifact files, **seeded from the current `aw-data` `index.json`** so existing legs/days (curr-stable/latest/prev-*, the existing nightly day) are preserved and extended. (`merge_summaries.py` overwrites on collision, so a union helper is needed.)
- **`.github/workflows/backfill-benchmarks.yml`** — `workflow_dispatch`-only (input `count`, default 19). `plan` → points JSON; `benchmark` → matrix `(version × 3 OS)`, mirrors the daily nightly leg (`write_version_props` → BDN → `track.py --date`), uploads one uniquely-named history artifact per leg. **No commit in CI** (`permissions: contents: read`), and it is **not mapped in `persist-aw-data.yml`**, so the daily green workflows are untouched.

## Baseline caveat

Every backfilled point is measured on **today's runners**. That makes the whole nightly line internally consistent version-over-version, but a different hardware baseline than the pre-existing/going-forward daily points — expected and acceptable for a visual trend seed.

## Not in scope

- `scripts/infra/perf/templates/dashboard.html` (owned by another session) is untouched.
- Promotion to `aw-data` is a documented, optional follow-up; this PR only provides the mechanism + scratch-branch workflow.

### DIFF
--- a/.github/workflows/backfill-benchmarks.yml
+++ b/.github/workflows/backfill-benchmarks.yml
@@ -27,6 +27,16 @@ on:
         required: false
         default: "19"
         type: string
+      since:
+        description: "Only include nightlies published on/after this date (YYYY-MM-DD)."
+        required: false
+        default: ""
+        type: string
+      until:
+        description: "Only include nightlies published on/before this date (YYYY-MM-DD)."
+        required: false
+        default: ""
+        type: string
   # On-demand escape hatch: also run on a PR that carries the 'run-backfill' label.
   # workflow_dispatch requires the workflow to already exist on the default branch, so
   # this lets the backfill be exercised from a PR before it lands on main. Normal PRs are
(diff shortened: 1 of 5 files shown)


mattleibow:
Closing: this was a one-off backfill. The data is fully merged into the aw-data branch (nightly legs now span 2026-05-01 → 2026-07-15, 48 days x 3 OS) and the daily tracker grows it from there, so the mechanism isn't needed going forward. Reopenable if we ever need to seed history for a new OS/benchmark.

PR #4487, opened 2026-07-20
[WASM] Fix net11 exception-handling mismatch in emsdk 5.0.6 native libs

mattleibow:
## Problem

A .NET 11 Blazor WebAssembly app referencing SkiaSharp links fine (PR #4459 routes net11.0+ to the emsdk 5.0.6 static lib) but **fails at runtime** in a modern browser:

```
CompileError: WebAssembly.compileStreaming(): module uses a mix of legacy and
new exception handling instructions (function #19801)
```

.NET 11 (preview 6+) links WASM apps with the **new** exnref exception-handling model (`WASM_LEGACY_EXCEPTIONS=0` → `try_table`/`throw_ref`). SkiaSharp's emsdk 5.0.6 `libSkiaSharp.a`/`libHarfBuzzSharp.a` are compiled with `-fwasm-exceptions` but **without pinning the instruction variant**, so they emit the **legacy** wasm-EH instructions (`try`/`catch`). V8 refuses to compile a module that mixes both schemes → the error.

## Fix

Compile the **5.0.6 (net11) variant only** with the new EH instruction set so it matches .NET 11:

- `native/wasm/build.cake`: add a `_newexc` emscripten feature that appends `-s WASM_LEGACY_EXCEPTIONS=0` to `extra_cflags` **and** `extra_cflags_cc` (a compile-time codegen flag — required both for C++ exceptions and for the wasm setjmp/longjmp in the C deps like libjpeg-turbo/libpng). Emscripten's `settings.js` documents that codegen-affecting settings must be passed at compile time when wasm object files are produced, which the `.a` build does.
- `scripts/azure-templates-stages-native-wasm.yml`: add `_newexc` to the four **5.0.6** matrix entries only.

**net8 (3.1.34) and net9/net10 (3.1.56) stay on legacy EH** to match their runtimes — those emsdks predate the exnref proposal. No packaging/targets changes are needed (`_`-prefixed feature tokens are excluded from the artifact path, and the net11.0+ → 5.0.6 selection already exists from #4459).

## Verification

Rebuilt the emsdk 5.0.6 `st` and `st,simd` libraries from source with this change and confirmed on a net11.0 Blazor WASM app in a modern browser:

| TFM | libSkiaSharp | Result |
|---|---|---|
| net11 | shipped `4.151.0-preview.3.2` (legacy EH) | ❌ `CompileError … mix of legacy and new exception handling instructions` |
[...]

PR #4488, opened 2026-07-20
[WASM] Add preview-.NET WASM test job to catch exception-handling regressions

mattleibow:
## What

Adds **automated preview-.NET WebAssembly regression coverage** by running the **existing** `tests/SkiaSharp.Tests.Wasm` suite as the preview .NET (currently net11) in CI. It is off by default; only this one test job opts in, so the rest of CI stays on net10 and the shipping TFMs are unchanged.

## Why

The emsdk 5.0.6 WASM `libSkiaSharp` can emit **legacy** wasm exception-handling instructions while the preview .NET links WASM apps with the **new** exnref model (`WASM_LEGACY_EXCEPTIONS=0`). Mixing them makes the WASM app fail **at runtime in the browser** with:

> `CompileError: WebAssembly.compileStreaming(): module uses a mix of legacy and new exception handling instructions`

There was **no CI test** that would have caught this. This PR is that guard: it runs the real WASM suite on the preview runtime, so an EH mismatch — or any preview-runtime WASM breakage — turns the build red.

## How

- **`source/SkiaSharp.Build.props`** — opt-in `UsePreviewTFM` toggle (default off): shifts the .NET version window forward one release (dropping the now-EOL oldest version) so projects build cleanly under a preview .NET SDK.
- **`scripts/infra/tests/tests-wasm.cake`** — `--previewtfm=true` runs the suite with `-p:UsePreviewTFM=true`. The cake is the run; it does not provision.
- **`scripts/azure-templates-jobs-bootstrapper.yml`** — reusable `installPreviewSdk` + `previewWorkloads` params (install the preview SDK, pin global.json to it, install the requested preview workloads) so other jobs can opt into a preview-runtime run.
- **`scripts/azure-templates-variables.yml`** — `DOTNET_VERSION_PREVIEW` + `DOTNET_WORKLOAD_VERSION_PREVIEW` (default `DOTNET_VERSION` stays net10).
- **`scripts/azure-templates-stages-test.yml`** — `tests_wasm_preview_linux` job (consumes the `native` artifact, `installPreviewSdk: true`, `previewWorkloads: wasm-tools`, runs `tests-wasm --previewtfm=true`). The net10 `tests_wasm_linux` job is unchanged.

## Validation

Locally verified end-to-end **both paths** (preview SDK, DeviceRunners headless browser, real 5.0.6 `st,simd` natives from CI artifacts):
- **Pass path** (new-EH `_wasmeh,_newexc,st,simd` lib): the module compiles and the suite runs (1180 total / 1081 passed / 0 failed / 99 skipped).
- **Fail path** (legacy-EH `_wasmeh,st,simd` lib, i.e. main today): the browser fails with the exact `module uses a mix of legacy and new exception handling instructions` CompileError and the job goes red.

## Note on branching

This is currently based on `main`, which has the 5.0.6 native builds but **not** the EH fix (PR #4487), so the `tests_wasm_preview_linux` job is **expected to be red here** — that red is the negative control proving the guard works. Once #4487 lands on main (5.0.6 built with `_newexc`) and this branch is rebased, it goes green.

### DIFF
--- a/scripts/azure-templates-jobs-bootstrapper.yml
+++ b/scripts/azure-templates-jobs-bootstrapper.yml
@@ -28,6 +28,8 @@ parameters:
   installLlvm: true                             # whether or not to install the LLVM compiler
   installEmsdk: false                           # whether or not to install the Emscripten SDK
   installNinja: false                           # whether or not to install the ninja build system
+  installPreviewSdk: false                      # whether to install the preview .NET SDK ($(DOTNET_VERSION_PREVIEW)) and pin global.json to it
+  previewWorkloads: ''                           # workloads to install under the preview .NET SDK (comma-separated, e.g. 'wasm-tools,android')
   installXcode: true                            # whether or not to install Xcode
   publishArtifacts: []                          # the additional artifacts to publish
   tools: []                                     # any additional .net global tools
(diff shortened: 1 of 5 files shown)


mattleibow:
### ⚠️ The `tests_wasm_preview_linux` job is expected to be RED here — this is the negative control

This PR is based on `main`, which has the 5.0.6 WASM native builds (#4459) but **not** the exnref exception-handling fix (#4487). So `main` builds `libSkiaSharp` with **legacy** wasm EH, and running the new net11 WASM test against it reproduces exactly the regression this test is meant to catch:

```
MONO_WASM: instantiate_wasm_module() failed CompileError:
WebAssembly.compileStreaming(): Compiling function #14265 failed:
module uses a mix of legacy and new exception handling instructions
```

The app never boots, so the DeviceRunners runner hits its 300s `--timeout` and exits non-zero → the job goes red (a bounded timeout failure, not a hang or a silent pass).

**Both paths were validated locally** with real `5.0.6/st,simd` natives pulled from CI artifacts:

| Native lib | Outcome |
|---|---|
| **new-EH** (`_wasmeh,_newexc,st,simd`, from #4487 build) | module compiles, suite runs **1180 total / 1081 passed / 0 failed / 99 skipped** → ✅ green |
| **legacy-EH** (`_wasmeh,st,simd`, current `main`) | `mix of legacy and new exception handling instructions` CompileError → ❌ red |

**This goes green once #4487 lands on `main`** (5.0.6 rebuilt with `_newexc` → new-EH) and this branch is rebased. The red here is proof the guard works.

PR #3968, opened 2026-05-12
Add SkiaSharp bindings for the Graphite GPU backend (Vulkan/Metal/Dawn)

ramezgerges:
**Description of Change**

Adds SkiaSharp bindings for Skia's **Graphite** GPU backend (Vulkan / Metal / Dawn-WebGPU). Depends on the companion C-shim PR in the skia submodule: [mono/skia#236](https://github.com/mono/skia/pull/236).

Graphite is Skia's next-generation GPU backend. Unlike Ganesh's auto-flushing draw stream, it uses an explicit **record → snap → insert → submit** model with a per-thread `Recorder`. The managed surface mirrors that shape.

**API Changes**

## New types

10 new public types under `binding/SkiaSharp/Gpu/Graphite/` (namespace `SkiaSharp`):

| Type | Wraps / role |
|---|---|
| `SKGraphiteContext` | `skgpu::graphite::Context` — static `CreateVulkan` / `CreateMetal` / `CreateDawn` factories (with/without options), `IsBackendAvailable(...)`, `CreateRecorder(...)`, `InsertRecording(...)`, `Submit(...)`, `RequestReadPixels(...)`, `CheckAsyncWorkCompletion()`, `FreeGpuResources()`, `PerformDeferredCleanup(...)`, `DeleteBackendTexture(...)` |
| `SKGraphiteRecorder` | `skgpu::graphite::Recorder` — `Snap()`, `CreateBackendTexture(...)`, `DeleteBackendTexture(...)` |
| `SKGraphiteRecording` | `skgpu::graphite::Recording` |
| `SKGraphiteBackendTexture` | `skgpu::graphite::BackendTexture` — `CreateVulkan` / `CreateMetal` / `CreateDawn` for wrapping a caller-allocated GPU texture; `IsValid`, `Backend` |
| `SKGraphiteTextureInfo` | `skgpu::graphite::TextureInfo` (per-backend construction); `IsValid`, `Backend`, `SampleCount`, `Mipmapped` |
| `SKGraphiteImageCache` | managed helper: an LRU image cache whose `FindOrCreate` satisfies Graphite's image-provider hook (Graphite does not auto-upload non-Graphite images the way Ganesh does — see below) |
| `SKGraphiteVkBackendContext` | `skgpu::VulkanBackendContext` (instance / physical device / device / queue + `GetProcedureAddress`) |
[...]

PR #4545, opened 2026-07-25
Add backend-neutral async RequestReadPixels to SKImage and SKSurface

mattleibow:
## Description

Adds a **backend-neutral** async *rescale-and-read-pixels* feature to `SKImage` and `SKSurface`, wrapping the new core Skia C API (companion: mono/skia#295).

The underlying C++ types (`SkImage::AsyncReadResult`, `SkImage::RescaleGamma`, `SkImage::RescaleMode`) are core and shared by the raster, Ganesh and Graphite backends — `SkSurface` merely *aliases* the `SkImage` ones. So this lands as a generic feature on `main`. The in-flight Graphite PR (#3968) can then drop its Graphite-prefixed `SKGraphiteAsyncReadResult` / `SKGraphiteRescaleGamma` / `SKGraphiteRescaleMode` and consume these instead, keeping only its genuinely Graphite-specific `Context`-based entry point.

Raster surfaces/images invoke the callback **synchronously**; Ganesh defers it until `GRContext.Submit()` + `GRContext.CheckAsyncWorkCompletion()` are pumped (per Skia's docs, async read is Ganesh-only and the Graphite path is the context API).

**Related issues**

Related to #3968

**Required skia PR**

Requires https://github.com/mono/skia/pull/295

**Areas affected**

- [ ] Native dependency or Skia update
- [ ] Views & integrations
- [ ] Performance
- [x] Tests
- [ ] Build, packaging, or CI

## Changes

**Public API** (additive only — no removals or signature changes):

```csharp
// SKImage / SKSurface (same shape on both)
public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, Action<SKImageReadPixelsResult> callback);            // default rescale: Src + Nearest
public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, SKImageRescaleGamma rescaleGamma, SKImageRescaleMode rescaleMode, Action<SKImageReadPixelsResult> callback);

// GRContext
public void CheckAsyncWorkCompletion ();

// new types
public sealed class SKImageReadPixelsResult : IDisposable
{
    public int PlaneCount { get; }
    public int GetPlaneRowBytes (int planeIndex);
    public ReadOnlySpan<byte> GetPlaneData (int planeIndex);         // raw, zero-copy view (may include row padding)
    public void CopyPlaneTo (int planeIndex, Span<byte> destination); // tightly-packed copy into your buffer (padding stripped, bounded)
    public byte[] ToArray (int planeIndex = 0);                       // tightly-packed byte[] that outlives the callback
    public SKImage ToImage ();                                       // owned SKImage (single-plane result)
    public SKBitmap ToBitmap ();                                     // owned SKBitmap (single-plane result)
}
public enum SKImageRescaleGamma { Src, Linear }
public enum SKImageRescaleMode  { Nearest, Linear, RepeatedLinear, RepeatedCubic }
```

**Behavior**

`SKImageReadPixelsResult` is a non-owning, callback-scoped view (it is *not* an `SKObject` — the native pointer is non-ref-counted and reused). It is invalidated when the callback returns and throws `ObjectDisposedException` from every member afterwards. `GetPlaneData` is the raw, zero-copy span; the `CopyPlaneTo` / `ToArray` / `ToImage` / `ToBitmap` helpers produce owned, tightly-packed copies that survive past the callback (and cannot over-read the native transfer buffer). On failure (e.g. an out-of-bounds `srcRect`) the callback is invoked with `null`. There is no user `context`/state argument — capture it in the closure.

Defaults are `RescaleGamma.Src` + `RescaleMode.Nearest` (the least-transform "as-is" choice); use `RepeatedLinear`/`RepeatedCubic` explicitly for higher-quality downscales.

## Testing

`dotnet test tests/SkiaSharp.Tests.Console` — 18 `SKAsyncReadPixelsTest` cases pass locally (macOS arm64):

- **Correctness** with a per-pixel colour pattern read from an **offset** `srcRect` (not `0,0..WxH`), validating every destination pixel for `GetPlaneData` / `CopyPlaneTo` / `ToArray` / `ToImage` / `ToBitmap`.
- **Padding** — the strided row copy is factored into an `internal CopyRows` and unit-tested with a fabricated padded source (real GL/raster RGBA reads are never row-padded), asserting the packed output drops the padding.
- **Edge cases** — exact / too-small / empty / oversized destinations; invalid plane indices; `ToImage`/`ToBitmap` outlive the callback.
- **Behavior** — raster fires synchronously; out-of-bounds `srcRect` → `callback(null)`; null callback → `ArgumentNullException`; every member throws `ObjectDisposedException` (never crashes) when used after the callback; `Dispose` is idempotent.
- **Ganesh** (`[GPU]`): completes with correct pixels after `Submit(sync)` + a `CheckAsyncWorkCompletion` pump, plus a GPU downscale. The async-proof (`done==false` before the pump) is asserted only when the backend actually defers, tolerating Skia's documented synchronous transfer-buffer fallback. Auto-skips when no GL context is available.

The full suite is green apart from one pre-existing, unrelated visual-golden delta (`ganesh-gl/DiagonalLines`); this change touches no rendering path.

## Future work — YUV (not blocked by this PR)

This ships the RGBA (`count() == 1`) read only, but the design is deliberately YUV-ready: the C API result type + accessors are multi-plane (`count()` can be 3/4), and the enums/proc are shared. `asyncRescaleAndReadPixelsYUV420` / `YUVA420` can be added purely additively later — see the follow-up note in the comments for the proposed C# surface.

### DIFF
--- a/binding/SkiaSharp/DelegateProxies.cs
+++ b/binding/SkiaSharp/DelegateProxies.cs
@@ -104,5 +104,19 @@ namespace SkiaSharp
 			var path = SKPath.GetObject (pathOrNull, false);
 			del.Invoke (path, *matrix);
 		}
+
+		private static partial void SKImageAsyncReadPixelsProxyImplementation (void* context, IntPtr result)
+		{
+			// The captured Action<IntPtr> is the closure built by SKImage/SKSurface.RequestReadPixels.
+			// `result` is non-owning and only valid for the duration of this invocation (it is IntPtr.Zero
+			// on failure); the closure must read all data before returning. This fires at most once, so the
+			// pinning handle is freed here.
+			var del = Get<Action<IntPtr>> ((IntPtr)context, out var gch);
+			try {
+				del.Invoke (result);
+			} finally {
+				gch.Free ();
+			}
+		}
 	}
 }
--- a/binding/SkiaSharp/GRContext.cs
+++ b/binding/SkiaSharp/GRContext.cs
@@ -176,6 +176,12 @@ namespace SkiaSharp
 			GC.KeepAlive (this);
 		}
 
+		public void CheckAsyncWorkCompletion ()
+		{
+			SkiaApi.gr_direct_context_check_async_work_completion (Handle);
+			GC.KeepAlive (this);
+		}
+
 		public void Flush (SKImage image)
 		{
 			if (image == null) {
--- a/binding/SkiaSharp/SKImage.cs
+++ b/binding/SkiaSharp/SKImage.cs
@@ -615,16 +615,19 @@ namespace SkiaSharp
 
 		// RequestReadPixels
 
-		public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, Action<SKImageAsyncReadResult> callback) =>
-			RequestReadPixels (info, srcRect, SKImageRescaleGamma.Src, SKImageRescaleMode.RepeatedLinear, callback);
+		public void RequestReadPixels (SKImageInfo info, Action<SKImageReadPixelsResult> callback) =>
+			RequestReadPixels (info, SKRectI.Create (info.Width, info.Height), SKImageRescaleGamma.Src, SKImageRescaleMode.Nearest, callback);
 
-		public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, SKImageRescaleGamma rescaleGamma, SKImageRescaleMode rescaleMode, Action<SKImageAsyncReadResult> callback)
+		public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, Action<SKImageReadPixelsResult> callback) =>
+			RequestReadPixels (info, srcRect, SKImageRescaleGamma.Src, SKImageRescaleMode.Nearest, callback);
+
+		public void RequestReadPixels (SKImageInfo info, SKRectI srcRect, SKImageRescaleGamma rescaleGamma, SKImageRescaleMode rescaleMode, Action<SKImageReadPixelsResult> callback)
 		{
 			if (callback == null)
 				throw new ArgumentNullException (nameof (callback));
 
 			Action<IntPtr> handler = raw => {
-				using var result = raw == IntPtr.Zero ? null : new SKImageAsyncReadResult (raw);
+				using var result = raw == IntPtr.Zero ? null : new SKImageReadPixelsResult (raw);
 				callback (result);
 			};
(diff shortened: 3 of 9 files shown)


mattleibow:
### Follow-up: YUV / YUVA read (additive — not blocked by this PR)

This PR wires only the RGBA (`count() == 1`) read, but nothing here blocks YUV. The result type and its accessors are already **per-plane** and the C API result/enums/proc are shared, so YUV lands as a pure addition.

**Native (mono/skia), reusing `sk_image_async_read_result_t` + the shared enums/proc:**

```c
SK_C_API void sk_image_async_rescale_and_read_pixels_yuv420(
    const sk_image_t* image, sk_yuvcolorspace_t yuvColorSpace, const sk_colorspace_t* dstColorSpace,
    const sk_irect_t* srcRect, int32_t dstWidth, int32_t dstHeight,
    sk_image_rescale_gamma_t gamma, sk_image_rescale_mode_t mode,
    sk_image_async_read_pixels_proc callback, void* context);
// + _yuva420 (4 planes), + the sk_surface_* and graphite Context equivalents
```

**Managed (additive overloads):**

```csharp
// SKImage / SKSurface
public void RequestReadPixelsYuv420 (SKImageInfo info, SKYuvColorSpace yuvColorSpace, SKColorSpace dstColorSpace,
    SKRectI srcRect, Action<SKImageReadPixelsResult> callback);
public void RequestReadPixelsYuva420 (...);
// (SKYuvColorSpace would be a new generated enum mapping SkYUVColorSpace)
```

The delivered `SKImageReadPixelsResult` already covers the multi-plane case:
- `PlaneCount` → `3` (YUV420) or `4` (YUVA420); `GetPlaneData(i)` / `GetPlaneRowBytes(i)` / `CopyPlaneTo(i, …)` / `ToArray(i)` are all per-plane.
- `ToImage()` / `ToBitmap()` deliberately throw for multi-plane results (a YUV result isn't a single interleaved image); a `YuvToImage()` / planar-`SKImage.FromYuvaPixmaps` helper could be added if wanted.

**Only internal change required:** the result currently captures the single dst `SKImageInfo`, which is used to size `GetPlaneData`/`ToArray`/`CopyPlaneTo`. YUV planes are subsampled (chroma is half-size), so the result would need to carry **per-plane dimensions** instead. That state is all `internal`, so it can change without touching the public API surface shipped here.

PR #4585, opened 2026-07-30
Add GRVkMemoryAllocator managed binding (#4567)

### LINKED ISSUE #4567: [FEATURE] Expose Vulkan memory allocators for Ganesh and Graphite

Skia m152 changed its Vulkan embedding contract: clients must populate `skgpu::VulkanBackendContext::fMemoryAllocator` before creating either a Ganesh or Graphite context.

SkiaSharp's native C struct contains an opaque `gr_vk_memory_allocator_t*` slot, but there is no public C API that creates an allocator and no managed `GRVkMemoryAllocator` object that callers can assign. `GRVkBackendContext.ToNative()` therefore always leaves the allocator null.

To preserve existing behavior, SkiaSharp now creates Skia's default VMA-backed allocator internally for **both Vulkan backends**:

- **Graphite:** `src/c/sk_graphite_vulkan.cpp`
- **Ganesh:** `src/c/gr_context.cpp`, added as part of mono/skia#300 / mono/SkiaSharp#4564 after m152 removed Ganesh's old `SK_USE_VMA` fallback

This compatibility behavior is correct, but both paths currently depend on `src/gpu/vk/vulkanmemoryallocator/VulkanMemoryAllocatorPriv.h`. Google still maintains that implementation and uses it in examples/tools, but deliberately treats it as a private module because allocator policy belongs to the embedder.

Expose a supported Vulkan allocator abstraction that can be shared by both Ganesh and Graphite while retaining automatic default creation for backward compatibility.

A likely first API surface is:

- a reference-counted native C API for creating/ref/unref of the default VMA-backed allocator;
- a managed `GRVkMemoryAllocator` wrapper, such as `GRVkMemoryAllocator.CreateDefault(...)`;
- `GRVkBackendContext.MemoryAllocator` support in the managed-to-native conversion;
- equivalent allocator input for Graphite's Vulkan context creation;
- explicit lifetime tests proving that Ganesh and Graphite retain the allocator safely;
- platform coverage for Android, Linux, Windows and Apple Vulkan consumers.

On Android, investigate using Google's public `SkiaVMA::Make()` wrapper. For other platforms, decide whether SkiaSharp should continue wrapping Skia's internal factory or own a small native VMA adapter so the public SkiaSharp API does not depend directly on a private upstream path.

The initial API should expose the ready-made native allocator, not managed allocation callbacks. A later API could permit fully custom native allocators if a concrete customer requires one.

1. Continue creating the allocator only inside each C context factory. This is simple and preserves compatibility, but gives advanced users no control and duplicates allocator setup across Ganesh and Graphite.
2. Restore Google's deleted `SK_USE_VMA` fallback inside Ganesh. This would fight upstream's new embedder-owned contract and would not solve Graphite consistently.
3. Implement a complete allocator in managed callbacks. This would require allocation, free, map, flush and invalidate callbacks across P/Invoke and native worker threads, creating substantial exception, lifetime and device-loss risk.
4. Require every caller to implement a native allocator. This follows upstream literally but would break existing SkiaSharp users and is impractical for typical .NET applications.

### Additional context

Google's public contract is `skgpu::VulkanMemoryAllocator`, stored as `sk_sp<...>` in `VulkanBackendContext::fMemoryAllocator`. Vulkan itself only provides raw `vkAllocateMemory`, `vkFreeMemory`, mapping and synchronization primitives; the allocator object is a Skia interface implemented by VMA or by an embedder-specific allocator.

Relevant upstream behavior:

- Google removed the legacy Ganesh VMA fallback in commit `5e183e5aea` (`Remove legacy_ganesh_vulkan target that had VMA baked in`).
- Google examples now assign `VulkanMemoryAllocators::Make(...)` before calling `MakeVulkan`.
- Android exposes `SkiaVMA::Make(...)` as a public convenience wrapper.
- Skia's private VMA module explicitly states that it is not intended as a one-size-fits-all public allocator.

Related PRs:

- Graphite C API and native allocator setup: https://github.com/mono/skia/pull/236
- Graphite managed bindings: https://github.com/mono/SkiaSharp/pull/3968
- m152 Ganesh allocator compatibility fix: https://github.com/mono/skia/pull/300
- m152 SkiaSharp update: https://github.com/mono/SkiaSharp/pull/4564

### Code of Conduct

ramezgerges:
## Description

Wraps the new `gr_vk_memory_allocator_make_default` C API so managed callers can supply a Vulkan memory allocator to both Ganesh and Graphite.

Motivation: Skia m152 removes Ganesh's internal `SK_USE_VMA` fallback (upstream commit `5e183e5aea`), after which any null allocator slot causes `gr_direct_context_make_vulkan` to fail. Graphite hits the same failure today via its own null check. Without a managed way to construct an allocator, SkiaSharp callers cannot fill the slot even though the C struct already exposes it.

**Related issues**

Fixes #4567

**Required skia PR**

Requires https://github.com/mono/skia/pull/307

**Areas affected**

- [ ] Native dependency or Skia update (libpng, HarfBuzz, FreeType, zlib, milestone bump, …)
- [ ] Performance
- [x] Tests
- [ ] Build, packaging, or CI

## Changes

**Public API**

ABI-stable additive changes only.

```csharp
// new type
public unsafe class GRVkMemoryAllocator : SKObject
{
    public static GRVkMemoryAllocator CreateDefault(
        IntPtr vkInstance,
        IntPtr vkPhysicalDevice,
        IntPtr vkDevice,
        uint maxApiVersion,
        GRVkGetProcedureAddressDelegate getProcedureAddress,
        bool threadSafe = false,
        bool protectedContent = false);
}

// new properties
public GRVkMemoryAllocator GRVkBackendContext.MemoryAllocator { get; set; }
public GRVkMemoryAllocator SKGraphiteVkBackendContext.MemoryAllocator { get; set; }
```

**Behavior**

- Contexts take an internal `sk_ref_sp` on the supplied allocator, so callers can `Dispose()` the managed wrapper immediately after `GRContext.CreateVulkan` / `SKGraphiteContext.CreateVulkan` returns.
- When `MemoryAllocator` is left null, the native shim continues to auto-create Skia's default VMA-backed allocator (same behavior as before this PR for Graphite; new safety net for Ganesh once m152 lands).
- Get-proc delegate lifetime scoped to `CreateDefault` only — VMA caches resolved Vulkan function pointers, so the `GCHandle` is released before the method returns.

## Testing

New `tests/VulkanTests/GRVkMemoryAllocatorTest.cs`:

- `CreateDefaultRejectsNullHandlesAndGetProc` — argument validation
- `CreateDefaultReturnsAllocator` — happy path on real Vulkan context
- `GaneshContextRetainsAllocatorAfterWrapperDispose` — dispose the wrapper after `CreateVulkan`, then draw + flush to prove Ganesh kept its own ref
- `GraphiteContextAcceptsCustomAllocator` — same pass-through on Graphite, plus a `Sync` submit

Locally on Linux/Lavapipe (Vulkan software rasterizer): 4/4 new tests pass, full `SkiaSharp.Vulkan.Tests` suite 27 pass / 2 skip / 0 fail.

Not yet tested on Windows Vulkan or Android Vulkan. The managed API and native impl are platform-independent; the only per-platform variable is whether the driver exposes protected-content memory (`protectedContent: true`), which none of these tests exercise.

## Checklist

- [x] Native change? Companion `mono/skia` PR linked above and bindings regenerated

### DIFF
--- a/binding/SkiaSharp/GRVkBackendContext.cs
+++ b/binding/SkiaSharp/GRVkBackendContext.cs
@@ -70,6 +70,15 @@ namespace SkiaSharp
 
 		public bool ProtectedContext { get; set; }
 
+		// Optional caller-provided memory allocator. When null, the C shim
+		// auto-creates Skia's default VMA-backed allocator at context creation
+		// time (see gr_context.cpp / gr_vk_allocator.cpp) so pre-#4567 callers
+		// keep working. Callers who need to share or tune the allocator can
+		// build one via GRVkMemoryAllocator.CreateDefault and assign it here;
+		// the Context takes its own internal ref, so it is safe to dispose the
+		// wrapper immediately after GRContext.CreateVulkan returns.
+		public GRVkMemoryAllocator MemoryAllocator { get; set; }
+
 		internal GRVkBackendContextNative ToNative () =>
 			new GRVkBackendContextNative {
 				fInstance = VkInstance,
@@ -83,6 +92,7 @@ namespace SkiaSharp
 				fDeviceFeatures2 = VkPhysicalDeviceFeatures2,
 				fGetProcUserData = getProcContext,
 				fGetProc = getProcContext is not null ? DelegateProxies.GRVkGetProcProxy : null,
+				fMemoryAllocator = MemoryAllocator?.Handle ?? IntPtr.Zero,
 				fProtectedContext = ProtectedContext ? (byte)1 : (byte)0
 			};
 	}
--- a/binding/SkiaSharp/GRVkMemoryAllocator.cs
+++ b/binding/SkiaSharp/GRVkMemoryAllocator.cs
@@ -1,0 +1,68 @@
+#nullable disable
+
+using System;
+
+namespace SkiaSharp
+{
+	// Public wrapper around skgpu::VulkanMemoryAllocator (see gr_vk_allocator.h).
+	// Implemented as ISKReferenceCounted so SKObject's dispose path calls
+	// sk_refcnt_safe_unref on our handle — that's the same virtual-refcount
+	// machinery every other SkRefCnt-derived Skia type uses in these bindings.
+	//
+	// Ganesh (GRVkBackendContext.MemoryAllocator) and Graphite
+	// (SKGraphiteVkBackendContext.MemoryAllocator) both consume this handle.
+	// A Context created from a backend context takes its own internal ref, so
+	// disposing the allocator after context creation is safe.
+	public unsafe class GRVkMemoryAllocator : SKObject, ISKReferenceCounted, ISKSkipObjectRegistration
+	{
+		internal GRVkMemoryAllocator (IntPtr handle, bool owns)
+			: base (handle, owns)
+		{
+		}
+
+		// Build Skia's default VMA-backed allocator. `getProcedureAddress` is used
+		// synchronously during allocator construction to resolve Vulkan entry points
+		// — its GCHandle can be released before this method returns because VMA
+		// caches the resolved function pointers, not the getter itself.
+		//
+		// Returns null if Vulkan support is not compiled into libSkiaSharp or if
+		// the underlying VMA initialisation refused the provided device/instance.
+		public static GRVkMemoryAllocator CreateDefault (
+			IntPtr vkInstance,
+			IntPtr vkPhysicalDevice,
+			IntPtr vkDevice,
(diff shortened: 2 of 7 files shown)


ramezgerges:
Rebased onto the current `main` — includes #4612 (deterministic bindings regen). The two failed `/azp run` attempts happened while I was force-pushing; the branch is stable now.

PR #4586, opened 2026-07-30
Fix Graphite/Metal GradientBlend on the iOS simulator + shader-error diagnostic API (#4555)

### LINKED ISSUE #4555: [BUG] Graphite/Metal: Recorder.Snap() returns null for the gradient scene on the iOS simulator
### Description

Graphite/Metal (the `SKGraphiteContext.CreateMetal` path) cannot render a gradient-shader scene on the **iOS simulator**: `recorder.Snap()` returns `null` for the `GradientBlend` visual scene, surfacing as `InvalidOperationException: Recorder.Snap() returned null.` and failing the iOS visual test leg.

It is very narrowly scoped — the same scene renders correctly everywhere else:

| Backend | Host | GradientBlend |
| --- | --- | --- |
| **Graphite**/Metal | **iOS simulator** | ❌ `Snap()` returns null |
| Ganesh/Metal | iOS simulator (same host) | ✅ renders (`ganesh-metal.ios` golden passes) |
| Graphite/Metal | macOS | ✅ renders (`graphite-metal.macos` golden passes) |

So it is specifically **Graphite + gradient pipeline + iOS simulator**. The simulator's Metal appears unable to compile the pipeline Graphite emits for the gradient shader, while Ganesh's equivalent pipeline compiles fine on the same simulator. The other four visual scenes (DiagonalLines, FilledCircle, RedRoundedRectOnWhite, Text) render correctly on `graphite-metal.ios`, which is why that golden directory has 4 of 5 scenes.

Discovered during CI validation of #3968 (the Graphite backend PR).

### Code

The failing path is `GraphiteMetalRenderer.RenderAsync` in the test harness (`tests/Tests/SkiaSharp/Visual/Renderers/GraphiteMetalRenderer.cs`), simplified:

```cs
using var context  = SKGraphiteContext.CreateMetal(new SKGraphiteMtlBackendContext { MtlDevice = device, MtlQueue = queue });
using var recorder = context.CreateRecorder();
using var surface  = SKSurface.Create(recorder, info);

scene.Draw(surface.Canvas);            // GradientBlend: a linear gradient + a blended circle

using var recording = recorder.Snap(); // <-- returns null ONLY for the gradient scene, ONLY on the iOS simulator
```

Repro: run the `SkiaSharp.Tests.Devices` iOS visual tests (`SkiaSharp.Tests.Visual.Tests.VisualMatrixTests.RenderMatchesGolden`) on an iOS simulator; the `graphite-metal` × `GradientBlend` case throws. The scene is `tests/Tests/SkiaSharp/Visual/Scenes/GradientBlendScene.cs`.

### Expected Behavior

`recorder.Snap()` returns a valid recording for the gradient scene on the iOS simulator (as Ganesh/Metal does on the same simulator, and as Graphite/Metal does on macOS), so the scene renders.

### Actual Behavior

`recorder.Snap()` returns `null` → `InvalidOperationException: Recorder.Snap() returned null.`, failing the visual test.

### Scope

This issue is **only** the gradient pipeline compile failure. The other Metal problems on our CI hosts are tracked separately, because they have different causes and different fixes:

- #4598 — macOS: virtualized Metal leaves state that hangs the test host on shutdown
- #4599 — Mac Catalyst: `MTLDevice` reports none of the GPU families Skia requires
- #4591 — umbrella for re-enabling every GPU backend currently opted out of CI

### Version of SkiaSharp

Nightly / CI build — the Graphite Metal backend is new in #3968 (never shipped).

N/A — new backend; no prior good version.

### IDE / Editor

N/A — surfaced in CI.

iOS

iOS 26.2 simulator

ramezgerges:
## Description

Fixes #4555 (Graphite/Metal `GradientBlend` cell returning null `Recorder.Snap()` only on the iOS simulator) and adds the C API surface used to diagnose it.

Two commits, two roles:

1. **`SKGraphiteContext` shader-error-handler overloads** — wraps the new `sk_graphite_context_options.fShaderErrorHandler` C API. New backwards-compatible `CreateDawn`/`CreateMetal`/`CreateVulkan` overloads take an `SKGraphiteShaderErrorHandlerDelegate`; delegate is pinned for the Context's lifetime and freed in `DisposeNative` after the native context tears down its handler pointer. The proxy never throws across the FFI boundary — a bad handler cannot crash Skia's shader-compile path.

2. **Fix Graphite/Metal `GradientBlend` on the iOS simulator** — bumps the submodule to pick up the two Skia-side fixes: PSO-creation failures now routed through `ShaderErrorHandler`, and framebuffer fetch disabled under `TARGET_OS_SIMULATOR`. Drops the simulator-only null-`Snap()` skip seam from `GraphiteMetalRenderer` (a null Snap is again a hard failure everywhere, so regressions can't hide behind the skip) and adds the newly-passing `graphite-metal.ios/GradientBlend.png` golden.

**Related issues**

Fixes #4555

**Required skia PR**

Requires https://github.com/mono/skia/pull/308

**Areas affected**

- [ ] Native dependency or Skia update
- [x] Rendering output / visual behavior
- [ ] Performance
- [x] Tests
- [ ] Build, packaging, or CI

## Changes

**Public API**

ABI-stable additive changes only.

```csharp
// new delegate
public delegate void SKGraphiteShaderErrorHandlerDelegate(string shader, string errors, bool shaderWasCached);

// new overloads on SKGraphiteContext
public static SKGraphiteContext CreateDawn(
    SKGraphiteDawnBackendContext backendContext,
    SKGraphiteContextOptions options,
    SKGraphiteShaderErrorHandlerDelegate shaderErrorHandler);

public static SKGraphiteContext CreateMetal(
    SKGraphiteMtlBackendContext backendContext,
    SKGraphiteContextOptions options,
    SKGraphiteShaderErrorHandlerDelegate shaderErrorHandler);

public static SKGraphiteContext CreateVulkan(
    SKGraphiteVkBackendContext backendContext,
    SKGraphiteContextOptions options,
    SKGraphiteShaderErrorHandlerDelegate shaderErrorHandler);
```

**Behavior**

- Graphite/Metal on iOS/tvOS simulator: `GradientBlend` (and any other draw needing a dst read with a non-coefficient blend — Multiply, Screen, Overlay, …) now renders correctly instead of returning null from `Recorder.Snap()`. Real Apple devices and macOS unchanged.
- The simulator-only null-`Snap()` skip in `GraphiteMetalRenderer` is removed — regressions to this class of bug now fail loudly.
- Callers who install a shader-error handler now also receive Metal PSO-creation failures (previously invisible outside `SkDebugf`).

## Testing

Full `SkiaSharp.Tests.Devices` run on iPhone 16 / iOS 26.1 simulator (arm64): 1219 total, 0 failed, 1138 succeeded, 81 skipped. `GradientBlend` cell rendering and matching its committed `graphite-metal.ios` golden.

`GraphiteVkContextAcceptsShaderErrorHandler` smoke test on Linux/Lavapipe (Vulkan): passes — proves no double-free / no `EntryPointNotFoundException` on Context teardown after installing a handler. The Metal simulator path itself needs an Apple host to reproduce, which is what the Devices-tests run above covered.

## Checklist

- [x] Native change? Companion `mono/skia` PR linked above and bindings regenerated


ramezgerges:
Rebased onto the current `main` and resolved the `GraphiteMetalRenderer.cs` conflict (kept the version that removes the null-Snap workaround, which is exactly what this PR fixes). The two failed `/azp run` attempts happened while I was force-pushing; the branch is stable now.

binding/SkiaSharp/SkiaApi.generated.cs
```diff
@@ -20558,9 +20601,16 @@ public bool SetBackendLabels {
 			set => fSetBackendLabels = value ? (byte)1 : (byte)0;
 		}
 
+		// public sk_graphite_shader_error_handler_t* fShaderErrorHandler
+		private sk_graphite_shader_error_handler_t fShaderErrorHandler;
+		public sk_graphite_shader_error_handler_t ShaderErrorHandler {
+			readonly get => fShaderErrorHandler;
+			set => fShaderErrorHandler = value;
```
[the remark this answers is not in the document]
mattleibow: Yeah, this one may be a case where we either make the whole struct internal and have a public version that has a delegate property. Or we make it have an internal public property that we override the native value in a copy in the Create*

I would maybe go more for a separate public/internal version and have the public be a real delegate and avoid the new overload in the API. This will also indicate that the delegate is a long-existing delegate past the duration of the method call.

tests/VulkanTests/GraphiteVkBackendContextTest.cs
```diff
@@ -53,5 +53,59 @@ public void GraphiteVkContextIsCreatedFromRawHandles()
...
+			using var recording = recorder.Snap();
+			Assert.NotNull(recording);
+			var status = graphiteContext.InsertRecording(recording);
+			Assert.Equal(SKGraphiteInsertStatus.Success, status);
+			Assert.True(graphiteContext.Submit(new SKGraphiteSubmitInfo { Sync = true }));
+
+			// On Lavapipe / any conformant Vulkan the built-ins compile cleanly, so the
+			// callback should not have fired — but the important assertion is that
+			// Dispose (which happens at using-scope exit) does not crash.
+			GC.KeepAlive(handler);
```
mattleibow: This test only verifies that installing the handler does not break context creation or disposal; it never verifies that the callback is actually invoked or that `shader`, `errors`, and `shaderWasCached` are forwarded correctly. Please add a deterministic failure path (or a focused native/proxy test) that asserts those callback arguments, so the core diagnostic API is covered rather than only its teardown path.

tests/VulkanTests/GraphiteVkBackendContextTest.cs
```diff
@@ -53,5 +53,59 @@ public void GraphiteVkContextIsCreatedFromRawHandles()
...
+		// for the Context's lifetime and does not crash the process on Dispose. The
+		// callback itself does not fire under normal draws (all built-in Skia pipelines
+		// compile cleanly on Lavapipe) — this smoke test just proves the wiring holds:
+		// no GC of the pinned delegate, no double-free of the leaked FfiShaderErrorHandler
+		// bridge, no EntryPointNotFoundException. The Metal-iOS-sim gradient failure
+		// path this API was added for (issue #4555) requires an iOS simulator device to
+		// reproduce.
+		[Trait(Traits.Category.Key, Traits.Category.Values.Gpu)]
+		[Fact]
+		public void GraphiteVkContextAcceptsShaderErrorHandler()
```
mattleibow: Please build this scenario on the existing `SKGraphiteReleaseTestsBase`/backend harness pattern instead of adding another Vulkan-specific bring-up sequence. That base already centralizes context/recorder ownership, backend texture setup, Snap/insert/submit, deferred-work pumping, and backend policy gating. Generalizing it to a `GraphiteBackendTestBase` would let shader-handler, release-callback, and future Graphite scenarios run consistently across Vulkan, Metal, and Dawn while keeping only device bring-up, texture creation, and disposal ordering backend-specific. It would also be a good place to make the shared helpers assert `Snap()`, `InsertRecording`, and `Submit` results rather than ignoring failures.

mattleibow (changes requested):
Requesting changes: the shader-handler test does not verify callback delivery, and the new backend-specific test setup should be consolidated with the existing Graphite backend harness so future Graphite scenarios can run consistently across Vulkan, Metal, and Dawn.

tests/VulkanTests/GraphiteVkBackendContextTest.cs
ramezgerges: Added `SKGraphiteShaderErrorHandlerTest` (3 proxy unit tests, in `SkiaSharp.Tests`) covering argument forwarding, `Marshal.PtrToStringAnsi` null-pointer normalization, and managed-exception swallowing across the FFI boundary. Also added `ShaderErrorHandlerWiringSurvivesFullRoundTrip` to `GraphiteBackendTestBase` so every backend that supports per-harness options exercises the real Context creation + Snap/Insert/Submit + Dispose loop with a handler installed and asserts it does not fire spuriously.

Driver-level rejection is not reachable on our CI drivers — Lavapipe and SwiftShader both accept every SkSL shader Skia emits, so I couldn't construct a deterministic firing path there. End-to-end delivery under a real driver rejection is verified by the iOS-simulator integration run described in the PR body.

tests/VulkanTests/GraphiteVkBackendContextTest.cs
ramezgerges: Done. Renamed `SKGraphiteReleaseTestsBase` → `GraphiteBackendTestBase` (and the inner harness `GraphiteReleaseHarness` → `GraphiteBackendHarness`), added an options-aware `CreateHarnessAsync(SKGraphiteContextOptions)` abstract on the base, and moved the shader-handler wiring test onto it so Metal + Vulkan run it uniformly. Dawn opts out via `SupportsPerHarnessOptions = false` since it shares the Context across tests (the non-yielding constraint). Subclasses renamed to `Graphite{Metal,Vulkan,Dawn}BackendTests`.

PR #4913, opened 2026-09-02
Add Apple pointer hover and scroll input

mattleibow:
## Description

Adds first-class passive pointer hover and wheel/two-finger trackpad scrolling to the shared Apple MAUI input path used by `SKCanvasView`, `SKGLView`, and the Mac Catalyst Metal path. Applications no longer need platform-specific adapters to receive these inputs through the existing `Touch` event.

The root cause was that Apple input was represented by one direct-contact `UIGestureRecognizer`, which only observed UIKit's `TouchesBegan`, `TouchesMoved`, `TouchesEnded`, and `TouchesCancelled` callbacks. UIKit exposes passive hover and indirect scrolling through separate recognizers, so those sources had no path into `SKTouchEventArgs`.

`SKTouchHandler` is now the coordinator for the complete Apple input path. It owns direct-contact, hover, and scroll recognizers as peers and manages their attachment, removal, state reset, and disposal together. Contact handling remains isolated in its own recognizer, preserving the existing touch semantics instead of layering pointer setup onto the contact recognizer.

Hover uses `UIHoverGestureRecognizer` and maps naturally to the existing contract:

- `Began` -> `SKTouchAction.Entered`
- `Changed` -> non-contact `SKTouchAction.Moved`
- `Ended` / `Cancelled` -> `SKTouchAction.Exited`
- mouse device type, unknown button, zero pressure, and coordinates scaled by the existing Canvas/GL/Metal proxy

Scrolling uses a scroll-only `UIPanGestureRecognizer` with all indirect scroll types enabled, direct touch types explicitly disabled, zero maximum direct touches, and touch cancellation disabled. Ordinary contact touch and mouse click-drag remain on the direct-contact recognizer. Scroll and pinch recognizers may recognize simultaneously, but broad simultaneity with contact input is intentionally not enabled.

`WheelDelta` is a pragmatic, best-effort vertical compatibility projection because the current public API is a one-dimensional `int`. The handler:

- treats 40 UIKit logical points as 120 nominal compatibility units;
- negates UIKit's down-positive Y translation so positive remains up/forward;
- preserves smaller whole integer deltas for precision scrolling;
- accumulates fractional remainder per handler, making the emitted sum invariant to callback partitioning;
- processes `Began`, `Changed`, and residual `Ended` translation;
- drops only a terminal sub-unit remainder and never forces +/-1 for sub-unit movement;
- keeps the latest hover location as the stable wheel focal point.

[...]

PR #4915, opened 2026-09-02
[infra] Replace the artifact-size check_run trigger with an hourly sweep

mattleibow:
## Description

`track-artifact-sizes` subscribed to `check_run: [completed]`. GitHub offers **no** server-side filter for that event — no app, name, or path filter — so every completed check run on the default branch started a runner, and filtering only happened once the job was already running.

Combined with `persist-aw-data` (which runs on `workflow_run` of this workflow and then publishes its *own* check run), the two formed a self-sustaining loop:

```
track-artifact-sizes → persist-aw-data (workflow_run)
     ↑                        ↓
  check_run ←────── "Persist to aw-data" check completes
```

Measured on live data:

| Signal | Value |
|---|---|
| Runs of each workflow | **40,000** |
| Useful work | **1 of 25** consecutive runs — 24 did nothing |
| Check runs on commit `60579244c4f` | **1,842 — every one `github-actions`, zero `azure-pipelines`** |
| Most repeated row on that commit | **477** identical `resolve` rows (plus 476 `pr`, 476 `nightly`) |
| An ordinary PR head, for contrast | 61 `azure-pipelines` vs 25 `github-actions` |
| One hour | **1** Azure check completion vs **608** Actions completions |

All figures above are read back from the API at the time of writing; the counts grew for as long as the loop ran, so earlier snapshots of this PR quoted smaller ones.

**A job-level `if:` cannot fix this** — a *skipped* job still publishes a check run. Each firing published three (`resolve`, `pr`, `nightly`), so gating would have kept feeding the loop while still paying for a runner.

Alternatives were measured and ruled out:

- **`check_suite`** — 1,842 check runs vs 1,657 suites on that same commit, and 1,654 of those suites are `github-actions`. Only ~1.11× fewer events, so the loop survives at suite granularity essentially intact.
- **`status`** — would be immune (Actions never creates commit statuses), but Azure DevOps posts **zero** commit statuses here; it uses the Checks API exclusively.
- **AzDO pushing to GitHub** — impossible for public PR builds, which get no GitHub credential.

### The replacement

**Hourly sweep of the Azure DevOps build list**, which is already the queue we need. One anonymous request returns every recent PR build, so discovery is **O(1) in open pull requests** rather than one call per PR:

```
builds?definitions=345&minTime=<now-2h>&queryOrder=finishTimeDescending
```

- **2-hour lookback on an hourly cron** — an entire sweep can be missed without losing a report.
- **Selection is by durable stamp, not the clock.** A `<!-- build=id.attempt -->` marker in the PR comment decides whether a build was reported, so a duplicate sweep is a no-op and a delayed one catches up. A time window used for *selection* would silently drop work a later sweep no longer covers.
- **The stamp carries the package stage's `attempt`** — AzDO reuses the build id when a stage is re-run, and `buildNumberRevision` is a *daily* counter that cannot distinguish one. Without this, corrected numbers after a re-run would be silently suppressed.
- **Watching the `Package NuGets` stage** reports as soon as the packages exist — measured **~1h30m before** the build itself finished.
- **No per-sweep cap.** Over 250 hours of real traffic the worst 2-hour window held **7 distinct PRs**, against a matrix limit of 256 — a cap would only add a silent-deferral path.
- **Only `schedule` and `workflow_dispatch`** — neither is an event Actions can emit into this workflow, so no trigger loop is possible by construction. `workflow_dispatch` remains the manual route for a single build or a backfill.

Run volume goes from ~3,600/day to **24/day**.

**Related issues**

Related to the dnceng migration follow-up audit.

**Required skia PR**

None.

**Areas affected**

- [ ] Managed API (`binding/`)
- [ ] Native / C API (`externals/skia/src/c`, `include/c`)
- [ ] Generated P/Invoke bindings
- [ ] Native dependency or Skia update (libpng, HarfBuzz, FreeType, zlib, milestone bump, …)
- [ ] Performance
- [x] Tests
- [x] Build, packaging, or CI

## Changes

None — CI/automation only. No public API or managed behavior change.

Observable automation behaviour that does change:

- PR size comments arrive from an hourly sweep rather than immediately on the Azure check. Because the sweep watches the *package stage* rather than the whole build, the report often still lands before the build finishes.
- The comment gains a hidden `<!-- build=id.attempt -->` stamp. The visible marker is unchanged, so existing comments still match.

## Testing

| Gate | Result |
|---|---|
| Artifact-size intake tests | **32** passed |
| Perf HTTP contract tests | **11** passed |
| YAML parse, all 28 workflows | 0 failures |
| `repo-deps.py validate` | PASS |

Verified against live Azure DevOps rather than only with fixtures:

- **One query covers everything** — 183 builds / 40 PRs in 2.5s; the 2h window is 90 KB in 0.9s. No per-PR iteration.
- **Sweep mode** correctly reports nothing to do when the only recent build failed.
- **Single-PR mode** (the manual backfill path) resolves PR #4913 → `1577884.1` — note `.1`, the real stage attempt, not the misleading `buildNumberRevision` of 29.
- **Package-stage timing** confirmed on a real build: package done 03:34, build finished 05:04.
- **Worst-case matrix width** measured at 7 PRs per 2h over 250 hours.

Tests pin the properties that matter, not just the happy path: that the workflow never re-subscribes to `check_run`/`check_suite` and subscribes *only* to `schedule` and `workflow_dispatch`, that both `fromJSON` inputs are guarded against `''` (matrix expands *before* the job `if:`, so an empty string fails the run and would take the nightly persist with it), that a re-run produces a different stamp, that only the newest build per PR is considered, and that an unreadable comment *skips* a PR rather than re-measuring it.

### Scope

This PR was split and now covers **only** the tracking workflow. The other work was raised separately and has since landed: the docs force-push guard (#4930), the workflow hygiene fixes (#4931), the ci-status accuracy work (#4932), the `Push-ReleaseTag` coverage (#4933), and `automation-tooling-tests.yml` itself (#4939). The `manage-nuget-feed` retirement (#4929) is still open. Because #4939 landed that workflow first, this PR's change to it is now purely additive: two perf test steps and one `paths:` entry, with no deletions.

## Checklist

- [ ] Native change? Companion `mono/skia` PR linked above and bindings regenerated


mattleibow:
```text
[infra] Break the artifact-size check_run loop and repair CI automation (#4915)

Context: https://github.com/github/gh-aw/releases/tag/v0.87.10

Three independent pieces of repository automation were failing at once, each
with its own root cause. The governing rule applied throughout is that an
automation trigger must be narrow enough to be evaluated before a runner
starts: GitHub cannot filter `check_run`, so the only way to stop the fan-out
is to stop subscribing to the event, and the only way to keep the behaviour it
provided is to ask Azure DevOps directly on a bounded schedule.

~~ A cross-workflow check_run loop, not an external fan-out ~~

`track-artifact-sizes` subscribed to `check_run: [completed]`. That event has
no app, name, or path filter, so every completed check run on the default
branch started a runner and the filtering only happened afterwards. Both
`track-artifact-sizes` and `persist-aw-data` had reached 40,000 runs.

The obvious reading — that broad external Azure checks drive this — is
wrong.
The commit carried 1,288 check runs, of which 1,229 were `github-actions` and
only 61 `azure-pipelines`, and the hour after the push saw 153 track runs
against exactly ONE Azure check completion and 608 Actions check completions.
Growth was steady at ~150/hr rather than exponential, which is only consistent
with GitHub suppressing a workflow re-triggering itself through its own Actions
check runs while leaving a cross-workflow cycle unsuppressed:
`track-artifact-sizes` -> `persist-aw-data` (`workflow_run`) -> its
`Persist to aw-data` check run completes -> `check_run` -> back to the start.
That relationship is 1:1, which is exactly the observed rate.

The trigger is removed rather than narrowed, because narrowing is not
available at any level that would help.

  * A job-level `if:` cannot fix this. Each firing published three check runs
    — `resolve` succeeded while `pr` and `nightly` were skipped — and a
    skipped job still publishes a check run, so the gate would have kept
    feeding the cycle while still paying for a runner.

  * PR reports are preserved by `scripts/infra/perf/sizes/find_pr_builds.py`,
    polled every 30 minutes, which asks for exactly the completed and
    succeeded build of pipeline 345 on `refs/pull/<n>/merge`. Worst-case
    latency becomes ~30 minutes; `workflow_dispatch` remains the manual path.

  * `render_pr_md.py` now emits a hidden `build=` stamp beside the unchanged
    visible marker so the poll can skip a build it already reported. Each
    measurement downloads roughly a gigabyte, so the stamp is written on
    failure too — otherwise an unmeasurable build would be re-selected and
    re-downloaded every 30 minutes indefinitely, recreating in miniature the
    runaway this change exists to remove.

  * `resolve` must emit `[]` rather than an unset output on the nightly path.
    GitHub expands `strategy.matrix` before the consuming job's `if:`, so
    `fromJSON('')` fails the run before the guard can skip it, and because
    `persist-aw-data` only persists a successful `workflow_run`, that failure
    would have silently stopped the nightly size history from being written.

  * `gh pr list` is assigned to a variable before the loop. `set -e` does not
    check a command substitution used in a `for` word list, so an auth or
    outage blip would otherwise produce a permanently green, permanently idle
    sweep. Widespread Azure query failure is now reported for the same
    reason: it is otherwise indistinguishable from "nobody has a build".

~~ A hardcoded Copilot path in generated agentic locks ~~

`update-release-notes`, `memory-leak-fixer`, `performance-fixer` and
`auto-triage` had failed on every run for twelve days with
`spawn /usr/local/bin/copilot ENOENT`, reported by the harness as
`pre-flight: command not found: /usr/local/bin/copilot`. The path is baked
into locks generated by gh-aw v0.84.3; the workflow sources do not contain it,
so it cannot be corrected in this repository except by regenerating.

Recompiling with gh-aw v0.87.10 emits `${RUNNER_TEMP}/gh-aw/bin/copilot`,
which the setup action actually installs. Verified on this pull request rather
than by inspection: the `Fixer - Memory Leak` and `Fixer - Performance` agent
jobs went green and logged
`pre-flight: command is accessible and executable`.

  * Every lock, `.github/aw/actions-lock.json`, and `.gitattributes` came from
    `gh aw compile`; none was hand-edited. The same run removed
    `agentics-maintenance.yml`, which was itself generated by
    `pkg/workflow/maintenance_workflow.go` and is no longer emitted — its
    scheduled cleanup of expiring safe outputs now lives inside the gh-aw
    runtime rather than in a workflow this repository stores.

~~ Untested irreversible release operations ~~

`Push-ReleaseTag` publishes the release tag and owns four behaviours that are
expensive to get wrong — refusing a tag that already points elsewhere, staying
silent when it already points at the right commit, refusing to act without
`-Push`, and verifying the remote after pushing — and had no test at all. A
tag is immutable once a release consumes it, so a regression here is not
recoverable by re-running.

The new cases drive a local bare remote, so they assert the guard, the
create-and-verify path, the idempotent repeat, and that a conflicting commit
is rejected in both dry-run and `-Push` form without touching the network.
`Resolve-NuGetPackageVersion` gains coverage of the versions that must never
reach nuget.org, which pins the boundary that decides when Finish disambiguates
an abbreviated preview. The suite moves from 69 to 84 assertions.

~~ Test suites that ran nowhere ~~

Separately from having no tests, several suites had tests that no workflow
executed: the release-note renderer and GitHub Release summary mutation under
`scripts/infra/docs/release_notes/tests`, the Skia milestone tooling under
`.agents/skills/update-skia/scripts`, and
`.github/scripts/tests/skia-sync-detect.test.sh`. All passed locally, so the
gap was purely in wiring, and a regression in any of them would have reached
`main` unremarked.

Release-note coverage joins `release-tooling-tests` because it is release
tooling, and its path filters grow to include `scripts/infra/docs` and
`update-github-release-summaries.yml`, whose input contract Finish dispatches.
The rest move to a new `automation-tooling-tests` workflow rather than
widening the release gate to cover unrelated automation. Both keep narrow path
filters so neither becomes a catch-all.

~~ A dashboard that under-reported instead of erroring ~~

`ci-status` queried `nightly-fix-finder.lock.yml`, deleted months earlier, and
advertised an `API Diff` workflow that never existed. Neither shows up as a
failure: the GitHub API returns nothing for an unknown workflow file, so the
dashboard silently omitted rows while looking healthy. It also tracked none of
the fixer, tracking, or release workflows.

The registry is corrected against the workflows that exist, and a new test
asserts that each tracked file is present, that its declared trigger appears
in its `on:` block, that its display name matches, and that anything declared
scheduled really defines a cron. The test parses committed YAML including
generated locks, so it does not need the gh-aw compiler. It was red when
written: it caught `persist-aw-data` registered as `push` when it is
`workflow_run`, alongside the stale rows.

~~ Retiring the xamarin feed mirror ~~

`manage-nuget-feed` could not run in its committed state. The workflow passed
`skiasharp-ci` while the script's `ValidateSet` accepted only `skiasharp` and
`skiasharp-transport`, so the default `both` always contained a leg that could
not bind. The break dates to the rename in 2dd508790e4; the last successful
run predates it by six days.

The destination the rename introduced does not exist — `skiasharp-transport`
returns 404, while `skiasharp`, `skiasharp-ci`, `dotnet-libraries` and
`dotnet-libraries-transport` all return 200. The one-shot xamarin-to-dnceng
mirroring it performed is complete, since `dnceng/public/skiasharp` is
populated, and nothing outside its own workflow referenced it.

Both the workflow and `scripts/infra/package/manage-nuget-feed.ps1` are
removed rather than repaired. Repointing a historical mirror at the shared
Arcade-managed feeds that now carry the release path would need feed-owner
agreement that this change does not have, and package publication is already
owned by the Arcade shipping and transport routing established in 2dd508790e4.

~~ An unguarded force push in docs automation ~~

`auto-docs-submodule-sync` used a bare `git push --force` on
`automation/update-docs-submodule`, which AGENTS.md forbids: automation may
force-update a branch it owns only with `--force-with-lease`.

The remote SHA is now captured with `git ls-remote` before the branch is
rebuilt, and the push leases against it explicitly. The explicit
`<ref>:<sha>` form is required rather than stylistic — `checkout -B` from
`origin/main` leaves the local branch with no reflog relationship to the
remote, so a bare `--force-with-lease` has no recorded expectation to compare
and degrades to an ordinary force. Confirmed against throwaway repositories:
an empty lease creates an absent branch and refuses an existing one, and a
lease captured before another writer moved the branch is rejected with
`stale info`.

~~ Workflow references the repository had moved past ~~

  * `build-site` excluded `CODE_OF_CONDUCT.md` from its path filter, but the
    file is `CODE-OF-CONDUCT.md`. The entry never matched, so every edit to
    the code of conduct rebuilt the full site.

  * `rebase` pinned `actions/checkout@v2`, a floating major tag on a path that
    runs rarely enough for the drift to go unnoticed. It now uses the same
    pinned SHA as the rest of the repository.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>
```

mattleibow:
```text
[infra] Break the artifact-size check_run loop and repair CI automation (#4915)

Context: https://github.com/github/gh-aw/releases/tag/v0.87.10

Three independent pieces of repository automation were failing at once, each
with its own root cause. The governing rule applied throughout is that an
automation trigger must be narrow enough to be evaluated before a runner
starts: GitHub cannot filter `check_run`, so the only way to stop the fan-out
is to stop subscribing to the event, and the only way to keep the behaviour it
provided is to ask Azure DevOps directly on a bounded schedule.

~~ A cross-workflow check_run loop, not an external fan-out ~~

`track-artifact-sizes` subscribed to `check_run: [completed]`. That event has
no app, name, or path filter, so every completed check run on the default
branch started a runner and the filtering only happened afterwards. Both
`track-artifact-sizes` and `persist-aw-data` had reached 40,000 runs.

The obvious reading — that broad external Azure checks drive this — is
wrong.
The commit carried 1,288 check runs, of which 1,229 were `github-actions` and
only 61 `azure-pipelines`, and the hour after the push saw 153 track runs
against exactly ONE Azure check completion and 608 Actions check completions.
Growth was steady at ~150/hr rather than exponential, which is only consistent
with GitHub suppressing a workflow re-triggering itself through its own Actions
check runs while leaving a cross-workflow cycle unsuppressed:
`track-artifact-sizes` -> `persist-aw-data` (`workflow_run`) -> its
`Persist to aw-data` check run completes -> `check_run` -> back to the start.
That relationship is 1:1, which is exactly the observed rate.

The trigger is removed rather than narrowed, because narrowing is not
available at any level that would help.

  * A job-level `if:` cannot fix this. Each firing published three check runs
    — `resolve` succeeded while `pr` and `nightly` were skipped — and a
    skipped job still publishes a check run, so the gate would have kept
    feeding the cycle while still paying for a runner.

  * PR reports are preserved by `scripts/infra/perf/sizes/find_pr_builds.py`,
    polled every 30 minutes, which asks for exactly the completed and
    succeeded build of pipeline 345 on `refs/pull/<n>/merge`. Worst-case
    latency becomes ~30 minutes; `workflow_dispatch` remains the manual path.

  * `render_pr_md.py` now emits a hidden `build=` stamp beside the unchanged
    visible marker so the poll can skip a build it already reported. Each
    measurement downloads roughly a gigabyte, so the stamp is written on
    failure too — otherwise an unmeasurable build would be re-selected and
    re-downloaded every 30 minutes indefinitely, recreating in miniature the
    runaway this change exists to remove.

  * `resolve` must emit `[]` rather than an unset output on the nightly path.
    GitHub expands `strategy.matrix` before the consuming job's `if:`, so
    `fromJSON('')` fails the run before the guard can skip it, and because
    `persist-aw-data` only persists a successful `workflow_run`, that failure
    would have silently stopped the nightly size history from being written.

  * Discovery and the per-PR stamp read live in `find_pr_builds.py` so they can
    paginate and fail closed. The per-issue comments endpoint ignores `sort`
    and `direction` — asc and desc return identical oldest-first ids — so the
    report sits on the LAST page once a PR passes 100 comments, and a
    single-page read would re-measure it forever. An unreadable stamp now
    skips that PR for the sweep rather than being read as "unmeasured";
    widespread lookup failure is reported, since it is otherwise
    indistinguishable from "nobody has a build".

  * The stamp is written by an unconditional claim step that runs before the
    download. `if: failure()` cannot cover a job hard-timeout, so a 30-minute
    timeout mid-download would otherwise leave the build unstamped and
    permanently re-selected. `workflow_dispatch` ignores stamps and remains
    the deliberate retry path.

  * The reader takes the FIRST marker comment, matching every writer's
    `find()`. Two marker comments are reachable — the concurrency group keys a
    manual dispatch on its build id but the sweep on the literal `sweep` — and
    a last-match reader would then read a comment the writers never update,
    re-downloading the same build every 30 minutes for the life of the PR.

~~ A hardcoded Copilot path in generated agentic locks ~~

`update-release-notes`, `memory-leak-fixer`, `performance-fixer` and
`auto-triage` had failed on every run for twelve days with
`spawn /usr/local/bin/copilot ENOENT`, reported by the harness as
`pre-flight: command not found: /usr/local/bin/copilot`. The path is baked
into locks generated by gh-aw v0.84.3; the workflow sources do not contain it,
so it cannot be corrected in this repository except by regenerating.

Recompiling with gh-aw v0.87.10 emits `${RUNNER_TEMP}/gh-aw/bin/copilot`,
which the setup action actually installs. Verified on this pull request rather
than by inspection: the `Fixer - Memory Leak` and `Fixer - Performance` agent
jobs went green and logged
`pre-flight: command is accessible and executable`.

  * Every lock, `.github/aw/actions-lock.json`, and `.gitattributes` came from
    `gh aw compile`; none was hand-edited. The same run removed
    `agentics-maintenance.yml`, which was itself generated by
    `pkg/workflow/maintenance_workflow.go` and is no longer emitted — its
    scheduled cleanup of expiring safe outputs now lives inside the gh-aw
    runtime rather than in a workflow this repository stores.

~~ Untested irreversible release operations ~~

`Push-ReleaseTag` publishes the release tag and owns four behaviours that are
expensive to get wrong — refusing a tag that already points elsewhere, staying
silent when it already points at the right commit, refusing to act without
`-Push`, and verifying the remote after pushing — and had no test at all. A
tag is immutable once a release consumes it, so a regression here is not
recoverable by re-running.

The new cases drive a local bare remote, so they assert the guard, the
create-and-verify path, the idempotent repeat, and that a conflicting commit
is rejected in both dry-run and `-Push` form without touching the network.
`Resolve-NuGetPackageVersion` gains coverage of the versions that must never
reach nuget.org, which pins the boundary that decides when Finish disambiguates
an abbreviated preview. The suite moves from 69 to 84 assertions.

~~ Test suites that ran nowhere ~~

Separately from having no tests, several suites had tests that no workflow
executed: the release-note renderer and GitHub Release summary mutation under
`scripts/infra/docs/release_notes/tests`, the Skia milestone tooling under
`.agents/skills/update-skia/scripts`, and
`.github/scripts/tests/skia-sync-detect.test.sh`. All passed locally, so the
gap was purely in wiring, and a regression in any of them would have reached
`main` unremarked.

Release-note coverage joins `release-tooling-tests` because it is release
tooling, and its path filters grow to include `scripts/infra/docs` and
`update-github-release-summaries.yml`, whose input contract Finish dispatches.
The rest move to a new `automation-tooling-tests` workflow rather than
widening the release gate to cover unrelated automation. Both keep narrow path
filters so neither becomes a catch-all.

~~ A dashboard that under-reported instead of erroring ~~

`ci-status` queried `nightly-fix-finder.lock.yml`, deleted months earlier, and
advertised an `API Diff` workflow that never existed. Neither shows up as a
failure: the GitHub API returns nothing for an unknown workflow file, so the
dashboard silently omitted rows while looking healthy. It also tracked none of
the fixer, tracking, or release workflows.

The registry is corrected against the workflows that exist, and a new test
asserts that each tracked file is present, that its declared trigger appears
in its `on:` block, that its display name matches, that anything declared
scheduled really defines a cron, and that any schedule the table states agrees
with the workflow. That last check matters because gh-aw re-jitters every
generated cron on upgrade, so documented minute precision goes stale the next
time the compiler runs; the literals are gone and the test refuses their
return. The test parses committed YAML including generated locks, so it does
not need the gh-aw compiler. It was red when written — it caught
`persist-aw-data` registered as `push` when it is `workflow_run` alongside the
stale rows, then caught `Track - Benchmarks` documented as nightly when it runs
every six hours.

~~ Retiring the xamarin feed mirror ~~

`manage-nuget-feed` could not run in its committed state. The workflow passed
`skiasharp-ci` while the script's `ValidateSet` accepted only `skiasharp` and
`skiasharp-transport`, so the default `both` always contained a leg that could
not bind. The break dates to the rename in 2dd508790e4; the last successful
run predates it by six days.

The destination the rename introduced does not exist — `skiasharp-transport`
returns 404, while `skiasharp`, `skiasharp-ci`, `dotnet-libraries` and
`dotnet-libraries-transport` all return 200. The one-shot xamarin-to-dnceng
mirroring it performed is complete, since `dnceng/public/skiasharp` is
populated, and nothing outside its own workflow referenced it.

Both the workflow and `scripts/infra/package/manage-nuget-feed.ps1` are
removed rather than repaired. Repointing a historical mirror at the shared
Arcade-managed feeds that now carry the release path would need feed-owner
agreement that this change does not have, and package publication is already
owned by the Arcade shipping and transport routing established in 2dd508790e4.

~~ An unguarded force push in docs automation ~~

`auto-docs-submodule-sync` used a bare `git push --force` on
`automation/update-docs-submodule`, which AGENTS.md forbids: automation may
force-update a branch it owns only with `--force-with-lease`.

The remote SHA is now captured with `git ls-remote` before the branch is
rebuilt, and the push leases against it explicitly. The explicit
`<ref>:<sha>` form is required rather than stylistic — `checkout -B` from
`origin/main` leaves the local branch with no reflog relationship to the
remote, so a bare `--force-with-lease` has no recorded expectation to compare
and degrades to an ordinary force. Confirmed against throwaway repositories:
an empty lease creates an absent branch and refuses an existing one, and a
lease captured before another writer moved the branch is rejected with
`stale info`.

~~ Workflow references the repository had moved past ~~

  * `build-site` excluded `CODE_OF_CONDUCT.md` from its path filter, but the
    file is `CODE-OF-CONDUCT.md`. The entry never matched, so every edit to
    the code of conduct rebuilt the full site.

  * `rebase` pinned `actions/checkout@v2`, a floating major tag on a path that
    runs rarely enough for the drift to go unnoticed. It now uses the same
    pinned SHA as the rest of the repository.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>
```

mattleibow:
```text
[infra] Break the artifact-size check_run loop and repair CI automation (#4915)

Context: https://github.com/github/gh-aw/releases/tag/v0.87.10

Three independent pieces of repository automation were failing at once, each
with its own root cause. The governing rule applied throughout is that an
automation trigger must be narrow enough to be evaluated before a runner
starts: GitHub cannot filter `check_run`, so the only way to stop the fan-out
is to stop subscribing to the event, and the only way to keep the behaviour it
provided is to ask Azure DevOps directly on a bounded schedule.

~~ A cross-workflow check_run loop, not an external fan-out ~~

`track-artifact-sizes` subscribed to `check_run: [completed]`. That event has
no app, name, or path filter, so every completed check run on the default
branch started a runner and the filtering only happened afterwards. Both
`track-artifact-sizes` and `persist-aw-data` had reached 40,000 runs.

The obvious reading — that broad external Azure checks drive this — is
wrong.
The commit carried 1,288 check runs, of which 1,229 were `github-actions` and
only 61 `azure-pipelines`, and the hour after the push saw 153 track runs
against exactly ONE Azure check completion and 608 Actions check completions.
Growth was steady at ~150/hr rather than exponential, which is only consistent
with GitHub suppressing a workflow re-triggering itself through its own Actions
check runs while leaving a cross-workflow cycle unsuppressed:
`track-artifact-sizes` -> `persist-aw-data` (`workflow_run`) -> its
`Persist to aw-data` check run completes -> `check_run` -> back to the start.
That relationship is 1:1, which is exactly the observed rate.

The trigger is removed rather than narrowed, because narrowing is not
available at any level that would help.

  * A job-level `if:` cannot fix this. Each firing published three check runs
    — `resolve` succeeded while `pr` and `nightly` were skipped — and a
    skipped job still publishes a check run, so the gate would have kept
    feeding the cycle while still paying for a runner.

  * PR reports are preserved by `scripts/infra/perf/sizes/find_pr_builds.py`,
    polled every 30 minutes, which asks for exactly the completed and
    succeeded build of pipeline 345 on `refs/pull/<n>/merge`. Worst-case
    latency becomes ~30 minutes; `workflow_dispatch` remains the manual path.

  * `render_pr_md.py` now emits a hidden `build=` stamp beside the unchanged
    visible marker so the poll can skip a build it already reported. Each
    measurement downloads roughly a gigabyte, so the stamp is written on
    failure too — otherwise an unmeasurable build would be re-selected and
    re-downloaded every 30 minutes indefinitely, recreating in miniature the
    runaway this change exists to remove.

  * `resolve` must emit `[]` rather than an unset output on the nightly path.
    GitHub expands `strategy.matrix` before the consuming job's `if:`, so
    `fromJSON('')` fails the run before the guard can skip it, and because
    `persist-aw-data` only persists a successful `workflow_run`, that failure
    would have silently stopped the nightly size history from being written.

  * Discovery and the per-PR stamp read live in `find_pr_builds.py` so they can
    paginate and fail closed. The per-issue comments endpoint ignores `sort`
    and `direction` — asc and desc return identical oldest-first ids — so the
    report sits on the LAST page once a PR passes 100 comments, and a
    single-page read would re-measure it forever. An unreadable stamp now
    skips that PR for the sweep rather than being read as "unmeasured";
    widespread lookup failure is reported, since it is otherwise
    indistinguishable from "nobody has a build".

  * The stamp is written by an unconditional claim step that runs before the
    download. `if: failure()` cannot cover a job hard-timeout, so a 30-minute
    timeout mid-download would otherwise leave the build unstamped and
    permanently re-selected. `workflow_dispatch` ignores stamps and remains
    the deliberate retry path.

  * The reader takes the FIRST marker comment, matching every writer's
    `find()`. Two marker comments are reachable — the concurrency group keys a
    manual dispatch on its build id but the sweep on the literal `sweep` — and
    a last-match reader would then read a comment the writers never update,
    re-downloading the same build every 30 minutes for the life of the PR.

~~ A hardcoded Copilot path in generated agentic locks ~~

`update-release-notes`, `memory-leak-fixer`, `performance-fixer` and
`auto-triage` had failed on every run for twelve days with
`spawn /usr/local/bin/copilot ENOENT`, reported by the harness as
`pre-flight: command not found: /usr/local/bin/copilot`. The path is baked
into locks generated by gh-aw v0.84.3; the workflow sources do not contain it,
so it cannot be corrected in this repository except by regenerating.

Recompiling with gh-aw v0.87.10 emits `${RUNNER_TEMP}/gh-aw/bin/copilot`,
which the setup action actually installs. Verified on this pull request rather
than by inspection: the `Fixer - Memory Leak` and `Fixer - Performance` agent
jobs went green and logged
`pre-flight: command is accessible and executable`.

  * Every lock, `.github/aw/actions-lock.json`, and `.gitattributes` came from
    `gh aw compile`; none was hand-edited. The same run removed
    `agentics-maintenance.yml`, which was itself generated by
    `pkg/workflow/maintenance_workflow.go` and is no longer emitted — its
    scheduled cleanup of expiring safe outputs now lives inside the gh-aw
    runtime rather than in a workflow this repository stores.

~~ The absent-package contract in the shared HTTP helper ~~

`_common.http_get` raises `RuntimeError` for every HTTP and connection failure,
and three callers depend on exactly that to mean "absent or unavailable":
`feed_versions`, `track.enumerate_feed_packages` and
`measure_pr.resolve_pr_number` all catch it and return an empty result.

Fast-failing a non-retryable 4xx — added so a doomed request stops sleeping
through four attempts — must therefore not surface `urllib.error.HTTPError`,
which is an `OSError` and escapes those handlers. Reproduced against nuget.org:
an unpublished package id answered `[]` on `main` but raised, which would have
broken the nightly size tracker the first time it looked for an EAP package
that had not shipped yet. The status is wrapped in `RuntimeError` with its code
and URL, so the answer is unchanged and the sleeps are still gone: that lookup
now returns `[]` in 0.6s where `main` needed 30s of backoff to say the same
thing.

  * 403, 408 and 429 keep their backoff. `http_get` gained a `headers`
    parameter so the size sweep's authenticated GitHub calls route through it,
    and GitHub answers a secondary rate limit with 403 — the shape a sweep of
    dozens of sequential requests every 30 minutes actually provokes. Treating
    that as permanent made the sweep less resilient than before the fast-fail
    existed.

  * The tests pin the contract from both directions, and are mutation-checked:
    reverting the wrapper turns six red, and narrowing the retryable set turns
    two red.

~~ Untested irreversible release operations ~~

`Push-ReleaseTag` publishes the release tag and owns four behaviours that are
expensive to get wrong — refusing a tag that already points elsewhere, staying
silent when it already points at the right commit, refusing to act without
`-Push`, and verifying the remote after pushing — and had no test at all. A
tag is immutable once a release consumes it, so a regression here is not
recoverable by re-running.

The new cases drive a local bare remote, so they assert the guard, the
create-and-verify path, the idempotent repeat, and that a conflicting commit
is rejected in both dry-run and `-Push` form without touching the network.
`Resolve-NuGetPackageVersion` gains coverage of the versions that must never
reach nuget.org, which pins the boundary that decides when Finish disambiguates
an abbreviated preview. The suite moves from 69 to 84 assertions.

~~ Test suites that ran nowhere ~~

Separately from having no tests, several suites had tests that no workflow
executed: the release-note renderer and GitHub Release summary mutation under
`scripts/infra/docs/release_notes/tests`, the Skia milestone tooling under
`.agents/skills/update-skia/scripts`, and
`.github/scripts/tests/skia-sync-detect.test.sh`. All passed locally, so the
gap was purely in wiring, and a regression in any of them would have reached
`main` unremarked.

Release-note coverage joins `release-tooling-tests` because it is release
tooling, and its path filters grow to include `scripts/infra/docs` and
`update-github-release-summaries.yml`, whose input contract Finish dispatches.
The rest move to a new `automation-tooling-tests` workflow rather than
widening the release gate to cover unrelated automation. Both keep narrow path
filters so neither becomes a catch-all.

~~ A dashboard that under-reported instead of erroring ~~

`ci-status` queried `nightly-fix-finder.lock.yml`, deleted months earlier, and
advertised an `API Diff` workflow that never existed. Neither shows up as a
failure: the GitHub API returns nothing for an unknown workflow file, so the
dashboard silently omitted rows while looking healthy. It also tracked none of
the fixer, tracking, or release workflows.

The registry is corrected against the workflows that exist, and a new test
asserts that each tracked file is present, that its declared trigger appears
in its `on:` block, that its display name matches, that anything declared
scheduled really defines a cron, and that any schedule the table states agrees
with the workflow. That last check matters because gh-aw re-jitters every
generated cron on upgrade, so documented minute precision goes stale the next
time the compiler runs; the literals are gone and the test refuses their
return. The test parses committed YAML including generated locks, so it does
not need the gh-aw compiler. It was red when written — it caught
`persist-aw-data` registered as `push` when it is `workflow_run` alongside the
stale rows, then caught `Track - Benchmarks` documented as nightly when it runs
every six hours.

~~ Retiring the xamarin feed mirror ~~

`manage-nuget-feed` could not run in its committed state. The workflow passed
`skiasharp-ci` while the script's `ValidateSet` accepted only `skiasharp` and
`skiasharp-transport`, so the default `both` always contained a leg that could
not bind. The break dates to the rename in 2dd508790e4; the last successful
run predates it by six days.

The destination the rename introduced does not exist — `skiasharp-transport`
returns 404, while `skiasharp`, `skiasharp-ci`, `dotnet-libraries` and
`dotnet-libraries-transport` all return 200. The one-shot xamarin-to-dnceng
mirroring it performed is complete, since `dnceng/public/skiasharp` is
populated, and nothing outside its own workflow referenced it.

Both the workflow and `scripts/infra/package/manage-nuget-feed.ps1` are
removed rather than repaired. Repointing a historical mirror at the shared
Arcade-managed feeds that now carry the release path would need feed-owner
agreement that this change does not have, and package publication is already
owned by the Arcade shipping and transport routing established in 2dd508790e4.

~~ An unguarded force push in docs automation ~~

`auto-docs-submodule-sync` used a bare `git push --force` on
`automation/update-docs-submodule`, which AGENTS.md forbids: automation may
force-update a branch it owns only with `--force-with-lease`.

The remote SHA is now captured with `git ls-remote` before the branch is
rebuilt, and the push leases against it explicitly. The explicit
`<ref>:<sha>` form is required rather than stylistic — `checkout -B` from
`origin/main` leaves the local branch with no reflog relationship to the
remote, so a bare `--force-with-lease` has no recorded expectation to compare
and degrades to an ordinary force. Confirmed against throwaway repositories:
an empty lease creates an absent branch and refuses an existing one, and a
lease captured before another writer moved the branch is rejected with
`stale info`.

~~ Workflow references the repository had moved past ~~

  * `build-site` excluded `CODE_OF_CONDUCT.md` from its path filter, but the
    file is `CODE-OF-CONDUCT.md`. The entry never matched, so every edit to
    the code of conduct rebuilt the full site.

  * `rebase` pinned `actions/checkout@v2`, a floating major tag on a path that
    runs rarely enough for the drift to go unnoticed. It now uses the same
    pinned SHA as the rest of the repository.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>
```

mattleibow:
```text
[infra] Break the artifact-size check_run loop and repair CI automation (#4915)

Context: https://github.com/github/gh-aw/releases/tag/v0.87.10

Three independent pieces of repository automation were failing at once, each
with its own root cause. The governing rule applied throughout is that an
automation trigger must be narrow enough to be evaluated before a runner
starts: GitHub cannot filter `check_run`, so the only way to stop the fan-out
is to stop subscribing to the event, and the only way to keep the behaviour it
provided is to ask Azure DevOps directly on a bounded schedule.

~~ A cross-workflow check_run loop, not an external fan-out ~~

`track-artifact-sizes` subscribed to `check_run: [completed]`. That event has
no app, name, or path filter, so every completed check run on the default
branch started a runner and the filtering only happened afterwards. Both
`track-artifact-sizes` and `persist-aw-data` had reached 40,000 runs.

The obvious reading — that broad external Azure checks drive this — is
wrong.
The commit carried 1,288 check runs, of which 1,229 were `github-actions` and
only 61 `azure-pipelines`, and the hour after the push saw 153 track runs
against exactly ONE Azure check completion and 608 Actions check completions.
Growth was steady at ~150/hr rather than exponential, which is only consistent
with GitHub suppressing a workflow re-triggering itself through its own Actions
check runs while leaving a cross-workflow cycle unsuppressed:
`track-artifact-sizes` -> `persist-aw-data` (`workflow_run`) -> its
`Persist to aw-data` check run completes -> `check_run` -> back to the start.
That relationship is 1:1, which is exactly the observed rate.

The trigger is removed rather than narrowed, because narrowing is not
available at any level that would help.

  * A job-level `if:` cannot fix this. Each firing published three check runs
    — `resolve` succeeded while `pr` and `nightly` were skipped — and a
    skipped job still publishes a check run, so the gate would have kept
    feeding the cycle while still paying for a runner.

  * PR reports are preserved by `scripts/infra/perf/sizes/find_pr_builds.py`,
    polled every 30 minutes, which asks for exactly the completed and
    succeeded build of pipeline 345 on `refs/pull/<n>/merge`. Worst-case
    latency becomes ~30 minutes; `workflow_dispatch` remains the manual path.

  * `render_pr_md.py` now emits a hidden `build=` stamp beside the unchanged
    visible marker so the poll can skip a build it already reported. Each
    measurement downloads roughly a gigabyte, so the stamp is written on
    failure too — otherwise an unmeasurable build would be re-selected and
    re-downloaded every 30 minutes indefinitely, recreating in miniature the
    runaway this change exists to remove.

  * `resolve` must emit `[]` rather than an unset output on the nightly path.
    GitHub expands `strategy.matrix` before the consuming job's `if:`, so
    `fromJSON('')` fails the run before the guard can skip it, and because
    `persist-aw-data` only persists a successful `workflow_run`, that failure
    would have silently stopped the nightly size history from being written.

  * Discovery and the per-PR stamp read live in `find_pr_builds.py` so they can
    paginate and fail closed. The per-issue comments endpoint ignores `sort`
    and `direction` — asc and desc return identical oldest-first ids — so the
    report sits on the LAST page once a PR passes 100 comments, and a
    single-page read would re-measure it forever. An unreadable stamp now
    skips that PR for the sweep rather than being read as "unmeasured";
    widespread lookup failure is reported, since it is otherwise
    indistinguishable from "nobody has a build".

  * The stamp is written by an unconditional claim step that runs before the
    download. `if: failure()` cannot cover a job hard-timeout, so a 30-minute
    timeout mid-download would otherwise leave the build unstamped and
    permanently re-selected. `workflow_dispatch` ignores stamps and remains
    the deliberate retry path.

  * The reader takes the FIRST marker comment, matching every writer's
    `find()`. Two marker comments are reachable — the concurrency group keys a
    manual dispatch on its build id but the sweep on the literal `sweep` — and
    a last-match reader would then read a comment the writers never update,
    re-downloading the same build every 30 minutes for the life of the PR.

~~ A hardcoded Copilot path in generated agentic locks ~~

`update-release-notes`, `memory-leak-fixer`, `performance-fixer` and
`auto-triage` had failed on every run for twelve days with
`spawn /usr/local/bin/copilot ENOENT`, reported by the harness as
`pre-flight: command not found: /usr/local/bin/copilot`. The path is baked
into locks generated by gh-aw v0.84.3; the workflow sources do not contain it,
so it cannot be corrected in this repository except by regenerating.

Recompiling with gh-aw v0.87.10 emits `${RUNNER_TEMP}/gh-aw/bin/copilot`,
which the setup action actually installs. Verified on this pull request rather
than by inspection: the `Fixer - Memory Leak` and `Fixer - Performance` agent
jobs went green and logged
`pre-flight: command is accessible and executable`.

  * Every lock, `.github/aw/actions-lock.json`, and `.gitattributes` came from
    `gh aw compile`; none was hand-edited. The same run removed
    `agentics-maintenance.yml`, which was itself generated by
    `pkg/workflow/maintenance_workflow.go` and is no longer emitted — its
    scheduled cleanup of expiring safe outputs now lives inside the gh-aw
    runtime rather than in a workflow this repository stores.

~~ The absent-package contract in the shared HTTP helper ~~

`_common.http_get` raises `RuntimeError` for every HTTP and connection failure,
and three callers depend on exactly that to mean "absent or unavailable":
`feed_versions`, `track.enumerate_feed_packages` and
`measure_pr.resolve_pr_number` all catch it and return an empty result.

Fast-failing a non-retryable 4xx — added so a doomed request stops sleeping
through four attempts — must therefore not surface `urllib.error.HTTPError`,
which is an `OSError` and escapes those handlers. Reproduced against nuget.org:
an unpublished package id answered `[]` on `main` but raised, which would have
broken the nightly size tracker the first time it looked for an EAP package
that had not shipped yet. The status is wrapped in `RuntimeError` with its code
and URL, so the answer is unchanged and the sleeps are still gone: that lookup
now returns `[]` in 0.6s where `main` needed 30s of backoff to say the same
thing.

  * 403, 408 and 429 keep their backoff. `http_get` gained a `headers`
    parameter so the size sweep's authenticated GitHub calls route through it,
    and GitHub answers a secondary rate limit with 403 — the shape a sweep of
    dozens of sequential requests every 30 minutes actually provokes. Treating
    that as permanent made the sweep less resilient than before the fast-fail
    existed.

  * The tests pin the contract from both directions, and are mutation-checked:
    reverting the wrapper turns six red, and narrowing the retryable set turns
    two red.

~~ Untested irreversible release operations ~~

`Push-ReleaseTag` publishes the release tag and owns four behaviours that are
expensive to get wrong — refusing a tag that already points elsewhere, staying
silent when it already points at the right commit, refusing to act without
`-Push`, and verifying the remote after pushing — and had no test at all. A
tag is immutable once a release consumes it, so a regression here is not
recoverable by re-running.

The new cases drive a local bare remote, so they assert the guard, the
create-and-verify path, the idempotent repeat, and that a conflicting commit
is rejected in both dry-run and `-Push` form without touching the network.
`Resolve-NuGetPackageVersion` gains coverage of the versions that must never
reach nuget.org, which pins the boundary that decides when Finish disambiguates
an abbreviated preview. The suite gains 15 assertions, 92 to 107.

~~ Test suites that ran nowhere ~~

Separately from having no tests, several suites had tests that no workflow
executed: the release-note renderer and GitHub Release summary mutation under
`scripts/infra/docs/release_notes/tests`, the Skia milestone tooling under
`.agents/skills/update-skia/scripts`, and
`.github/scripts/tests/skia-sync-detect.test.sh`. All passed locally, so the
gap was purely in wiring, and a regression in any of them would have reached
`main` unremarked.

Release-note coverage joins `release-tooling-tests` because it is release
tooling, and its path filters grow to include `scripts/infra/docs` and
`update-github-release-summaries.yml`, whose input contract Finish dispatches.
The rest move to a new `automation-tooling-tests` workflow rather than
widening the release gate to cover unrelated automation. Both keep narrow path
filters so neither becomes a catch-all.

~~ A dashboard that under-reported instead of erroring ~~

`ci-status` queried `nightly-fix-finder.lock.yml`, deleted months earlier, and
advertised an `API Diff` workflow that never existed. Neither shows up as a
failure: the GitHub API returns nothing for an unknown workflow file, so the
dashboard silently omitted rows while looking healthy. It also tracked none of
the fixer, tracking, or release workflows.

The registry is corrected against the workflows that exist, and a new test
asserts that each tracked file is present, that its declared trigger appears
in its `on:` block, that its display name matches, that anything declared
scheduled really defines a cron, and that any schedule the table states agrees
with the workflow. That last check matters because gh-aw re-jitters every
generated cron on upgrade, so documented minute precision goes stale the next
time the compiler runs; the literals are gone and the test refuses their
return. The test parses committed YAML including generated locks, so it does
not need the gh-aw compiler. It was red when written — it caught
`persist-aw-data` registered as `push` when it is `workflow_run` alongside the
stale rows, then caught `Track - Benchmarks` documented as nightly when it runs
every six hours.

~~ Retiring the xamarin feed mirror ~~

`manage-nuget-feed` could not run in its committed state. The workflow passed
`skiasharp-ci` while the script's `ValidateSet` accepted only `skiasharp` and
`skiasharp-transport`, so the default `both` always contained a leg that could
not bind. The break dates to the rename in 2dd508790e4; the last successful
run predates it by six days.

The destination the rename introduced does not exist — `skiasharp-transport`
returns 404, while `skiasharp`, `skiasharp-ci`, `dotnet-libraries` and
`dotnet-libraries-transport` all return 200. The one-shot xamarin-to-dnceng
mirroring it performed is complete, since `dnceng/public/skiasharp` is
populated, and nothing outside its own workflow referenced it.

Both the workflow and `scripts/infra/package/manage-nuget-feed.ps1` are
removed rather than repaired. Repointing a historical mirror at the shared
Arcade-managed feeds that now carry the release path would need feed-owner
agreement that this change does not have, and package publication is already
owned by the Arcade shipping and transport routing established in 2dd508790e4.

~~ An unguarded force push in docs automation ~~

`auto-docs-submodule-sync` used a bare `git push --force` on
`automation/update-docs-submodule`, which AGENTS.md forbids: automation may
force-update a branch it owns only with `--force-with-lease`.

The remote SHA is now captured with `git ls-remote` before the branch is
rebuilt, and the push leases against it explicitly. The explicit
`<ref>:<sha>` form is required rather than stylistic — `checkout -B` from
`origin/main` leaves the local branch with no reflog relationship to the
remote, so a bare `--force-with-lease` has no recorded expectation to compare
and degrades to an ordinary force. Confirmed against throwaway repositories:
an empty lease creates an absent branch and refuses an existing one, and a
lease captured before another writer moved the branch is rejected with
`stale info`.

~~ Workflow references the repository had moved past ~~

  * `build-site` excluded `CODE_OF_CONDUCT.md` from its path filter, but the
    file is `CODE-OF-CONDUCT.md`. The entry never matched, so every edit to
    the code of conduct rebuilt the full site.

  * `rebase` pinned `actions/checkout@v2`, a floating major tag on a path that
    runs rarely enough for the drift to go unnoticed. It now uses the same
    pinned SHA as the rest of the repository.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>
```

mattleibow:
```text
[infra] Break the artifact-size check_run loop and repair CI automation (#4915)

Context: https://github.com/github/gh-aw/releases/tag/v0.87.10

Three independent pieces of repository automation were failing at once, each
with its own root cause. The governing rule applied throughout is that an
automation trigger must be narrow enough to be evaluated before a runner
starts: GitHub cannot filter `check_run`, so the only way to stop the fan-out
is to stop subscribing to the event, and the only way to keep the behaviour it
provided is to ask Azure DevOps directly on a bounded schedule.

~~ A cross-workflow check_run loop, not an external fan-out ~~

`track-artifact-sizes` subscribed to `check_run: [completed]`. That event has
no app, name, or path filter, so every completed check run on the default
branch started a runner and the filtering only happened afterwards. Both
`track-artifact-sizes` and `persist-aw-data` had reached 40,000 runs.

The obvious reading — that broad external Azure checks drive this — is
wrong.
The commit carried 1,288 check runs, of which 1,229 were `github-actions` and
only 61 `azure-pipelines`, and the hour after the push saw 153 track runs
against exactly ONE Azure check completion and 608 Actions check completions.
Growth was steady at ~150/hr rather than exponential, which is only consistent
with GitHub suppressing a workflow re-triggering itself through its own Actions
check runs while leaving a cross-workflow cycle unsuppressed:
`track-artifact-sizes` -> `persist-aw-data` (`workflow_run`) -> its
`Persist to aw-data` check run completes -> `check_run` -> back to the start.
That relationship is 1:1, which is exactly the observed rate.

The trigger is removed rather than narrowed, because narrowing is not
available at any level that would help.

  * A job-level `if:` cannot fix this. Each firing published three check runs
    — `resolve` succeeded while `pr` and `nightly` were skipped — and a
    skipped job still publishes a check run, so the gate would have kept
    feeding the cycle while still paying for a runner.

  * PR reports are preserved by `scripts/infra/perf/sizes/find_pr_builds.py`,
    polled every 30 minutes, which asks for exactly the completed and
    succeeded build of pipeline 345 on `refs/pull/<n>/merge`. Worst-case
    latency becomes ~30 minutes; `workflow_dispatch` remains the manual path.

  * `render_pr_md.py` now emits a hidden `build=` stamp beside the unchanged
    visible marker so the poll can skip a build it already reported. Each
    measurement downloads roughly a gigabyte, so the stamp is written on
    failure too — otherwise an unmeasurable build would be re-selected and
    re-downloaded every 30 minutes indefinitely, recreating in miniature the
    runaway this change exists to remove.

  * `resolve` must emit `[]` rather than an unset output on the nightly path.
    GitHub expands `strategy.matrix` before the consuming job's `if:`, so
    `fromJSON('')` fails the run before the guard can skip it, and because
    `persist-aw-data` only persists a successful `workflow_run`, that failure
    would have silently stopped the nightly size history from being written.

  * Discovery and the per-PR stamp read live in `find_pr_builds.py` so they can
    paginate and fail closed. The per-issue comments endpoint ignores `sort`
    and `direction` — asc and desc return identical oldest-first ids — so the
    report sits on the LAST page once a PR passes 100 comments, and a
    single-page read would re-measure it forever. An unreadable stamp now
    skips that PR for the sweep rather than being read as "unmeasured";
    widespread lookup failure is reported, since it is otherwise
    indistinguishable from "nobody has a build".

  * The stamp is written by an unconditional claim step that runs before the
    download. `if: failure()` cannot cover a job hard-timeout, so a 30-minute
    timeout mid-download would otherwise leave the build unstamped and
    permanently re-selected. `workflow_dispatch` ignores stamps and remains
    the deliberate retry path.

  * The reader takes the FIRST marker comment, matching every writer's
    `find()`. Two marker comments are reachable — the concurrency group keys a
    manual dispatch on its build id but the sweep on the literal `sweep` — and
    a last-match reader would then read a comment the writers never update,
    re-downloading the same build every 30 minutes for the life of the PR.

~~ A hardcoded Copilot path in generated agentic locks ~~

`update-release-notes`, `memory-leak-fixer`, `performance-fixer` and
`auto-triage` had failed on every run for twelve days with
`spawn /usr/local/bin/copilot ENOENT`, reported by the harness as
`pre-flight: command not found: /usr/local/bin/copilot`. The path is baked
into locks generated by gh-aw v0.84.3; the workflow sources do not contain it,
so it cannot be corrected in this repository except by regenerating.

Recompiling with gh-aw v0.87.10 emits `${RUNNER_TEMP}/gh-aw/bin/copilot`,
which the setup action actually installs. Verified on this pull request rather
than by inspection: the `Fixer - Memory Leak` and `Fixer - Performance` agent
jobs went green and logged
`pre-flight: command is accessible and executable`.

  * `.gitattributes` loses `merge=ours` on the lock attribute line. Keep it
    that way: `merge=ours` resolves a conflicting generated lock by silently
    retaining the local side, which is precisely how a compiler upgrade like
    this one would be swallowed during a merge and leave the old hardcoded
    path in place.

  * Every lock, `.github/aw/actions-lock.json`, and `.gitattributes` came from
    `gh aw compile`; none was hand-edited, and a fresh `gh aw compile` is a
    no-op against the committed tree. The same run removed
    `agentics-maintenance.yml`, which was itself generated by
    `pkg/workflow/maintenance_workflow.go` and is no longer emitted — its
    scheduled cleanup of expiring safe outputs now lives inside the gh-aw
    runtime rather than in a workflow this repository stores.

~~ The absent-package contract in the shared HTTP helper ~~

`_common.http_get` raises `RuntimeError` for every HTTP and connection failure,
and three callers depend on exactly that to mean "absent or unavailable":
`feed_versions`, `track.enumerate_feed_packages` and
`measure_pr.resolve_pr_number` all catch it and return an empty result.

Fast-failing a non-retryable 4xx — added so a doomed request stops sleeping
through four attempts — must therefore not surface `urllib.error.HTTPError`,
which is an `OSError` and escapes those handlers. Reproduced against nuget.org:
an unpublished package id answered `[]` on `main` but raised, which would have
broken the nightly size tracker the first time it looked for an EAP package
that had not shipped yet. The status is wrapped in `RuntimeError` with its code
and URL, so the answer is unchanged and the sleeps are still gone: that lookup
now returns `[]` in 0.6s where `main` needed 30s of backoff to say the same
thing.

  * 403, 408 and 429 keep their backoff. `http_get` gained a `headers`
    parameter so the size sweep's authenticated GitHub calls route through it,
    and GitHub answers a secondary rate limit with 403 — the shape a sweep of
    dozens of sequential requests every 30 minutes actually provokes. Treating
    that as permanent made the sweep less resilient than before the fast-fail
    existed.

  * The tests pin the contract from both directions, and are mutation-checked:
    reverting the wrapper turns six red, and narrowing the retryable set turns
    two red.

~~ Untested irreversible release operations ~~

`Push-ReleaseTag` publishes the release tag and owns four behaviours that are
expensive to get wrong — refusing a tag that already points elsewhere, staying
silent when it already points at the right commit, refusing to act without
`-Push`, and verifying the remote after pushing — and had no test at all. A
tag is immutable once a release consumes it, so a regression here is not
recoverable by re-running.

The new cases drive a local bare remote, so they assert the guard, the
create-and-verify path, the idempotent repeat, and that a conflicting commit
is rejected in both dry-run and `-Push` form without touching the network.
`Resolve-NuGetPackageVersion` gains coverage of the versions that must never
reach nuget.org, which pins the boundary that decides when Finish disambiguates
an abbreviated preview. The suite gains 15 assertions, 92 to 107.

~~ Test suites that ran nowhere ~~

Separately from having no tests, several suites had tests that no workflow
executed: the release-note renderer and GitHub Release summary mutation under
`scripts/infra/docs/release_notes/tests`, the Skia milestone tooling under
`.agents/skills/update-skia/scripts`, and
`.github/scripts/tests/skia-sync-detect.test.sh`. All passed locally, so the
gap was purely in wiring, and a regression in any of them would have reached
`main` unremarked.

Release-note coverage joins `release-tooling-tests` because it is release
tooling, and its path filters grow to include `scripts/infra/docs` and
`update-github-release-summaries.yml`, whose input contract Finish dispatches.
The rest move to a new `automation-tooling-tests` workflow rather than
widening the release gate to cover unrelated automation. Both keep narrow path
filters so neither becomes a catch-all.

~~ A dashboard that under-reported instead of erroring ~~

`ci-status` queried `nightly-fix-finder.lock.yml`, deleted months earlier, and
advertised an `API Diff` workflow that never existed. Neither shows up as a
failure: the GitHub API returns nothing for an unknown workflow file, so the
dashboard silently omitted rows while looking healthy. It also tracked none of
the fixer, tracking, or release workflows.

The registry is corrected against the workflows that exist, and a new test
asserts that each tracked file is present, that its declared trigger appears
in its `on:` block, that its display name matches, that anything declared
scheduled really defines a cron, and that any schedule the table states agrees
with the workflow. That last check matters because gh-aw re-jitters every
generated cron on upgrade, so documented minute precision goes stale the next
time the compiler runs; the literals are gone and the test refuses their
return. The test parses committed YAML including generated locks, so it does
not need the gh-aw compiler. It was red when written — it caught
`persist-aw-data` registered as `push` when it is `workflow_run` alongside the
stale rows, then caught `Track - Benchmarks` documented as nightly when it runs
every six hours.

~~ Retiring the xamarin feed mirror ~~

`manage-nuget-feed` could not run in its committed state. The workflow passed
`skiasharp-ci` while the script's `ValidateSet` accepted only `skiasharp` and
`skiasharp-transport`, so the default `both` always contained a leg that could
not bind. The break dates to the rename in 2dd508790e4; the last successful
run predates it by six days.

The destination the rename introduced does not exist — `skiasharp-transport`
returns 404, while `skiasharp`, `skiasharp-ci`, `dotnet-libraries` and
`dotnet-libraries-transport` all return 200. The one-shot xamarin-to-dnceng
mirroring it performed is complete, since `dnceng/public/skiasharp` is
populated, and nothing outside its own workflow referenced it.

Both the workflow and `scripts/infra/package/manage-nuget-feed.ps1` are
removed rather than repaired. Repointing a historical mirror at the shared
Arcade-managed feeds that now carry the release path would need feed-owner
agreement that this change does not have, and package publication is already
owned by the Arcade shipping and transport routing established in 2dd508790e4.

~~ An unguarded force push in docs automation ~~

`auto-docs-submodule-sync` used a bare `git push --force` on
`automation/update-docs-submodule`, which AGENTS.md forbids: automation may
force-update a branch it owns only with `--force-with-lease`.

The remote SHA is now captured with `git ls-remote` before the branch is
rebuilt, and the push leases against it explicitly. The explicit
`<ref>:<sha>` form is required rather than stylistic — `checkout -B` from
`origin/main` leaves the local branch with no reflog relationship to the
remote, so a bare `--force-with-lease` has no recorded expectation to compare
and degrades to an ordinary force. Confirmed against throwaway repositories:
an empty lease creates an absent branch and refuses an existing one, and a
lease captured before another writer moved the branch is rejected with
`stale info`.

~~ Workflow references the repository had moved past ~~

  * `build-site` excluded `CODE_OF_CONDUCT.md` from its path filter, but the
    file is `CODE-OF-CONDUCT.md`. The entry never matched, so every edit to
    the code of conduct rebuilt the full site.

  * `rebase` pinned `actions/checkout@v2`, a floating major tag on a path that
    runs rarely enough for the drift to go unnoticed. It now uses the same
    pinned SHA as the rest of the repository.

Co-authored-by: Matthew Leibowitz <mattleibow@live.com>
Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>
```

